PR #3, opened 2018-09-06
Fix multu

simonlindholm:


simonlindholm:
@matt-kempster FWIW I have managed to stack up a larger patch queue at https://github.com/matt-kempster/mips_to_c/compare/master...simonlindholm:wip which you might be interested in when you get back to this. (I can send that as one or several PRs if wanted.)

matt-kempster:
When doubles are implemented, the `%hi` result of `multu` would have to change, right?

simonlindholm:
Doubles aren't connected to this AFAIK -- `%hi` just stores the high part of the multiplication, which cannot be represented in plain C (well, I guess you could write `(u32)((u64)a * (u64)b >> 32)`, but that would likely give different codegen).

PR #41, opened 2019-06-25
Use black to autoformat codebase and add pre-commit config

matt-kempster:
To get this all to work, just `pip install pre-commit black`, making sure you can `which pre-commit` and `which black`.

Use `black .` to autoformat all code manually, or let `pre-commit` handle it for you at commit time.

Shrug. We don't have to do this, but I like black :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+-   repo: https://github.com/python/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.7
+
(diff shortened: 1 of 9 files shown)


matt-kempster:
We should probably just have a `pip-requirements.txt` at this point...

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+-   repo: https://github.com/python/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.7
```
simonlindholm: should this be 3.6?

PR #42, opened 2019-06-30
 Restructure tests and rewrite shell-scripts in Python

matt-kempster:
- Restructure tests to have `tests/end_to_end/NAME_OF_TEST/{orig.c,irix-g.s,irix-o2.s,irix-g-out.c,irix-o2-out.c`
 - Rewrite the `add-test` and `run-tests` scripts in Python. They can be unit tested as desired
 - For some reason I had to make all the imports in `src` relative


tests/add_test.py
```diff
@@ -0,0 +1,228 @@
...
+        asm_file_path = Path(str(test_dir / asm_filename) + ".s")
+        irix_compile_with_flag(in_file, asm_file_path, flag)
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description="Add or update end-to-end decompiler tests."
+    )
+    parser.add_argument(
+        "files", help="files containing C code to compile (then decompile)", nargs="+"
```
simonlindholm: What's the expected form of these paths? `./tests/<name>/orig.c`? `any-dir/<name>.c`?

tests/add_test.py
```diff
@@ -0,0 +1,228 @@
...
+    out_file.write_bytes(final_asm)
+    logging.info(f"Successfully wrote disassembly to {out_file}.")
+
+
+def add_test_from_file(orig_file: Path) -> None:
+    test_dir = Path("end_to_end") / orig_file.stem
+    try:
+        test_dir.mkdir()
+    except FileExistsError:
+        raise Exception(f"{test_dir} already exists. Name your test something else.")
```
simonlindholm: we need to support test updates as well

run_tests.py
```diff
@@ -0,0 +1,125 @@
...
+                    final_contents,
+                ]
+            )
+        )
+    return changed
+
+
+def decompile_and_capture_output(output_path: Path, asm_file_path: Path) -> str:
+    out_string = io.StringIO()
+    with contextlib.redirect_stdout(out_string):
```
simonlindholm: redirect stderr too?

run_tests.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
```
simonlindholm: chmod this file and the other one 755

run_tests.py
```diff
@@ -0,0 +1,125 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Run and record end-to-end decompiler tests."
+    )
+    parser.add_argument(
+        "--debug", dest="debug", help="print debug info", action="store_true"
+    )
+    parser.add_argument(
+        "--overwrite",
```
simonlindholm: Ok, so the new workflow is "run ./run_tests.py; if it outputs changes and those are expected, rerun with --overwrite"? That sounds reasonable to me, especially if running tests is much faster now, though it's a slight regression from always having --overwrite and reading old versions from git history IMO

run_tests.py
```diff
@@ -0,0 +1,125 @@
...
+    return any_changed
+
+
+def main(should_overwrite: bool) -> int:
+    any_changed = False
+    for e2e_test_path in Path("tests/end_to_end").iterdir():
+        any_changed = run_e2e_test(e2e_test_path, should_overwrite) or any_changed
+
+    if any_changed:
+        logging.info("The output of some tests changed.")
```
simonlindholm: this seems redundant given all the individual printouts

run_tests.py
```diff
@@ -0,0 +1,125 @@
...
+    changed = final_contents != original_contents
+    if changed:
+        logging.info(
+            "\n".join(
+                [
+                    f"Output of {asm_file_path} changed!",
+                    "PREVIOUS:",
+                    original_contents,
+                    "NEW:",
+                    final_contents,
```
simonlindholm: (I wonder if it will hurt to not have diffs here)

matt-kempster: My motivation here was that if you accidentally run with `--overwrite`, you'd be able to use your terminal scrollback history to recover any lost data without having to muck with git (since perhaps you didn't commit your ongoing work).

matt-kempster:
Performance (from cold cache):
```
real    0m5.931s
user    0m0.313s
sys     0m0.375s
```
Performance (hot cache, i.e. immediately rerunning the tests):
```
real    0m0.507s
user    0m0.297s
sys     0m0.141s
```
So it takes "a couple seconds" to run all the tests. Pretty good IMO.

run_tests.py
matt-kempster: Yeah, I'd say "run `./run_tests.py` throughout development, monitoring changes per your expectations, then run `./run_tests.py --overwrite` one final time just prior to committing."

The advantage of having `--overwrite` off by default is to recover from potential minuscule regressions via terminal scrollback history, e.g. if you introduce a bug at the very end of development and aren't sure exactly what caused it. You could use your scrollback history and the good test output to try to diagnose what happened.

Do you think we should add a `./run_tests.py --overwrite` rule in our `.pre-commit` config?

run_tests.py
matt-kempster: Our code-output pipeline only goes to stdout at the moment, right? Redirecting stderr would be good for debugging crashes, though, yeah.

tests/add_test.py
matt-kempster: yeah. Was trying to decide between:
 - Adding an `--update` cli flag
 - Detecting `orig.c` as the filename and/or adding tests that are already in `end_to_end/SOME_DIR/`.

tests/add_test.py
matt-kempster: Pretty much `any-dir/<name>.c`. Should probably write this expectation explicitly.

simonlindholm: I think requiring `tests/<name>/orig.c` might be preferable, since:
a) it makes add and update workflows closer
b) it means you don't have to create a temporary file and then remove it to add a new test
c) if there are two copies of the same file floating around (both the original and orig.c), you risk accidentally modifying the wrong file

matt-kempster:
Please sanity check that I got everything we discussed :)

PR #43, opened 2019-07-18
Add compatibility with n64split's func_XXXXXXXX: notation

onlymx13:
So you can use mips_to_c on functions in files obtained from n64split.

### DIFF
--- a/src/parse_file.py
+++ b/src/parse_file.py
@@ -156,6 +156,13 @@ def parse_file(f: typing.TextIO, options: Options) -> MIPSFile:
                         mips_file.new_jumptable_label(function_name)
                     else:
                         mips_file.new_function(function_name)
+                elif line.startswith('func'):
+                    # Other kind of function label.
+                    function_name: str = line[:-1]
+                    if re.match('L(_U_)?[0-9A-F]{8}', function_name):
+                        mips_file.new_jumptable_label(function_name)
+                    else:
+                        mips_file.new_function(function_name)
                 else:
                     # Instruction.
                     instr: Instruction = parse_instruction(line, emit_goto)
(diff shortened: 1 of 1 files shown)


src/parse_file.py
```diff
@@ -156,6 +156,13 @@ def parse_file(f: typing.TextIO, options: Options) -> MIPSFile:
                         mips_file.new_jumptable_label(function_name)
                     else:
                         mips_file.new_function(function_name)
+                elif line.startswith('func'):
+                    # Other kind of function label.
+                    function_name: str = line[:-1]
```
matt-kempster: I assume this `[:-1]` is to trim off the colon. Can we use `.rstrip()` here instead?

src/parse_file.py
```diff
@@ -156,6 +156,13 @@ def parse_file(f: typing.TextIO, options: Options) -> MIPSFile:
...
+                    if re.match('L(_U_)?[0-9A-F]{8}', function_name):
```
matt-kempster: Hmm, IMO too much hardcoding here. Maybe we should just take `L_` as a required suffix and leave it at that?

onlymx13: I just copied this from above, I'm not sure what this is checking for.

simonlindholm: I agree it's a bit much hard-coding, I'd be in favor of changing both occurrences to checking just that it starts with `L`. (Eventually we'll get rid of that magic as well, but that's a harder problem. `L` *is* well-established as meaning a local label, at least.)

simonlindholm: On second thought, this particular check is pretty silly. It checks whether the line stars with "L", after we have already established that it starts with "func".

onlymx13: You're right, should I just assume that the statement is always false and remove the check?

onlymx13: like what I now have?

onlymx13: like what I now have?

simonlindholm: yep, looks good! maybe add `and line.endswith(":")` to the if condition?

also, you have a `'` while our coding style says `"` -- running `black` will fix that.

PR #52, opened 2020-03-29
Parse C code and use it for type information

simonlindholm:
Kinda WIP, only uses it to determine register sets for function calls so far, and isn't all that well tested.


simonlindholm:
Lots of follow-up potential, but this is useful enough to merge at this point.

PR #63, opened 2020-04-15
Fix jtbl identification

Emill:


simonlindholm:
Thanks! (For the record the bug fixed here is that the `lui %hi(jtbl)` may be in a different basic block and not be picked up by the jump table detection code.)

PR #67, opened 2020-04-18
Emit assignments inside &&/|| chains as comma expressions

### LINKED ISSUE #64: && detection can drop statements
I was a case of this where it dropped an important statement of the form `sp830 = strchr(sp830, 0x2C)`. We should never skip emitting blocks with statements in them, better just to fail && detection in that case.

(See also #33.)

matt-kempster:
Fixes #64.


src/translate.py
```diff
@@ -2437,6 +2459,7 @@ def process_instr(instr: Instruction) -> None:
             process_instr(instr)
 
     if branch_condition is not None:
+        assert not isinstance(branch_condition, CommaConditionExpr)
```
simonlindholm, line 2462: this assert isn't really needed

src/translate.py
```diff
@@ -579,6 +579,27 @@ def __str__(self) -> str:
...
+        return []
+
+    def negated(self) -> "CommaConditionExpr":
+        return CommaConditionExpr(self.statements, self.condition.negated())
+
+    def __str__(self) -> str:
+        str_statements: List[str] = []
+        for statement in self.statements:
+            str_statements.append(str(statement).rstrip(";"))
+        comma_joined = ", ".join(str_statements)
```
simonlindholm, line 599: how about `", ".join(str(stmt).rstrip(";") for stmt in self.statements)`?

src/translate.py
```diff
@@ -579,6 +579,27 @@ def __str__(self) -> str:
...
+        return f"{comma_joined}, {self.condition}"
```
simonlindholm, line 600: add parentheses around this, precedence will be wrong otherwise

tests/end_to_end/andor_assignment/irix-o2-out.c
```diff
@@ -0,0 +1,16 @@
+s32 test(s32 arg0, s32 arg1, s32 arg2, s32 arg3)
+{
+    s32 sp1C;
+    s32 temp_a0;
+    s32 phi_v1;
+
+    if (((((arg0 + arg1) != 0) || temp_a0 = arg1 + arg2, (temp_a0 != 0)) || sp1C = 0, (func_00400090(temp_a0) != 0)) || phi_v1 = sp1C, (arg3 != 0))
```
simonlindholm, line 7: this is one terrifying conditional... but it checks out

src/translate.py
```diff
@@ -530,7 +530,7 @@ def f64(left: "Expression", op: str, right: "Expression") -> "BinaryOp":
     def is_boolean(self) -> bool:
         return self.op in ["==", "!=", ">", "<", ">=", "<="]
 
-    def negated(self) -> "Condition":
+    def negated(self) -> Union["BinaryOp", "UnaryOp"]:
```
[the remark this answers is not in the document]
matt-kempster: I did this because there were lots of mypy errors in the callers of `.negated()` otherwise. I will try again but I'm pretty sure this was the simplest solution, and it's also in accordance with [the robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) (conservative in what you send, liberal in what you accept).

src/translate.py
matt-kempster: You're probably right, let me see what happens when I get rid of it

tests/end_to_end/andor_assignment/irix-o2-out.c
matt-kempster: yep...

PR #71, opened 2020-04-20
Separate logging from printing output

matt-kempster:
We use `print()` throughout the codebase to do two different things: log information for the reader, and print the actual desired output. This PR uses the python `logging` module to separate the former from the latter.


matt-kempster:
Closing this after some offline discussion.

PR #76, opened 2020-04-26
Detect and reroll unrolled do-while loops

matt-kempster:
Don't merge this yet - just putting some WIP work up.


matt-kempster:
This is in a stable state. The scope doesn't need to get much bigger for this PR; fixing &&/|| detection and more complex do-while detection can be in different PRs.

simonlindholm (commented):
review comments on flow graph munging, haven't looked at if_statements yet

src/flow_graph.py
```diff
@@ -507,22 +508,51 @@ def process(item: Union[Instruction, Label]) -> None:
...
+
+
 def is_loop_edge(node: "Node", edge: "Node") -> bool:
     # Loops are represented by backwards jumps.
     return edge.block.index <= node.block.index
 
 
+# TODO: Nodes need to be frozen.
+# We use dictionaries and sets of nodes throughout the codebase,
+# and it would be helpful to define __eq__ using `self.block.index`.
```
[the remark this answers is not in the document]
matt-kempster: I think you're right. However, I think using `is` the way I've been doing is kind of a hack - it feels wrong to be able to use `is` and not `==`, you know? I think it's likely that a Node should equal another Node if _something_ about it is equivalent - be it the `Block`, or the `block.index`... any ideas?

src/flow_graph_munge.py
```diff
@@ -0,0 +1,118 @@
...
+    if not modify_node_1_instructions(node_1.block.instructions):
+        return False
+
+    new_node_1 = node_1.to_basic_node(
+        successor=node_2  # node_2 doesn't know it's a parent yet
+    )
+    replace_node(flow_graph, node_1, new_node_1)  # now it does
+    remove_node(flow_graph, node_4, node_7)
+    remove_node(flow_graph, node_5, node_7)
+    remove_node(flow_graph, node_6, node_7)  # TODO: assert didn't execute anything?
```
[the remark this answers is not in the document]
matt-kempster: Can you clarify what you mean by "self-contained"? I'm not sure what cases the pattern-matching above this could miss.

matt-kempster (commented):
Agreed, will address everything in upcoming commits. Have a couple questions in the mean time.

matt-kempster:
So far this diff comprises:
 - `reroll.py`, which needs to be refactored so that it doesn't suck
   - I have some local code which can help this, namely regarding replacing graphs with subgraphs in an automated way
 - changes to `translate.py` that are ostensibly all bugs and need to be diagnosed and fixed
 - adding `--no-reroll` as a CLI arg
 - adding `replace_any_children()` to `BaseNode`: is this acceptable?
 - adding a test that doesn't really need to be there (I should remove it)

src/flow_graph.py
```diff
@@ -1322,7 +1359,10 @@ def is_reducible(self) -> bool:
             for s in n.children():
                 if s.loop is not None and n in s.loop.backedges:
                     continue
-                incoming_edges[s].remove(n)
+                try:
+                    incoming_edges[s].remove(n)
+                except KeyError:
+                    break
```
zbanks, line 1365: At first glance, I think the `break` should be `continue` (if it's OK to have missing edges) or `return False` (if it's not OK) -- otherwise, right now the behavior seems dependent on the order of the children in the list?

I think the fix here might actually be to iterate over `set(children)` instead. If `n` isn't in `incoming_edges[s]`, then I think that means `s.parents` was not constructed correctly? (Sorry that this might be a pre-existing bug)

```python
            for s in set(n.children()):
                if s.loop is not None and n in s.loop.backedges:
                    continue
                incoming_edges[s].remove(n)
```

src/translate.py
```diff
@@ -3032,11 +3033,17 @@ def assign_phis(used_phis: List[PhiExpr], stack_info: StackInfo) -> None:
...
+            i += 1
+            continue
         exprs = []
         for node in phi.node.parents:
             block_info = node.block.block_info
-            assert isinstance(block_info, BlockInfo)
+            # assert isinstance(block_info, BlockInfo)
+            if not block_info:
+                i += 1
+                continue
```
zbanks, line 3046: This `continue`s the `for node in phi.node.parents`, but the `i += 1` makes it look like you want to continue the `while i < len(used_phis)` loop? 

Is this needed? I'd be kind of worried if we have non-terminal blocks without `block_info` -- I think even with re-writing, `block_info` can be preserved?

(also remove commented-out asserts)

src/loop_rerolling.py
```diff
@@ -0,0 +1,116 @@
...
+        instructions[instructions.index(andi)] = move
+        return True
+
+    if not modify_node_1_instructions(node_1.block.instructions):
+        return False
+
+    new_node_1 = node_1.to_basic_node(
+        successor=node_2  # node_2 doesn't know it's a parent yet
+    )
+    replace_node(flow_graph, node_1, new_node_1)  # now it does
```
zbanks, line 94: I'm guessing these comments were from your old branch - and might not be needed now that `parents` is computed later?

src/loop_rerolling.py
```diff
@@ -0,0 +1,116 @@
...
+    while changed:
+        changed = False
+        for node in flow_graph.nodes:
+            if not isinstance(node, ConditionalNode):
+                continue
+            changed = reroll_loop(flow_graph, node)
+            if changed:
+                break
+    compute_relations(flow_graph.nodes)
+    return flow_graph
```
zbanks, line 116: This function modifies `flow_graph` in-place, so I don't think it needs to return anything.

zbanks (changes requested):
This is great - I really like how it's relatively small/readable it feels for what it's doing.

src/translate.py
matt-kempster: Yeah, the `block_info` issues are top on my list to debug. I'm not really sure why there are nodes that are lacking `block_info`...

src/flow_graph.py
matt-kempster: Does that mean `n.children()` might have repeats? That makes sense actually, and is probably why my extremely old version of this PR didn't need this fix (I changed `children()` to a `Set`.)

src/flow_graph.py
```diff
@@ -858,6 +883,15 @@ def __str__(self) -> str:
...
     cases: List["Node"] = attr.ib()
 
+    def replace_any_children(self, replace_this: "Node", with_this: "Node") -> None:
+        new_cases: List["Node"] = []
+        for case in self.cases:
+            if case is replace_this:
+                new_cases.append(with_this)
+            else:
+                new_cases.append(case)
+        self.cases = new_cases
```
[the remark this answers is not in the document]
matt-kempster: Hmm, scary... I never like to modify lists as I loop through them. Seems safe enough if you're using `enumerate()` though...

src/flow_graph.py
zbanks: Oh wait, I'm wrong, my bad! 

`n.children()` *shouldn't* have repeats: the implementations for that fn in `SwitchNode` & `ConditionalNode` explicitly deduplicate the children.

I can take a look and figure out what's going on here.

src/translate.py
zbanks: I put back the two `assert isinstance(..., BlockInfo)` checks + the `ValueError` except in `flow_graph.py` for `incoming_edges`, then re-ran mips2c against all of OOT, MM, & PM -- and there was no difference. Maybe some other refactor means that those cases no longer come up? (Seems like a good thing!)

src/loop_rerolling.py
```diff
@@ -0,0 +1,116 @@
...
+def reroll_loop(flow_graph: FlowGraph, start: ConditionalNode) -> bool:
+    node_1 = start.fallthrough_edge
+    node_7 = start.conditional_edge
+
+    if not isinstance(node_1, ConditionalNode):
+        return False
+    node_2 = node_1.fallthrough_edge
+    node_5 = node_1.conditional_edge
+
+    if not isinstance(node_2, BasicNode):
```
simonlindholm, line 40: I think we should also check that these nodes have the parents we expect. I feel like this pass has a large potential for misdetection and changing the asm into something non-equivalent, so we should be as careful as we can in the pattern-matching.

src/loop_rerolling.py
```diff
@@ -0,0 +1,116 @@
...
+    replace_node_references(flow_graph, to_delete, new_child)
+
+
+def replace_node(flow_graph: FlowGraph, replace_this: Node, with_this: Node) -> None:
+    replacement_index = flow_graph.nodes.index(replace_this)
+    flow_graph.nodes[replacement_index] = with_this
+    replace_node_references(flow_graph, replace_this, with_this)
+
+
+def reroll_loop(flow_graph: FlowGraph, start: ConditionalNode) -> bool:
```
simonlindholm, line 31: Add a comment with the structure/asm we're trying to pattern-match here (or refer to a test?).

zbanks: It'd be nice to include a `dot` representation of what's going on:

```dot
    digraph before {
        start -> node_1 [label=F, color=blue];
        start -> node_7 [label=T, color=red];
        node_1 -> node_2 [label=F, color=blue];
        node_1 -> node_5 [label=T, color=red];
        node_2 -> node_3;
        node_3 -> node_4 [label=F, color=blue];
        node_3 -> node_3 [label=T, color=red];
        node_4 -> node_5 [label=F, color=blue];
        node_4 -> node_7 [label=T, color=red];
        node_5 -> node_6;
        node_6 -> node_7 [label=F, color=blue];
        node_6 -> node_6 [label=T, color=red];
    }

    digraph after {
        start -> new_node_1 [label=F, color=blue];
        start -> node_7 [label=T, color=red];
        new_node_1 -> node_2 [label=F, color=blue];
        new_node_1 -> node_7 [label=T, color=red];
        node_2 -> node_3;
        node_3 -> node_7 [label=F, color=blue];
        node_3 -> node_3 [label=T, color=red];
    }
```
And maybe a link to an online render, like https://bit.ly/3A2MEJt

matt-kempster: This is actually nearly exactly what the new reroll.py does -- stay tuned

src/loop_rerolling.py
```diff
@@ -0,0 +1,138 @@
...
+                    isinstance(node, ConditionalNode)
+                    and node.conditional_edge is flow_graph.nodes[offset + conditional]
+                    and node.fallthrough_edge is flow_graph.nodes[offset + fallthrough]
+                ):
+                    return None
+        except IndexError:
+            return None
+
+    # TODO: Check that the subgraph is self-contained except for the entry
+    # and exit nodes, which themselves should be somehow "designated" using an enum.
```
zbanks, line 73: I think the following check is sufficient, without requiring you to explicitly mark an exit node in the pattern? (I think an exit node is any node that is not in `pattern.keys()` but is in `pattern.values()`? And the start node is `0`?) 
I think it can be up to the caller to check `start.parents`: I don't think the start's parents matter in this case.

```python
# pattern_nodes could be the `tuple(flow_graph.nodes[i] for i in sorted(all_nodes_in_pattern))` return value I suggest later
for node in pattern_nodes:
    if node is not start and not all(p in pattern_nodes for p in node.parents):
        return None
```

dbalatoni13:
I wonder if we could revive this

PR #77, opened 2020-04-27
Remove --allman for tests

matt-kempster:
I don't like --allman style braces.


matt-kempster:
I added `--allman` for loop_with_if, which I admit is a little hidden, but the code is still covered. It didn't feel right to add a duplicated test directory just for allman.

PR #83, opened 2020-05-09
Fix &&/|| detection crashes by improving their detection

matt-kempster:
This fixes several bugs in the code that would detect &&/|| compound if-statements, and improves the documentation of the functions involved.


matt-kempster:
Not quite done: still need to improve the docstring.

tests/end_to_end/andor_assignment/orig.c
```diff
@@ -13,6 +13,20 @@ int test(int a, int b, int c, int d) {
...
+    if (var1 && var2 && ((var1 += var2), (var2 = foo(var1))) && d) {
+        while (ret < 5) {
+            ret += 1;
+            ret *= 2;
+        }
+        ret += 5;
+    }
+
     return ret;
 }
```
simonlindholm, line 32: can you add another test for mixed &&/||? output won't be perfect ofc, but we need to make sure we don't crash or do something wildly wrong

src/if_statements.py
```diff
@@ -353,55 +288,208 @@ def immediate_postdominator(context: Context, start: Node, end: Node) -> Node:
...
-        print(
-            "Warning: confusing control flow, output may have incorrect && "
-            "and || detection. Run with --no-andor to disable detection and "
-            "print gotos instead.\n"
+    if_block_info = start.block.block_info
+    assert isinstance(if_block_info, BlockInfo)
+    assert if_block_info.branch_condition is not None
+
+    # If one of the output edges is the end, it's a "fake" if-statement. That
+    # is, it actually just resides one indentation level above the start node.
```
simonlindholm, line 314: This comment is now even more confusing than before, since the `if start.conditional_edge == end` check is gone

src/if_statements.py
```diff
@@ -353,55 +288,208 @@ def immediate_postdominator(context: Context, start: Node, end: Node) -> Node:
     return postdominators[0]
 
 
-def count_non_postdominated_parents(
-    context: Context, child: Node, curr_end: Node
-) -> int:
+def is_and_statement(next_node: Node, bottom: Node) -> bool:
+    return (
+        not isinstance(next_node, ConditionalNode)
+        or next_node not in bottom.parents
```
simonlindholm, line 294: I think `(next_node.conditional_edge is bottom or next_node.fallthrough_edge is bottom)` is clearer

src/if_statements.py
```diff
@@ -353,55 +288,208 @@ def immediate_postdominator(context: Context, start: Node, end: Node) -> Node:
...
+        # A strange edge-case of our pattern-matching technology:
+        # self-loops match the pattern. Avoiding that...
+        or next_node.is_loop()
```
simonlindholm, line 297: can you explain this case further? if this method is inlined into get_full_if_condition the comment may find a more natural place to live

src/if_statements.py
```diff
@@ -353,55 +288,208 @@ def immediate_postdominator(context: Context, start: Node, end: Node) -> Node:
...
+                    conditions + [next_node_condition], "||", only_negate_last=True
+                ),
+                indent,
+                context.options.coding_style,
+                if_body=build_flowgraph_between(context, bottom, end, indent + 4),
+                # The else-body is wherever the code jumps to instead of the
+                # fallthrough (i.e. if-body).
+                else_body=build_flowgraph_between(
+                    context, next_node.conditional_edge, end, indent + 4
+                ),
```
simonlindholm, line 486: the order in which build_flowgraph_between is called matters in rare cases, because of goto emission. Use local variables for the results of the two calls. Also, this else_body should sometimes be None, just like for &&

PR #93, opened 2021-05-21
Attach TypeMap to CType-related Types

zbanks:
I was interested in improving some of the type detection stuff, and it seemed like a prereq for that was having the `TypeMap` available to the `Type.unify()` function.

There are two calls on the `format` codepath that called `unify`, and it was really hard to get the typemap there. My understanding of the code is that the types should already be resolved at that point -- the `unify` function is mostly a check if the two types are the same. I added a `Type.is_unified()` (`is_equivalent`?) method to start to try to refactor it even though it's a bit ugly.

As mentioned on the Discord, this won't handle all cases: the typemap still isn't available everywhere. I'll need to take another pass at this.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -1916,7 +1916,7 @@ def handle_addi_real(
         var = stack_info.get_stack_var(imm.value, store=False)
         if isinstance(var, LocalVar):
             stack_info.add_local_var(var)
-        return AddressOf(var, type=Type.ptr(var.type))
+        return AddressOf(var, type=Type.ptr(var.type, stack_info.typemap))
     else:
         return add_imm(source, imm, stack_info)
 
@@ -2279,7 +2279,7 @@ def array_access_from_add(
             target_size=None,
             field_name=sub_field_name,
             stack_info=stack_info,
-            type=Type.ptr(elem_type),
+            type=Type.ptr(elem_type, typemap),
         )
         offset -= sub_offset
         target_type = type_from_ctype(elem_type, typemap)
--- a/src/translate.py
+++ b/src/translate.py
@@ -1916,7 +1916,7 @@ def handle_addi_real(
         var = stack_info.get_stack_var(imm.value, store=False)
         if isinstance(var, LocalVar):
             stack_info.add_local_var(var)
-        return AddressOf(var, type=Type.ptr(var.type, stack_info.typemap))
+        return AddressOf(var, type=Type.ptr(var.type))
     else:
         return add_imm(source, imm, stack_info)
 
@@ -2279,7 +2279,7 @@ def array_access_from_add(
             target_size=None,
             field_name=sub_field_name,
             stack_info=stack_info,
-            type=Type.ptr(elem_type, typemap),
+            type=Type.cptr(elem_type, typemap),
         )
         offset -= sub_offset
         target_type = type_from_ctype(elem_type, typemap)
--- a/src/types.py
+++ b/src/types.py
@@ -30,8 +30,8 @@ class Type:
     K_INT = 1
     K_PTR = 2
     K_FLOAT = 4
-    K_INTPTR = 3
-    K_ANY = 7
+    K_INTPTR = K_INT | K_PTR
+    K_ANY = K_INT | K_PTR | K_FLOAT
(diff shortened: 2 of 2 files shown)


src/translate.py
```diff
@@ -502,7 +504,7 @@ class Expression(abc.ABC):
     def dependencies(self) -> List["Expression"]:
         ...
 
-    def use(self) -> None:
+    def use(self, stack_info: StackInfo) -> None:
```
zbanks, line 507: The biggest style question I had is if I should *just* pass the `typemap` to `use(...)`, or pass the whole `stack_info`. 
Not sure how gross/abstraction-breaking this is.

src/translate.py
```diff
@@ -170,52 +170,54 @@ def indent(self, indent: int, line: str) -> str:
...
+    if expr.type.unify(type, typemap):
         if not silent and not isinstance(expr, Literal):
             return Cast(expr=expr, reinterpret=True, silent=False, type=type)
         return expr
     return Cast(expr=expr, reinterpret=True, silent=False, type=type)
 
 
 def as_f32(expr: "Expression") -> "Expression":
-    return as_type(expr, Type.f32(), True)
+    return as_type(expr, Type.f32(), None, True)
```
simonlindholm, line 184: having to pass None here without a lot more messy plumbing seems like a downside of this approach. It may well be fine in practice, especially since we'll unify on use(), but passing typemaps consistently would be nicer

src/translate.py
simonlindholm: passing stack_info to use() makes sense to me. Many expressions already include a reference to it, so in principle I think it would also be fine to put it within the Expression base class, but I guess that gets too annoying plumbing-wise

src/translate.py
```diff
@@ -809,16 +811,16 @@ class Cast(Expression):
...
-        super().use()
+        self.expr.type.unify(self.type, stack_info.typemap)
+        super().use(stack_info)
 
     def needed_for_store(self) -> bool:
         if not self.reinterpret:
             # int <-> float casts should be emitted even for stores.
             return True
-        if not self.expr.type.unify(self.type):
+        if not self.expr.type.is_unified(self.type):
```
simonlindholm, line 823: as an alternative we could have use() store a property that tells whether unification succeeded

simonlindholm (commented):
Looks good, generally, but I'm not sure this is the best approach. Seems like it would be simpler and cover more cases to store a typemap in Type whenever it refers to a CType.

src/translate.py
zbanks: Oh that's a way better nicer, good call.

zbanks:
Totally redid this based on our Discord conversation, I'm way happier with this result. 

I added some `assert`s which make the assumption that if there are any `CType`s involved, then `typemap` should not be `None`. I think this is valid, but I wanted to check?

(`./run_tests.py` passes; ran black & mypy)

src/types.py
```diff
@@ -178,8 +185,14 @@ def intptr32() -> "Type":
         return Type(kind=Type.K_INTPTR, size=32, sign=Type.ANY_SIGN)
 
     @staticmethod
-    def ptr(type: Optional[Union["Type", CType]] = None) -> "Type":
-        return Type(kind=Type.K_PTR, size=32, sign=Type.ANY_SIGN, ptr_to=type)
+    def ptr(
+        type: Optional[Union["Type", CType]] = None, typemap: Optional[TypeMap] = None
+    ) -> "Type":
```
simonlindholm, line 190: I suspect this might nicely split into two methods `ptr(type: Optional[Type] = None)` and `cptr(type: CType, typemap: TypeMap)`, which gets rid of the need for the assert

simonlindholm (approved):
Nice, this ended up a lot cleaner indeed!

simonlindholm:
> I added some asserts which make the assumption that if there are any CTypes involved, then typemap should not be None. I think this is valid, but I wanted to check?

Yeah, this is a valid assumption

src/types.py
zbanks: This is great, it helps tighten up the type bounds on the args to `Type.ptr`/`Type.cptr`

PR #94, opened 2021-05-23
Use TypeMap information to unify CTypes & non-CTypes

### LINKED ISSUE #79: Support type unification between C types and non-C types
```py
            elif not isinstance(x.ptr_to, Type) and not isinstance(y.ptr_to, Type):
                # TODO: deep resolve_typedefs (needs a typemap)
                if not equal_types(x.ptr_to, y.ptr_to):
                    return False
            else:
                # TODO: unify Type and CType (needs a typemap)
                return False
```

We need some additional context for this; maybe store typemap in Type or CType? Or use some "thread local state"-like pattern, since we have only a single typemap that we should use anyway?

zbanks:
I think this closes #79 .

The number of lines is about the same, but now there are only two main cases to handle: two `CType`s, and at least one `Type`.

I also updated `add_test.py` to support the recompiled IDO compiler. The `MIPS_CC` is the path to the `cc` command.

If I run the tests using the `origin/master` code, there's the following diff, from where it was unable to unify the type of `arg0`.
```
[INFO] Output of /home/zbanks/mips_to_c/tests/end_to_end/typedef/irix-o2.s changed! Diff:
--- 

+++ 

@@ -3,6 +3,6 @@

     s32 sp18;
 
     sp18 = func_00400090(arg0, arg1);
-    sp1C = func_00400090(arg0, &arg0);
+    sp1C = func_00400090(arg0, (int_t *) &arg0);
     func_00400090(*arg1, arg1);
 }
```

- All of the existing tests pass.
- Ran `mypy`, `black`


src/types.py
```diff
@@ -70,32 +70,27 @@ def unify(self, other: "Type") -> bool:
...
-                    typemap = y.typemap
-                elif isinstance(y.ptr_to, Type) and not isinstance(x.ptr_to, Type):
-                    type = y.ptr_to
-                    ctype = x.ptr_to
-                    typemap = x.typemap
+                if isinstance(x.ptr_to, Type):
+                    x_type = x.ptr_to
+                else:
+                    assert x.typemap is not None
+                    x_type = type_from_ctype(x.ptr_to, x.typemap)
```
simonlindholm, line 86: `type_from_ctype` is lossy, because there's no Type corresponding to a C struct for instance (perhaps a design mistake, but it *is* convenient to keep up the pretense that all local variables/expressions/etc. are register sized). I think maybe if `type_from_ctype(x.ptr_to, x.typemap)` returns Type.any() and `y.ptr_to` is not Type.any() we should consider returning False here?

tests/end_to_end/typedef/context.c
```diff
@@ -0,0 +1,2 @@
+typedef int int_t;
+int_t func_00400090(int x, int_t *y);
```
simonlindholm, line 2: you need a -flags.txt file for this file to be used

(one thing I've considered is running all tests both with and without orig.c as context for some cheap extra coverage)

zbanks: Good catch - I figured this out late, and didn't `git add` it.

src/types.py
zbanks: I think trying to unifying them is still correct?

We only use `type_from_ctype` when one of `x.ptr_to`/`y.ptr_to` is a `CType`, and when one is a `Type`. I think the only lossy case would be when *both* are `CType`s?

Let's say `x.ptr_to` is the `CType` & `y.ptr_to` is the `Type`. We lose information if `type_from_ctype(x.ptr_to, ...) == Type.any()`, and that means `x` is either a `void *` or a struct/union pointer.

If `y.ptr_to` is `Type.any()`, then that means `y` is a `void *`, so it's valid to unify it with the struct pointer. On the other hand, if `y.ptr_to` is an `float` or something, it won't be able to unify with `x.ptr_to`.

simonlindholm: The case I'm worried about is if x.ptr_to is a Struct and y.ptr_to is say a Type.f32(), then unification will currently succeed which seems sketchy. Claiming `void*` = `float*` also feels suspect, being non-transitive.

If x.ptr_to is Type.any() a reasonable unification would be to set x.ptr_to = y.ptr_to, no need to do type_from_ctype. Otherwise I'm dubious about the struct/void cases. But it's 3 AM so I may be missing something :)

zbanks: I think I see the issue, and I refactored this to avoid unifying with `Type.any()`'s.

However, I think in both the original code and my PR, the unify implementation has `void * == float *`. 

It seems like `void *` could be represented two ways: `Type.ptr(None)` or `Type.ptr(Type.any())`. 

In the first case, `x.ptr_to` would just be `None` so this whole `if` block I'm modifying wouldn't be hit, and the types would just unify taking `y.ptr_to`.
In the second case, with `x.ptr_to = Type.any()`, that type could unify with `y.ptr_to = Type.f32()`.

simonlindholm (approved):

PR #95, opened 2021-05-29
Add `--project` flag to run tests against all mm/oot asm

zbanks:
This was discussed briefly on the Discord yesterday. Here are some notes about this PR:

- I'm calling them "project tests" -- not super happy with that name, but had trouble thinking of a better one.
- This PR doesn't commit the results of the tests. There was some brief discussion on the discord about how best to design this so that we can keep this repo isolated from the decomp project repos. 
    - I'm personally excited enough to get this code in as-is, even if it's a bit trickier to work with. Even though the "baseline" output `.c` files are not checked in, I can run the tests locally before making changes to populate the `tests/` directory with the baseline files.
- The project tests use `ctx.c` if it exists in the project base directory. This also makes the tests *a lot* slower.
    - I think a huge speedup could come from building the `typemap` once (from parsing the context), then re-using it between tests. I believe it isn't modified once the parsing is done. Unfortunately this would require some slightly weird plumbing: right now arguments are passed through argparse to the decompiler API.
- For the project tests, if there's an error for one function, it includes the error message in the output `.c` and then continues to try to decompile the other functions.
- I made 2 bugfixes. I think they're pretty benign but I put them in a separate PR if that's preferred. Now all MM `.asm` files can be processed without error! (except `boot/exceptasm.asm`, which seems hand-written)
    - I added ignoring `nop`s at the start of `.text` sections, which was needed for a few OOT files.
    - Skip duplicate consecutive labels. There are some parts of the asm that have `glabel X; .X:`.
- All `.py` files formatted with `black` 21.5b0 and checked with `mypy` 0.812.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -8,7 +8,7 @@ import re
 import shlex
 import sys
 from pathlib import Path
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 from src.main import parse_flags
 from src.main import run as decompile
@@ -139,6 +139,7 @@ def run_e2e_test(
 def run_project_tests(
     base_dir: Path,
     output_dir: Path,
+    use_context: bool,
     should_overwrite: bool,
     filter_regex: Optional[str],
     name_prefix: str,
@@ -147,6 +148,12 @@ def run_project_tests(
     ret = True
     asm_dir = base_dir / "asm"
     context_file = base_dir / "ctx.c"
+    if use_context and not context_file.exists():
+        logging.error(
+            f"{name_prefix} tests require context file, but {context_file} does not exist"
+        )
+        return False
+
     for asm_file in asm_dir.rglob("*"):
         if asm_file.suffix not in (".asm", ".s"):
             continue
@@ -164,9 +171,8 @@ def run_project_tests(
         ):
             continue
 
-        has_context = context_file.exists()
         flags = []
-        if has_context:
+        if use_context:
             flags.extend(["--context", str(context_file)])
 
         # Guess the name of .rodata file(s) for the MM decomp project
@@ -188,7 +194,7 @@ def run_project_tests(
                 flags.extend(["--rodata", str(f)])
 
         test_path = asm_file.relative_to(asm_dir)
-        name = f"{name_prefix}{'_ctx' if has_context else ''}:{test_path}"
+        name = f"{name_prefix}{'_ctx' if use_context else ''}:{test_path}"
         if filter_regex is not None and not re.search(filter_regex, name):
             continue
         if coverage:
@@ -209,7 +215,7 @@ def run_project_tests(
 
 
 def main(
-    project_dirs: List[Path],
+    project_dirs: List[Tuple[Path, bool]],
     should_overwrite: bool,
     filter_regex: Optional[str],
     coverage: Any,
@@ -222,11 +228,17 @@ def main(
         ):
             ret = 1
 
-    for project_dir in project_dirs:
+    for project_dir, use_context in project_dirs:
         name = project_dir.name
         output_dir = Path(__file__).parent / "tests" / "project" / name
         if not run_project_tests(
-            project_dir, output_dir, should_overwrite, filter_regex, name, coverage
+            project_dir,
+            output_dir,
+            use_context,
+            should_overwrite,
+            filter_regex,
+            name,
+            coverage,
         ):
             ret = 1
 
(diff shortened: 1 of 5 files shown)


zbanks:
Hmm - I discovered a bad wrinkle with the MM repo.
The MM build system has two relevant steps here:
1. `./tools/disasm.py -d ./asm -l ./tables/files.txt -f ./tables/functions.txt -o ./tables/objects.txt -v ./tables/variables.txt -v ./tables/vrom_variables.txt`
2. `./tools/split_asm.py asm/$name.asm asm/non_matchings/$name` (repeated for each code/boot/overlay)

Step 1 generates the .asm files, then step 2 mostly populates the `non_matching/` folder with the asm split per function. *However*, step 2 *also* overwrites the `rodata.asm` files. Some rodata is moved out of the `rodata.asm` files and into the per-function asms, like jump tables and constants. I think this is done so that the `rodata.asm` files can be compiled raw into the `.rodata` sections of the `.o`: if these constants were left in rodata, they'd be duplicated in the `.o`. 

This is a problem because the constants that are moved out of rodata are now gone. It makes `split_asm.py` non-idempotent. Until I can submit a PR to fix this in MM (or MM gets a new build system) -- it makes running these tests slightly harder. I'd recommend manually running step 1 instead of using `make` so that the full rodata is available.

run_tests.py
```diff
@@ -80,39 +88,146 @@ def decompile_and_compare(
...
+            or asm_name.startswith("boot_rodata")
+            or asm_name.endswith("_data.asm")
+            or asm_name.endswith("_rodata.asm")
+        ):
+            continue
+
+        has_context = context_file.exists()
+        flags = []
+        if has_context:
+            flags.extend(["--context", str(context_file)])
```
simonlindholm, line 170: yeah this repeated parsing seems disastrous for performance. possible suggestions:
- a `--cache-context-for-tests` flag with `help=argparse.SUPPRESS` that globally caches the typemap if passed (with the context filename as key, or clearing the cache from here between tests; it's okay if it's ugly)
- passing an optional `typemap` through to main.run

run_tests.py
```diff
@@ -80,39 +88,146 @@ def decompile_and_compare(
...
+            asm_name.startswith("code_data")
+            or asm_name.startswith("code_rodata")
+            or asm_name.startswith("boot_data")
+            or asm_name.startswith("boot_rodata")
+            or asm_name.endswith("_data.asm")
+            or asm_name.endswith("_rodata.asm")
+        ):
+            continue
+
+        has_context = context_file.exists()
```
simonlindholm, line 167: maybe worth basing this on a command line parameter since both behaviors are interesting to test and they use different expectation filenames

simonlindholm (commented):
Nice!

run_tests.py
zbanks: I had another idea recently - put the caching in `build_typemap`: if you call `build_typemap` twice in a row with the same string, it could return the cached `TypeMap`?

src/parse_file.py
```diff
@@ -164,7 +169,7 @@ def parse_ascii_directive(line: str, z: bool) -> bytes:
 
 
 def parse_file(f: typing.TextIO, options: Options) -> MIPSFile:
-    filename = f.name
+    filename = Path(f.name).name
```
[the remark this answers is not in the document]
zbanks: This was changed to remove the full filepath from the error message -- now it only prints the basename instead of the full path. This would make the test reproducible on machines with different paths. I could punt this to a later PR if you think keeping the full path is worth it.

Example:
```diff
-Found jr instruction at /home/zbanks/mm/asm/code/code_0x80161180.asm line 522, but the corresponding jump table is not provided.
+Found jr instruction at code_0x80161180.asm line 522, but the corresponding jump table is not provided.
```

run_tests.py
simonlindholm: I think it's probably worthwhile to avoid repeated file reads -- context files can be pretty big

(also I'm slightly unhappy about making this non-threadsafe outside of the "anything goes" invoked-from-test case, not that it would actually come up in practice)

src/parse_file.py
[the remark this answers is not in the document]
simonlindholm: ah, makes sense

run_tests.py
zbanks: I was thinking of just using [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache), which is threadsafe. 

The context files *are* pretty big (a few MB), I think in practice it's not that bad.

simonlindholm: can we lru_cache a function that takes in a filename and returns a typemap?

zbanks: Yeah - we could do that. I was a bit afraid of having the contents of the file change though. (ex: I'm not sure how the website is implemented)

I tried `lru_cache(maxsize=4)` with a ~2MB context and it only takes ~3ms to read the file & load it from the cache. That seems worth it to me to avoid some potential cache invalidation issues?
```python
In [8]: tm = src.c_types.build_typemap(open("../mm/ctx.c").read())                                                             

In [9]: %timeit src.c_types.build_typemap(open("../mm/ctx.c").read())                                                          
2.73 ms ± 41.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

run_tests.py
zbanks: Yeah you bring up a great point. I was trying to keep the arguments simple, but keep the program flexible enough to run with partial context. 

I'm going to try to refactor this a bit.

run_tests.py
simonlindholm: Huh, that's surprisingly fast. I guess it would be fine then (if still maybe not personal preference). The website runs the script repeatedly through subprocess fwiw, without any persistence -- one reason being to deal gracefully with infinite loops.

simonlindholm (approved):

PR #97, opened 2021-05-30
Simplify `x = x op y` into `x op= y`

zbanks:
Some notes:

- There are two other places that can emit `lhs = rhs;` statements that I didn't change: `EvalOnceStmt` & `SetPhiStmt`. In these two cases, I don't think it's possible to have `lhs` be in the `rhs` expression. 
- I didn't simplify `x += 1;` into `x++;` to avoid having to deal with precedence rules. (`*x += 1;` is not equivalent to `*x++`).
    - IMO refactoring `x += 1;` by hand is still easier to than `x = x + 1;`
- It looks like `late_unwrap(...)` is completely safe to use inside `format(...)` functions. I think `late_unwrap(expr).format(fmt) == expr.format(fmt)` is an invariant.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -47,6 +47,7 @@ from .types import (
 )
 
 ASSOCIATIVE_OPS: Set[str] = {"+", "&&", "||", "&", "|", "^", "*"}
+COMPOUND_ASSIGNMENT_OPS: Set[str] = {"+", "-", "*", "/", "%", "&", "|", "^", "<<", ">>"}
 
 ARGUMENT_REGS: List[Register] = list(
     map(Register, ["a0", "a1", "a2", "a3", "f12", "f14"])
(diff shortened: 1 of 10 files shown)


src/translate.py
```diff
@@ -1299,13 +1300,22 @@ def should_write(self) -> bool:
...
-            isinstance(self.dest, StructAccess) and self.dest.late_has_known_type()
-        ) or isinstance(self.dest, (ArrayAccess, LocalVar, RegisterVar, SubroutineArg)):
+            isinstance(dest, StructAccess) and dest.late_has_known_type()
+        ) or isinstance(dest, (ArrayAccess, LocalVar, RegisterVar, SubroutineArg)):
             # Known destination; fine to elide some casts.
             source = elide_casts_for_store(source)
-        return f"{self.dest.format(fmt)} = {format_expr(source, fmt)};"
+        if isinstance(source, BinaryOp) and source.op in COMPOUND_ASSIGNMENT_OPS:
+            rhs = None
+            if late_unwrap(source.left) == dest:
```
simonlindholm, line 1312: could also check `source.left.format() == dest.format()` if this turns out to be too strong

simonlindholm (approved):

PR #98, opened 2021-06-01
Support GCC by allowing locals above saved registers

zbanks:
[Trello card](https://trello.com/c/0r3GPRGE/133-allow-using-subroutineargs-as-normal-sps) + some small discussion [on discord](https://discord.com/channels/688807550715560050/703351323445428306/849010792283111445).

- GCC & IDO arrange locals & saved regs differently within the stack frame.
    - GCC puts locals below the saved regs, IDO puts them above.
    - It's very easy to determine the boundary of the saved regs region in a single pass, so we were previously using that to distinguish between local variables and subroutine arguments by assuming that anything *below* the saved regs was a subroutine arg.
- The new heuristc for detecting subroutine argument boundary also improves output from IDO-compiled asm. It is now better at correctly guessing the number of arguments to functions with >4 args.
    - There are a few examples of this in the MM code, run without context 

Example improvement from MM (IDO) without context (`func_800BE680` indeed should take 8 args)
```diff
...
-    sp14 = arg0->unk264;
-    sp18 = arg0->unk25C;
-    sp1C = (?32) arg0->unk24F;
-    func_800BE680(arg1, arg0, arg0 + 0x2B0, 5, arg0->unk260);
-}
+    func_800BE680(arg1, arg0, arg0 + 0x2B0, 5, arg0->unk260, arg0->unk264, arg0->unk25C, (?32) arg0->unk24F);
+}
```

Example improvement from Paper Mario (GCC) without context. I've interleaved the diff lines to make it easier to compare. Also note that `sp10` is probably an 8-byte struct, and it's a separate issue that mips_to_c emits references to `sp14` (really `sp10.unk4`) without declaring it.
```diff
 void dma_load_string(u32 arg0, ? arg1) {
+    s32 sp10;
     s32 temp_a0;
     s32 temp_a0_2;
 
     temp_a0 = (arg0 >> 0xE) + 0x1B83000;
-    dma_copy(temp_a0, temp_a0 + 4, &subroutine_arg4);
+    dma_copy(temp_a0, temp_a0 + 4, &sp10);
-    temp_a0_2 = subroutine_arg4 + (((arg0 & 0xFFFF) * 4) + 0x1B83000);
+    temp_a0_2 = sp10 + (((arg0 & 0xFFFF) * 4) + 0x1B83000);
-    dma_copy(temp_a0_2, temp_a0_2 + 8, &subroutine_arg4);
+    dma_copy(temp_a0_2, temp_a0_2 + 8, &sp10);
-    dma_copy(subroutine_arg4 + 0x1B83000, subroutine_arg5 + 0x1B83000, arg1);
+    dma_copy(sp10 + 0x1B83000, sp14 + 0x1B83000, arg1);
 }
```

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -228,9 +228,10 @@ class StackInfo:
     is_leaf: bool = attr.ib(default=True)
     is_variadic: bool = attr.ib(default=False)
     uses_framepointer: bool = attr.ib(default=False)
-    local_vars_region_bottom: int = attr.ib(default=0)
+    subroutine_arg_top: Optional[int] = attr.ib(default=None)
     return_addr_location: int = attr.ib(default=0)
     callee_save_reg_locations: Dict[Register, int] = attr.ib(factory=dict)
+    callee_save_reg_region: Tuple[int, int] = attr.ib(default=(0, 0))
     unique_type_map: Dict[Any, "Type"] = attr.ib(factory=dict)
     local_vars: List["LocalVar"] = attr.ib(factory=list)
     temp_vars: List["EvalOnceStmt"] = attr.ib(factory=list)
@@ -250,16 +251,12 @@ class StackInfo:
     def in_subroutine_arg_region(self, location: int) -> bool:
         if self.is_leaf:
             return False
-        subroutine_arg_top = self.return_addr_location
-        if self.callee_save_reg_locations:
-            subroutine_arg_top = min(
-                subroutine_arg_top, min(self.callee_save_reg_locations.values())
-            )
+        assert self.subroutine_arg_top is not None
+        return location < self.subroutine_arg_top
 
-        return location < subroutine_arg_top
-
-    def in_local_var_region(self, location: int) -> bool:
-        return self.local_vars_region_bottom <= location < self.allocated_stack_size
+    def in_callee_save_reg_region(self, location: int) -> bool:
+        lower_bound, upper_bound = self.callee_save_reg_region
+        return lower_bound <= location < upper_bound
 
     def location_above_stack(self, location: int) -> bool:
         return location >= self.allocated_stack_size
@@ -336,19 +333,21 @@ class StackInfo:
         return False
 
     def get_stack_var(self, location: int, *, store: bool) -> "Expression":
-        if self.in_local_var_region(location):
-            return LocalVar(location, type=self.unique_type_for("stack", location))
-        elif self.location_above_stack(location):
+        # See `get_stack_info` for explanation
+        if self.location_above_stack(location):
             ret, arg = self.get_argument(location - self.allocated_stack_size)
             if not store:
                 self.add_argument(arg)
             return ret
         elif self.in_subroutine_arg_region(location):
             return SubroutineArg(location, type=Type.any())
-        else:
+        elif self.in_callee_save_reg_region(location):
             # Some annoying bookkeeping instruction. To avoid
             # further special-casing, just return whatever - it won't matter.
             return LocalVar(location, type=Type.any())
+        else:
+            # Local variable
+            return LocalVar(location, type=self.unique_type_for("stack", location))
 
     def maybe_get_register_var(self, reg: Register) -> Optional["RegisterVar"]:
         return self.reg_vars.get(reg)
@@ -387,7 +386,7 @@ class StackInfo:
                 f"Stack info for function {self.function.name}:",
                 f"Allocated stack size: {self.allocated_stack_size}",
                 f"Leaf? {self.is_leaf}",
-                f"Bottom of local vars region: {self.local_vars_region_bottom}",
+                f"Bounds of callee-saved vars region: {self.callee_save_reg_locations}",
                 f"Location of return addr: {self.return_addr_location}",
                 f"Locations of callee save registers: {self.callee_save_reg_locations}",
             ]
@@ -397,10 +396,26 @@ class StackInfo:
 def get_stack_info(
     function: Function, rodata: Rodata, start_node: Node, typemap: Optional[TypeMap]
 ) -> StackInfo:
+    def update_min(old_or_none: Optional[int], new: int) -> int:
+        if old_or_none is None:
+            return new
+        return min(old_or_none, new)
+
     info = StackInfo(function, rodata, typemap)
 
     # The goal here is to pick out special instructions that provide information
     # about this function's stack setup.
+    #
+    # IDO puts local variables *above* the saved registers on the stack, but
+    # GCC puts local variables *below* the saved registers.
+    # To support both, we explicitly determine both the upper & lower bounds of the
+    # saved registers. Then, we estimate the boundary of the subroutine arguments
+    # by finding the lowest stack offset that is loaded from or computed. (This
+    # assumes that the compiler will never reuse a section of stack for *both*
+    # a local variable *and* a subroutine argument.) Anything within the stack frame,
+    # but outside of these two regions, is considered a local variable.
+    callee_saved_words: List[int] = []
+    callee_saved_doubles: List[int] = []
     for inst in start_node.block.instructions:
         if not inst.args or not isinstance(inst.args[0], Register):
             continue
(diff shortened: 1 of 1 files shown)


zbanks:
Test diff output from running (e2e, MM (no ctx), OOT (no ctx), MM with ctx, OOT with ctx, papermario (no ctx)): https://gist.github.com/zbanks/4230905da934cd6c7e70b9301ef4e759

Overall looks good, no *new* `subroutine_arg`s added to mm/oot. Some still in papermario, but a big improvement.

This is the one change I found in MM that looks worse, from `ovl_En_Go_0x80A10FD0`. It looks like the second arg to `func_80A10FD0` is a 52-byte struct, so the entire struct is passed on the stack?

```diff
[INFO] Output of ../mm/asm/overlays/ovl_En_Go_0x80A10FD0.asm changed! Diff:
--- 

+++ 

@@ -1359,7 +1359,21 @@

     f32 sp4C;
     ? sp48;
     s16 sp46;
+    ?32 sp30;
+    f32 sp2C;
+    f32 sp28;
+    s32 sp24;
+    s32 sp20;
+    s32 sp1C;
+    s32 sp18;
+    s32 sp14;
+    s32 sp10;
+    s32 spC;
+    s32 sp8;
+    s32 sp4;
     f32 temp_f10;
+    s32 temp_a2;
+    s32 temp_a3;
 
     sp46 = (s16) (s32) (Rand_ZeroOne() * 360.0f * 182.0f);
     Math_Vec3f_Copy(&sp54, &D_801D15B0);
@@ -1368,7 +1382,21 @@

     temp_f10 = (Rand_ZeroOne() * 10.0f) + 4.0f;
     sp4C = temp_f10;
     sp4C = temp_f10 + arg0->unk88;
-    func_80A10FD0(arg0 + 0x7B8, subroutine_arg1, sp48.unk4, sp48.unk8, D_80A16698.unk0, D_80A16698.unk4, D_80A16698.unk8, D_801D15B0.unk0, D_801D15B0.unk4, D_801D15B0.unk8, 0.01f, 0.002f, 0x10);
+    sp4 = sp48.unk0;
+    temp_a2 = sp48.unk4;
+    sp8 = temp_a2;
+    temp_a3 = sp48.unk8;
+    spC = temp_a3;
+    sp10 = D_80A16698.unk0;
+    sp14 = D_80A16698.unk4;
+    sp18 = D_80A16698.unk8;
+    sp1C = D_801D15B0.unk0;
+    sp20 = D_801D15B0.unk4;
+    sp30 = 0x10;
+    sp28 = 0.01f;
+    sp2C = 0.002f;
+    sp24 = D_801D15B0.unk8;
+    func_80A10FD0(arg0 + 0x7B8, sp4, temp_a2, temp_a3);
 }
 
 s32 func_80A13B1C(void *arg0, void *arg1) {
```

src/translate.py
```diff
@@ -431,29 +446,78 @@ def get_stack_info(
...
+            info.callee_save_reg_locations[destination] = stack_offset
+            if inst.mnemonic == "sdc1":
+                callee_saved_doubles.append(stack_offset)
+            else:
+                callee_saved_words.append(stack_offset)
+        elif (
+            inst.mnemonic in ["lw", "lwc1", "ldc1"]
+            and isinstance(inst.args[1], AsmAddressMode)
+            and inst.args[1].rhs.register_name == "sp"
+        ):
```
simonlindholm, line 470: the surrounding loop is only over the first basic block, but we may want to run this on the entire function

src/translate.py
```diff
@@ -431,29 +446,78 @@ def get_stack_info(
...
+        callee_saved_words.sort()
+        callee_saved_doubles.sort()
+        bottom = callee_saved_words[0]
+
+        # Both GCC & IDO save registers in two blocks: one for word-sized registers,
+        # followed by one for doubles. The word-sized block is also padded to a
+        # multiple of 8 bytes. Check that there are no gaps in either of these blocks.
+        top = bottom
+        for offset in callee_saved_words:
+            assert offset == top, DecompFailure(
```
simonlindholm, line 495: this still raises an AssertionError, change to `if offset != top: raise DecompFailure(...)`

src/translate.py
```diff
@@ -431,29 +446,78 @@ def get_stack_info(
...
+                f"saved doubles: {callee_saved_doubles}, "
+                f"gap at: {offset} != {top}."
+            )
+            top += 4
+        # Padding: round up to multiple of 8
+        if (top % 8) == 4:
+            top += 4
+        for offset in callee_saved_doubles:
+            if offset < callee_saved_words[-1]:
+                # This offset is fake: it's a local variable; exclude it
```
simonlindholm, line 507: can this happen? I'd expect it not to because all saved registers must be saved before being used in any other way

zbanks: Good catch - I think when I added the `and destination not in info.callee_save_reg_locations` check, it removed the issue I was originally seeing.

...but when I went to verify that, I discovered that IDO actually puts the doubles *below* the words. So, I'll update this logic to work in either orientation.

src/translate.py
zbanks: Great point - running over the whole function fixes the `ovl_En_Go_0x80A10FD0` weirdness I commented about.

src/translate.py
```diff
@@ -459,60 +461,71 @@ def update_min(old_or_none: Optional[int], new: int) -> int:
...
+
+            if (
+                inst.mnemonic in ["lw", "lwc1", "ldc1"]
+                and isinstance(inst.args[1], AsmAddressMode)
+                and inst.args[1].rhs.register_name == "sp"
+                and inst.args[1].lhs_as_literal() > 16
+            ):
+                info.subroutine_arg_top = update_min(
+                    info.subroutine_arg_top, inst.args[1].lhs_as_literal()
+                )
```
simonlindholm, line 484: now that allocated_stack_size is computed in the loop above, we could initialize `subroutine_arg_top` to `allocated_stack_size` and make it non-Optional

src/translate.py
```diff
@@ -431,29 +448,89 @@ def get_stack_info(
...
+                        f"Gap in callee-saved word stack region. "
+                        f"Saved: {callee_saved_offset_and_size}, "
+                        f"gap at: {offset} != {top}."
+                    )
+            top += size
+            last_size = size
+        # Expand boundaries to multiples of 8 bytes
+        bottom -= 4 if bottom % 8 == 4 else 0
+        top += 4 if top % 8 == 4 else 0
+        info.callee_save_reg_region = (bottom, top)
```
zbanks, line 529: I know this whole bit is a lot stricter than it needs to be to function -- but the code is making a lot of assumptions/guesses, so I'd rather quickly identify corner cases that break these assumptions.

src/translate.py
```diff
@@ -459,60 +461,71 @@ def update_min(old_or_none: Optional[int], new: int) -> int:
...
+                    raise DecompFailure(
+                        f"Gap in callee-saved word stack region. "
+                        f"Saved: {callee_saved_offset_and_size}, "
+                        f"gap at: {offset} != {top}."
+                    )
+            top += size
+            last_size = size
+        # Expand boundaries to multiples of 8 bytes
+        bottom -= 4 if bottom % 8 == 4 else 0
+        top += 4 if top % 8 == 4 else 0
```
simonlindholm, line 528: This part is surprising to me. Can't it cause sp vars to not be identified?

src/translate.py
```diff
@@ -459,60 +461,71 @@ def update_min(old_or_none: Optional[int], new: int) -> int:
...
+                ):
+                    top += 4
+                    internal_padding_added = True
+                else:
+                    raise DecompFailure(
+                        f"Gap in callee-saved word stack region. "
+                        f"Saved: {callee_saved_offset_and_size}, "
+                        f"gap at: {offset} != {top}."
+                    )
+            top += size
```
simonlindholm, line 524: `top = offset + size`

src/translate.py
```diff
@@ -459,60 +461,71 @@ def update_min(old_or_none: Optional[int], new: int) -> int:
...
+        internal_padding_added = False
+        for offset, size in callee_saved_offset_and_size:
+            if offset != top:
+                if (
+                    not internal_padding_added
+                    and size != last_size
+                    and offset == top + 4
+                ):
+                    top += 4
+                    internal_padding_added = True
```
simonlindholm, line 517: forcing at most one gap seems a bit unnecessarily strict imo

simonlindholm (approved):

src/translate.py
zbanks: As far as I can tell, the compiler always pads these subregions to 8 bytes.
This seems to match up with the different files I was looking at, and [came from this PDF](https://www.cs.unm.edu/~jeffk/cs341f09/_media/o32callingconvention.pdf):
> • All of the different sections on the stack must be double-word (8 byte) aligned. This is just so
that we can store double words onto the stack without them being unaligned.

IDO seems to put the padding low, and GCC seems to put the padding high. 

I could also just remove this pair of lines for simplicity, the 4-byte padded region is unused.

simonlindholm: How odd... that just seems wasteful. +1 to removing the lines, I feel like they just add risk.

PR #99, opened 2021-06-01
Option to run tests in parallel

zbanks:
In general, this PR just touches `run_tests.py`, so I'm not sure how much it needs to be PR'd, but I'm happy to incorporate any feedback. 

----
- Adds `-j N`/`--parallel=N` argument to `run_tests.py` 
- Test name + output is grouped together and printed from the main thread atomically, to keep the log easy to read & avoid tearing.
- Print test summary at the end w/ # of tests
- Basic test on runtime on my 4-core laptop, `./run_tests.py --project ~/mm`
    - Baseline: 183 seconds
    - `-j 2`: 92 seconds
    - `-j 4`: 58 seconds
    - `-j 5`: 58 seconds

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import argparse
+import attr
 import contextlib
 import difflib
 import io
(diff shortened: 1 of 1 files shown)


simonlindholm (approved):
Awesome! 10/10 would review again

PR #101, opened 2021-06-04
Add `--valid-syntax` option to use macros to emit valid C

zbanks:
When the option is enabled, the following things are replaced by macros:

- Unknown types (`?` -> `MIPS2C_UNK`)
- Bitwise casts (`(bitwise f32)` -> `MIPS2C_BITWISE`)
- Unknown struct field accesses (`x->unk4` -> `MIPS2C_FIELD`)

----

Most of this PR is just plumbing the formatter to `Type` so that we can use the formatting options while stringifying types.
I've also written a `mips2c_macros.h` that could be `#include`'d from an output file if the plan is to try to recompile the output. 

In particular, I'm wondering if there's a better way to handle the naming & stuff. Is `--valid-syntax` clear enough? Is `MIPS2C` a good macro prefix? Should we automatically prepend `mips2c_macros.h` to any output when this flag is enabled?


mips2c_macros.h
```diff
@@ -0,0 +1,17 @@
...
+
+// Unknown types
+#define MIPS2C_UNK      s32
+#define MIPS2C_UNK8     s8
+#define MIPS2C_UNK16    s16
+#define MIPS2C_UNK32    s32
+#define MIPS2C_UNK64    s64
+
+// Bitwise (reinterpret) cast
+#define MIPS2C_BITWISE(_type, _expr) (*(_type *)&(_expr))
```
simonlindholm, line 12: why the underscores?

note that this won't compile if expr is an rvalue, so another option, bad in another way, is to define this as the identity function

mips2c_macros.h
```diff
@@ -0,0 +1,17 @@
...
+
+// Unknown field access, like `(_type) _expr->unk_offset`
+#define MIPS2C_FIELD(_expr, _type, _offset) (*(_type *) ((s8 *)(_expr) + (_offset)))
+
+#endif
```
simonlindholm, line 17: `#define ERROR(x) (0)` would be another thing we could add here; it's blatantly wrong but may cause some misidentified float -> unsigned conversions to compile

mips2c_macros.h
```diff
@@ -0,0 +1,17 @@
+#ifndef _MIPS2C_MACROS_H_
+#define _MIPS2C_MACROS_H_
```
simonlindholm, line 2: `MIPS2C_MACROS_H` to avoid the compiler-reserved "underscore, capital letter" name pattern. Not that it matters

Maybe add a comment about what this file is for?

simonlindholm (approved):

mips2c_macros.h
zbanks: Yeah good call, I'll also add `TRAP_IF`, `BREAK`, and `SYNC`. 
I think it's OK if they're always prefixed with `MIPS2C_`? (e.g `MIPS2C_ERROR`?) If not that should be easy to fix.

simonlindholm: Sure, that sounds okay. TRAP_IF and SYNC won't come up in practice, they are restricted to handwritten asm. But BREAK may, for misidentified divisions/modulo.

mips2c_macros.h
zbanks: Yeah good point, I was copying mm & oot's style but that's not really necessary to follow here.

Added a header comment, and also switched all comments to C89-style for IDO.

mips2c_macros.h
zbanks: Underscores is just a habit from a different style, you're right they're not needed here. 

And I originally thought it was rare for expr to not be an rvalue - but it's more common than I thought.  I changed this to just a plain cast.

PR #102, opened 2021-06-06
Add GitHub actions

simonlindholm:


simonlindholm:
Hm, doesn't seem to be triggering from a PR. Let's try merging it into master...

PR #106, opened 2021-06-07
Add Type representation for functions

zbanks:
- Added `K_FN` kind to Type, with individual Type tracking of return type & parameters via a `FunctionSignature` object.
    - I don't love that there's now *three* representations of function signatures (`types.FunctionSignature`, `c_types.Function`, `ca.FuncDecl`) but I don't think it's that bad. `c_types.Function` would be the only candidate to delete, but it's nice to have a frozen representation of `ca.FuncDecl` that is easier to parse. \
    - Tracking all of the return/param types separately, rather than through a single CType, allows for much better type inference. Before this change, almost all function pointers could only be determined to be `void *`. 
- Added `K_VOID` kind to Type, solely to represent void returns. Most `void *` should be represented as `Type.ptr(None)` - "a pointer to anything", not `Type.ptr(Type._void())` - "a pointer to nothing".
    - This maybe could be represented with a `CType` instead? 
- I think future work could add kinds for structs, arrays, & unions -- but that's not the immediate direction I'm heading. I wanted this change to be able to infer the types of globals, and this was on the critical path. It seems we're set up to slowly migrate types out of `K_CTYPE` and into individual kinds.
- Stringifying types now always goes through the `pycparser` library to handle all the weird edge cases.
- Change `MIPS2C_FIELD`'s second parameter to take the field's *pointer* type. This is required because the previous `(type *)` doesn't work if `type` is a function pointer, the `*` is in the wrong place.
    - I think it's possible to correctly generate the type using the `__typeof__` GCC extension, but I don't think that's worth it.


tests/end_to_end/function-pointer/irix-o2-out.c
```diff
@@ -1,3 +1,3 @@
-s32 test(void *arg0, ? arg1) {
+s32 test(s32 (*arg0)(?, ? *), ? arg1) {
```
zbanks, line 1: This is kind of a cool example (even though mips2c incorrectly guessed that the function takes two args). The type here is recursive, the second arg has the type of the function pointer itself!

tests/end_to_end/misc1/irix-o2-validsyntax-out.c
```diff
@@ -1,7 +1,8 @@
 extern s32 D_410170;
 extern MIPS2C_UNK D_410178;
-extern MIPS2C_UNK func_00400140;
-extern MIPS2C_UNK func_00400158;
+extern s32 (*func_00400140)(MIPS2C_UNK, MIPS2C_UNK, s32, MIPS2C_UNK, s32);
+extern MIPS2C_UNK (*func_00400158)(MIPS2C_UNK32, s32, s32);
```
zbanks, line 4: This doesn't seem right - it should be able to tell that these are bare fns and not fn pointers. I'm going to investigate this a bit further.

src/translate.py
```diff
@@ -3695,28 +3712,27 @@ def make_arg(offset: int, type: Type) -> PassedInArg:
...
-                    initial_regs[slot.reg] = make_arg(slot.offset, slot.type)
-            for reg in possible_regs:
-                offset = 4 * int(reg.register_name[1])
-                initial_regs[reg] = make_arg(offset, Type.any_reg())
-            known_params = True
-
-    if not known_params:
+        fn_type, _ = ptr_type_from_ctype(typemap.functions[function.name].type, typemap)
+    else:
+        fn_type = Type.function_ptr()
```
simonlindholm, line 3754: It would feel more natural to more to not pointer-wrap this

src/types.py
```diff
@@ -117,64 +132,153 @@ def is_int(self) -> bool:
...
+                    name=param.name,
+                    type=param.type._to_ctype(seen.copy(), fmt),
+                    quals=[],
+                    storage=[],
+                    funcspec=[],
+                    init=None,
+                    bitsize=None,
+                )
+                set_decl_name(decl)
+                params.append(decl)
```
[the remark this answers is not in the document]
zbanks: I'm going to stick this on the `K_CTYPE `branch, so that `_to_ctype(...)` always returns a clean copy of `CType` that can be modified.

The function parameters do have names by default, the names are manually stripped out by `type_from_global_decl` when parsing the `TypeMap.var_types` dict. The types are *not* stripped out from the `TypeMap.functions` dict.

src/types.py
```diff
@@ -117,64 +132,153 @@ def is_int(self) -> bool:
...
+
+        return simple_ctype(f"{sign}{size}")
 
     def format(self, fmt: Formatter) -> str:
-        return self._stringify(set(), fmt)
+        return self.to_decl("", fmt)
 
     def __str__(self) -> str:
-        return self._stringify(set(), Formatter(debug=True))
+        return self.format(Formatter())
```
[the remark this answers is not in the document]
zbanks: You're right, but this was an oversight on my part, I'm going to put it back in here.

zbanks:
I think I addressed all of your points:

- Refactored GlobalSymbol & `translate_to_ast` to use a bare function type instead of function pointer
   - I think this addresses a lot of your comments about wrapping functions with `Type.ptr()`
- Make `Type._to_ctype()` return a copied `CType` that can be modified
- `Type.is_function_variadic_with` -> `FunctionSignature.unify_with_args`

And I made two changes unrelated to your comments:
- Tweak `has_return` detection to use TypeMap info
   - This is slightly unrelated to the rest of the PR, but as I looked at the tests I found some cases where we were throwing out the function type declaration from the TypeMap because there wasn't a custom return. 
- Add `--emit-globals` to `--project`/`--project-with-context` tests

src/translate.py
```diff
@@ -3082,7 +3097,7 @@ def compute_has_custom_return(nodes: List[Node]) -> None:
         for n in nodes:
             block_info = n.block.block_info
             assert isinstance(block_info, BlockInfo)
-            if block_info.has_custom_return or block_info.has_function_call:
+            if block_info.has_custom_return or block_info.has_function_call_return:
```
simonlindholm, line 3099: This change is wrong: function calls act as boundaries for the "has custom return" computation because they clobber registers, and they do so even if the function returns `void` (sorry for not commenting this code properly)

src/translate.py
```diff
@@ -3726,28 +3762,29 @@ def make_arg(offset: int, type: Type) -> PassedInArg:
...
-                initial_regs[reg] = make_arg(offset, Type.any_reg())
-            known_params = True
-
-    if not known_params:
+        fn_type = type_from_ctype(typemap.functions[function.name].type, typemap)
+        fn_decl_provided = True
+    else:
+        fn_type = Type.function()
+        fn_decl_provided = False
+    fn_type.unify(function_sym.type)
```
simonlindholm, line 3771: nit: I'd inline the definition of function_sym here to reduce the number of live variables, since it's only used once

src/translate.py
```diff
@@ -3726,28 +3762,29 @@ def make_arg(offset: int, type: Type) -> PassedInArg:
...
+
+    fn_sig = Type.ptr(fn_type).get_function_pointer_signature()
+    assert fn_sig is not None, "fn_type is known to be a function pointer"
```
simonlindholm, line 3774:
```suggestion
    assert fn_sig is not None, "fn_type is known to be a function"
```

src/types.py
```diff
@@ -117,64 +132,141 @@ def is_int(self) -> bool:
...
+            bitsize=None,
+        )
+        set_decl_name(decl)
+        return to_c(decl)
 
-    def to_decl(self, var: str, fmt: Formatter) -> str:
-        ret = self.format(fmt)
-        prefix = ret if ret.endswith("*") else ret + " "
-        return prefix + var
+    def _to_ctype(self, seen: Set["Type"], fmt: Formatter) -> CType:
```
simonlindholm, line 201: btw, I wonder if we should use this pattern of creating a pycparser type for printing in more places. Could help with precedence and avoiding excessive parentheses.

src/types.py
```diff
@@ -273,13 +374,88 @@ def of_size(size: int) -> "Type":
...
+                can_unify &= x.type.unify(y.type)
+        can_unify &= self.return_type.unify(other.return_type)
+
+        if can_unify:
+            if not self.params_known:
+                self.params = other.params
+                self.params_known = other.params_known
+            elif not other.params_known:
+                other.params = self.params
+                other.params_known = self.params_known
```
simonlindholm, line 414: If neither params_known is true, is it worth trying to unify up to the min of the two param lengths, and setting both `params` equal to the longer one?

src/translate.py
```diff
@@ -1673,14 +1691,14 @@ def deref(
             type = new_type
 
     # Dereferencing pointers of known types
-    target = get_pointer_target(var.type, typemap)
-    if field_name is None and target is not None:
-        sub_size, sub_type = target
-        if sub_size == size and offset % size == 0:
+    target = var.type.get_pointer_target()
+    if field_name is None and target is not None and (target.size or target.is_ctype()):
```
simonlindholm, line 1695: why is this check needed?

simonlindholm (approved):

src/translate.py
```diff
@@ -1173,6 +1184,10 @@ def dependencies(self) -> List[Expression]:
         return [self.expr]
 
     def format(self, fmt: Formatter) -> str:
+        if self.expr.type.is_function():
+            # Functions are automatically converted to function pointers
+            # without an explicit `&` by the compiler
```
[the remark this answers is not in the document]
zbanks: Yeah - we handle it in `address_of_gsym`. (Although I think I want to try refactoring that and making `address_of_gsym` *always* return an `AddressOf`, and then doing the type detection here at format time? I think that would solve the `StringLiteral` problem you mentioned, and make it easier to distinguish arrays in `.rodata`/`.data`.)

src/types.py
zbanks: Yeah it's a good point. I feel like it might make the code look a bit more verbose, but it'd have much fewer branches & be more robust.
I think types are particularly hard to print right, so this seemed like a good place to start.

src/translate.py
```diff
@@ -936,8 +933,18 @@ def dependencies(self) -> List[Expression]:
         return self.args + [self.function]
 
     def format(self, fmt: Formatter) -> str:
+        if isinstance(self.function, Cast):
```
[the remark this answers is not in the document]
zbanks: (This was moved into `Cast.format` where I added comments)

src/translate.py
zbanks: I'll revert it.
Is splitting the `has_return` into `could_have_return` + `has_custom_return` in `translate_ast()` OK? I think that's separate, right?

simonlindholm: Yeah, that part looked fine to me

src/translate.py
zbanks: You're right, I can remove the `target.is_ctype()` check. 
(The `target.size` is needed because it can be 0, and we try computing `offset % target.size`)

simonlindholm: I'm only seeing `offset % size`, not `offset % target.size`, and `size` should be statically known to be non-zero at this point. (I also feel like `.size` is a private property of Type, and its semantics are kinda weird, so seems iffy to base logic on)

zbanks: Oh ha, I forgot why I put this in (I should at the very least put in a `TODO` comment!) 

The `target.is_ctype()` is there to port over the old logic that was in `get_pointer_target()`: https://github.com/matt-kempster/mips_to_c/pull/106/files/eace377752e41206c3c85bd26c896f28606278b1#diff-5e877229fb015bfa4a784ad6bca335f1baccf27a146db474ebc04266b358ea19L399

If I remove it, honestly the output gets a bit better... but it's mostly unrelated to what I was trying to change here, and I didn't want to confuse myself. 

Removing the check results in a few diffs like this, in the `break` end-to-end test:
```diff
-        if (D_4101D0.unk4 == 2) {
-            D_4101D0.unk8 = 3;
-        } else if (D_4101D0.unk8 == 2) {
-            D_4101D0.unkC = 3;
+        if ((&D_4101D0)[1] == 2) {
+            (&D_4101D0)[2] = 3;
+        } else if ((&D_4101D0)[2] == 2) {
+            (&D_4101D0)[3] = 3;
```

I'm going to leave it in for now, with a TODO?

zbanks: Oh yeah, you're right, I'll fiddle with this a bit more.

simonlindholm: Sure!

src/translate.py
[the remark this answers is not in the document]
simonlindholm: The `StringLiteral` logic I agree should be moved to formatting (though it may still be nice to annotating the type as char* early). Array->pointer decay I'm less sure about, it's not just a formatting concern since it's really changing the type. But if the decay can be moved into other parts of the type inference code (e.g. decaying whenever we do a dereference or arithmetic) it could work.

PR #107, opened 2021-06-07
Add GlobalInfo class for managing top-level declarations

zbanks:
- Created a `GlobalInfo` class for managing any (potentially mutable) cross-function state.
    - Right now this is the TypeMap, Rodata, and types of global symbols
    - Happy to have a better name, it felt like a good complement to `FunctionInfo`?
- Added `--emit-globals` flag, making the default behavior to *not* print these detected globals.
    - I think it'd be nice to make this on by default once there's a chance for it to "burn in" for a bit, and #106 is merged which will fix global function decls which currently show up as `void *` or `? *`.
    - Added `--emit-globals` to a smattering of existing tests
- Without modifying the test flags, there is no change in test output.

### DIFF
--- a/tests/end_to_end/function-pointer2/irix-o2-out.c
+++ b/tests/end_to_end/function-pointer2/irix-o2-out.c
@@ -1,3 +1,6 @@
+extern int (float) bar;
+extern int (*)(float) glob2;
+
 void test(void) {
     glob = foo;
     glob = &bar;
--- a/tests/end_to_end/misc1/irix-o2-validsyntax-out.c
+++ b/tests/end_to_end/misc1/irix-o2-validsyntax-out.c
@@ -1,3 +1,8 @@
+extern s32 D_410170;
+extern MIPS2C_UNK D_410178;
+extern MIPS2C_UNK func_00400140;
+extern MIPS2C_UNK func_00400158;
+
 s32 test(s32 arg0, MIPS2C_UNK arg1) {
     s32 sp2C;
     s32 sp28;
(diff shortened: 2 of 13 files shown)


tests/end_to_end/function-pointer2/irix-o2-out.c
```diff
@@ -1,3 +1,6 @@
+extern int (float) bar;
+extern int (*)(float) glob2;
```
zbanks, line 2: `glob2`'s type is right here; the formatting of `bar` is wrong but that gets fixed by #106

tests/end_to_end/misc1/irix-o2-validsyntax-out.c
```diff
@@ -1,3 +1,8 @@
+extern s32 D_410170;
+extern MIPS2C_UNK D_410178;
+extern MIPS2C_UNK func_00400140;
+extern MIPS2C_UNK func_00400158;
```
simonlindholm, line 4: I expect these will also change to real function declarations with the other PR?

tests/end_to_end/function-pointer2/irix-o2-out.c
simonlindholm: and `glob2`'s formatting too I imagine

simonlindholm (approved):

tests/end_to_end/misc1/irix-o2-validsyntax-out.c
zbanks: Yep! Exactly. 
(I wanted to keep the PRs separate because I think that makes it easier to review)

simonlindholm: Thanks for doing that, it helped!

PR #108, opened 2021-06-09
Split out TypeData from Type

simonlindholm:
Having the only accessor for TypeData do get_representative prevents bugs.


zbanks (approved):
This is a nice change (even if there are some mis-uses it can't prevent)

PR #109, opened 2021-06-09
Improve function type unification

zbanks:
This PR mostly resolves that TODO I had left open from my previous PR.
```
# TODO: If neither params_known is true, can we try to unify up
# to the min of the two param lengths, and set both params equal
# to the longer one?
```
- Now `FunctionSignature.unify()` does that ^
- Added an explicit note about how a function type without `params_known` can have its *number* of parameters change over time: this can result in early calls having fewer arguments than the eventually determined type. I added a warning about this to `FuncCall.format()`
   - An interesting idea is to try to patch these calls up at a later time: either remove args, or add args from the `possible_regs` candidates that were previously discarded. I think this is tricky to do right; doing it at format-time would make `FuncCall.dependencies()` unreliable.
   - On the plus side, this seems pretty rare and the errors aren't too bad for a human to patch. (It is also fixable by re-running mips2c with the derived function prototype in the context)
- Use casts to make sure that function arguments will have the right types, so that `FunctionSignature.unify_with_args()` is more likely to succeed. 
- Overall, the changes to the MM/OOT tests are small & explainable.

### DIFF
--- a/src/types.py
+++ b/src/types.py
@@ -406,35 +406,46 @@ class FunctionSignature:
     is_variadic: bool = attr.ib(default=False)
 
     def unify(self, other: "FunctionSignature") -> bool:
-        if self.is_variadic != other.is_variadic:
-            return False
-
-        # Try to unify *all* ret/param types, without returning early
-        can_unify = True
         if self.params_known and other.params_known:
+            if self.is_variadic != other.is_variadic:
+                return False
             if len(self.params) != len(other.params):
                 return False
-            for x, y in zip(self.params, other.params):
-                can_unify &= x.type.unify(y.type)
-        can_unify &= self.return_type.unify(other.return_type)
 
-        # TODO: If neither params_known is true, can we try to unify up
-        # to the min of the two param lengths, and set both params equal
-        # to the longer one?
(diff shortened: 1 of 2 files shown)


simonlindholm:
> I think this is tricky to do right; doing it at format-time would make FuncCall.dependencies() unreliable.

Yeah. It may work to fixed-point-iterate translate_to_ast. (Which would help with inter-procedural type analysis as well.) I agree with ignoring it at this point.

src/types.py
```diff
@@ -406,35 +406,46 @@ class FunctionSignature:
...
-        # TODO: If neither params_known is true, can we try to unify up
-        # to the min of the two param lengths, and set both params equal
-        # to the longer one?
+        # Try to unify *all* ret/param types, without returning early
+        # TODO: If not all the types unify, roll back any changes made
+        can_unify = self.return_type.unify(other.return_type)
+        for x, y in zip(self.params, other.params):
+            can_unify &= x.type.unify(y.type)
+
         if can_unify:
```
simonlindholm, line 421: nit: I'd make this an early return

simonlindholm (approved):

PR #110, opened 2021-06-09
Support loading in .data & .bss in addition to .rodata

zbanks:
This PR has a lot of lines changed, but it's mostly renaming. The technical aspects are pretty small, I wanted to check that the renaming I did makes sense.

- `Rodata` --> `DataSection`, which now holds `.rodata`, `.data`, and `.bss`. (although `.bss` is untested)
- `RodataEntry` --> `DataSectionEntry`, which has a `is_readonly` attr for distinguishing rodata from data/bss.
- Changed the command line arg `--rodata` to `--data` & updated help/error messages. Left in support for `--rodata` as a hidden option to preserve compatibility.

### DIFF
--- a/src/flow_graph.py
+++ b/src/flow_graph.py
@@ -5,7 +5,7 @@ from typing import Any, Counter, Dict, Iterator, List, Optional, Set, Tuple, Uni
 import attr
 
 from .error import DecompFailure
-from .parse_file import Function, Label, Rodata
+from .parse_file import Function, Label, DataSection
 from .parse_instruction import (
     AsmAddressMode,
     AsmGlobalSymbol,
@@ -240,11 +240,13 @@ def normalize_likely_branches(function: Function) -> Function:
     return new_function
 
 
-def prune_unreferenced_labels(function: Function, rodata: Rodata) -> Function:
+def prune_unreferenced_labels(
+    function: Function, data_section: DataSection
+) -> Function:
     labels_used: Set[str] = {
         label.name
         for label in function.body
-        if isinstance(label, Label) and label.name in rodata.mentioned_labels
+        if isinstance(label, Label) and label.name in data_section.mentioned_labels
     }
     for item in function.body:
         if isinstance(item, Instruction) and item.is_branch_instruction():
@@ -628,12 +630,12 @@ def simplify_standard_patterns(function: Function) -> Function:
     return new_function
 
 
-def build_blocks(function: Function, rodata: Rodata) -> List[Block]:
+def build_blocks(function: Function, data_section: DataSection) -> List[Block]:
     verify_no_trailing_delay_slot(function)
     function = normalize_likely_branches(function)
-    function = prune_unreferenced_labels(function, rodata)
+    function = prune_unreferenced_labels(function, data_section)
     function = simplify_standard_patterns(function)
-    function = prune_unreferenced_labels(function, rodata)
+    function = prune_unreferenced_labels(function, data_section)
 
     block_builder = BlockBuilder()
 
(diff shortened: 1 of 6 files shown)


src/flow_graph.py
```diff
@@ -917,20 +921,27 @@ def find_block_by_label(label: str) -> Optional[Block]:
...
+                    "file pointed to by --data.\n"
                     'It needs to be within ".section .rodata" or ".section .late_rodata".\n'
                     "\n"
                     "(You might need to pass --goto and --no-andor flags as well, "
                     "to get correct control flow for non-jtbl switch jumps.)"
                 )
 
-            jtbl_entries = rodata.values[jtbl_name].data
+            jtbl_data_entry = data_section.values[jtbl_name]
+            if not jtbl_data_entry.is_readonly:
```
simonlindholm, line 938: this seems like an unnecessary check to me, I could see decompiling code where sections are improperly annotated

src/flow_graph.py
```diff
@@ -240,11 +240,13 @@ def normalize_likely_branches(function: Function) -> Function:
     return new_function
 
 
-def prune_unreferenced_labels(function: Function, rodata: Rodata) -> Function:
+def prune_unreferenced_labels(
+    function: Function, data_section: DataSection
```
simonlindholm, line 244: Hm, this naming does feel very misleading to me. It makes me ask "what is it using the .data section for? which .data section?" when the answer is "it passes in all data-y sections from all files merged together"

src/main.py
```diff
@@ -193,12 +193,21 @@ def parse_flags(flags: List[str]) -> Options:
...
         metavar="ASM_FILE",
-        dest="rodata_files",
+        dest="data_files",
         action="append",
         default=[],
-        help="read jump table data from this file",
+        help=argparse.SUPPRESS,
+    )
+    parser.add_argument(
+        "--data",
```
simonlindholm, line 205: an alternative to this btw would be supporting multiple .s files as input, where it would decompile all of them, and data/rodata files would just be .s files with no asm in them. we should perhaps discuss how we envision inter-function type inference working

src/parse_file.py
```diff
@@ -281,40 +295,45 @@ def process_label(label: str, *, glabel: bool) -> None:
             elif ifdef_level == 0:
                 if line.startswith(".section"):
                     curr_section = line.split(" ")[1].split(",")[0]
-                    if curr_section == ".late_rodata":
+                    if curr_section in (".rdata", ".late_rodata"):
                         curr_section = ".rodata"
                 elif (
                     line.startswith(".rdata")
                     or line.startswith(".rodata")
+                    or line.startswith(".rdata")
```
simonlindholm, line 303: already checked two lines up

(`.section .rdata` is also invalid I believe, but it's fine to support)

src/translate.py
```diff
@@ -2144,6 +2144,7 @@ def handle_load(args: InstrArgs, type: Type) -> Expression:
             if (
                 ent
                 and ent.data
+                and ent.is_readonly
```
simonlindholm, line 2147: already checked in rodata_value

src/main.py
zbanks: Yeah this is a great point. 

I could revert the changes to the CLI (except for help messages) if you think it's better to avoid changing the interface when we might change it again soon? Or do you think that should get resolved sooner?

src/flow_graph.py
zbanks: Yeah, it is misleading. Do you think I could solve this with a comment/docstring?

src/flow_graph.py
zbanks: I was kind of worried about emitting incorrect output after assuming that a jumptable was read-only -- but you're right, I think that's way less common than improperly-annotated code.

zbanks:
After discussion on discord, the CLI changes are reverted (in favor of updating them soon to support multiple input/output files) and the `Rodata` class name was changed to `AsmData`.

PR #111, opened 2021-06-10
Improve global_decls() for C-like output

zbanks:
- When a function fails to decompile, wrap the error message/traceback in `/* ... */` so that you can still try to compile the output.
    - I considered only doing this when `--valid-syntax` is enabled, but it seems benign?
- Added new `MIPS2C` macros for unaligned reads. Rearranged existing macros to be in order of "most-to-least common" to be easier to read.
- ~~Use `/* ... */` for comments, because IDO/c89 does not support `//`~~
- `GlobalInfo` now handles `.global_symbol(...)` and `.address_of_gsym()`
- Expose more information in the `global_decls()` section.
    - If a function is marked `/* extern */`, this means it should be defined in a common header file.
    - Indicate when variables in file scope look like arrays.
    - For simple types, generate the initializer. 

---

I tried directly compiling the C output from this branch (with `--valid-syntax`).
I compiled with `mips-linux-gnu-gcc` because it supports `-include`, which made it easier to script this. (I spot checked a few files with IDO too.)

- OOT, per-function asm, with context: **80%** success
- MM, per-object asm, with context: **25%** success
- Paper Mario, per-function asm, no context: **20%** success

The 10 most common remaining errors, across all 3 projects:
```
  count  error message
   8491  case label not within a switch statement
   1152  conflicting types for ‘get_variable’
    782  assignment to expression with array type
    778  ‘default’ label not within a switch statement
    763  pointer value used where a floating point value was expected
    732  conflicting types for ‘get_npc_unsafe’
    718  too few arguments to function ‘Audio_PlayActorSound2’
    599  cannot convert to a pointer type
    487  expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ at end of input
    439  conflicting types for ‘gPlayerStatus’
```

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -90,8 +90,8 @@ def run(options: Options) -> int:
                 print(f"Function {name} not found.", file=sys.stderr)
                 return 1
 
-    return_code = 0
-    global_info = GlobalInfo(mips_file.asm_data, typemap)
+    function_names = {fn.name for fn in mips_file.functions}
+    global_info = GlobalInfo(mips_file.asm_data, function_names, typemap)
     function_infos: List[Union[FunctionInfo, Exception]] = []
     for function in functions:
         try:
@@ -106,7 +106,7 @@ def run(options: Options) -> int:
             function_infos.append(e)
 
     if options.visualize_flowgraph:
-        return return_code
+        return 0
 
     fmt = options.formatter()
     if options.emit_globals:
@@ -114,6 +114,7 @@ def run(options: Options) -> int:
         if global_decls:
             print(global_decls)
 
+    return_code = 0
     for index, (function, function_info) in enumerate(zip(functions, function_infos)):
         if index != 0:
             print()
@@ -128,11 +129,16 @@ def run(options: Options) -> int:
             function_text = get_function_text(function_info, options)
             print(function_text)
         except DecompFailure as e:
-            print(f"Failed to decompile function {function.name}:\n\n{e}")
+            print("/*")
+            print(f"Failed to decompile function {function.name}:\n")
+            print(e)
+            print("*/")
             return_code = 1
         except Exception:
+            print("/*")
             print(f"Internal error while decompiling function {function.name}:\n")
             print_exception(sanitize=options.sanitize_tracebacks)
+            print("*/")
             return_code = 1
 
     return return_code
--- a/src/translate.py
+++ b/src/translate.py
@@ -2112,7 +2112,8 @@ def add_imm(source: Expression, imm: Expression, stack_info: StackInfo) -> Expre
 def handle_load(args: InstrArgs, type: Type) -> Expression:
     # For now, make the cast silent so that output doesn't become cluttered.
     # Though really, it would be great to expose the load types somehow...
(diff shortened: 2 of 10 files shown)


src/main.py
```diff
@@ -90,8 +90,8 @@ def run(options: Options) -> int:
                 print(f"Function {name} not found.", file=sys.stderr)
                 return 1
 
-    return_code = 0
-    global_info = GlobalInfo(mips_file.asm_data, typemap)
+    function_names = {fn.name for fn in mips_file.functions}
+    global_info = GlobalInfo(mips_file.asm_data, function_names, typemap)
```
zbanks, line 94: For this, I'm not totally sure which is better:
I could give the whole `mips_file` to `GlobalInfo` -- but that seems overkill? On the other hand, the python API is getting a bit complex for 3rd parties to set up and use.

Trevor89:
>>- Use /* ... */ for comments, because IDO/c89 does not support //
FYI IDO does support // comments, use `-Xcpluscomm` parameter

Trev

zbanks:
> > > * Use /* ... */ for comments, because IDO/c89 does not support //
> 
> FYI IDO does support // comments, use `-Xcpluscomm` parameter
> 
> Trev

Thanks! Switched back to `//` then.

src/translate.py
```diff
@@ -3717,20 +3701,147 @@ def symbol_type(self, sym_name: str) -> Type:
...
+                        comments.append("const")
+                    data_size = entry.size_bytes()
+                    type_size = type.get_size_bits() // 8
+                    if data_size != type_size:
+                        # TODO: Use heuristic to detect padding
+                        if data_size // type_size > 1:
+                            comments.append(f"array: [{data_size // type_size}]")
+                        if data_size % type_size != 0:
+                            comments.append(f"extra bytes: {data_size % type_size}")
+                    elif entry.is_readonly and type.is_float():
```
simonlindholm, line 3831: maybe move this above the other check

src/main.py
simonlindholm: Both options feel good to me, I don't think it matters much either way.

simonlindholm (approved):
A test for this would be nice (with vars in all three of .data, .rodata, .bss)

src/translate.py
zbanks: Oh, this was intentional, to not skip float arrays?
(It'd be better to have a way of marking which constants need to be emitted, but TODO?)

src/translate.py
```diff
@@ -3717,20 +3701,147 @@ def symbol_type(self, sym_name: str) -> Type:
...
+                    comments.append(qualifier)
+                    qualifier = ""
+
+                # Try to convert the data from .data/.rodata into an initializer
+                entry = self.asm_data.values.get(name)
+                if entry is not None:
+                    if entry.is_readonly:
+                        comments.append("const")
+                    data_size = entry.size_bytes()
+                    type_size = type.get_size_bits() // 8
```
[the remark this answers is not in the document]
zbanks: `type.get_size_bits()` can't return 0 or None - in those cases it returns 32.

PR #113, opened 2021-06-13
Show code, if available, in `--visualize` flow graph

zbanks:
Small change!

Not sure if anyone had attachments to the previous flow graph format. If the blocks have not been translated, then it produces very similar output to the old format (except the nodes are rectangles with `// Node N`)

The code in the blocks is not perfect -- it's based solely on the steps from `translate_to_ast()` (before any `if_statements.py`). It still helps to get an approximation of what each branch represents though.

Here's an example from OOT's `ovl_En_Dekubaba/func_809E80D8.s` of what it looks like now - it's a bit easier to follow what's going on.

graphviz_render gv

### DIFF
--- a/src/flow_graph.py
+++ b/src/flow_graph.py
@@ -5,6 +5,7 @@ from typing import Any, Counter, Dict, Iterator, List, Optional, Set, Tuple, Uni
 import attr
 
 from .error import DecompFailure
+from .options import Formatter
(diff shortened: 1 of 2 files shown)


simonlindholm (approved):

matt-kempster:
OMG! Awesome!

PR #114, opened 2021-06-16
Add single exit node (TerminalNode) to FlowGraph

zbanks:
This PR is (mostly? entirely?) orthogonal to my other PR, so I don't think there's any nontrivial merge conflicts between the two and it doesn't matter which order they get merged in.

As discussed on Discord, any ReturnNode could be the exit of the FlowGraph, which complicated some of the interval/structural analysis. Adding a single common exit node helps reduce edge cases.

While making this change, I also made several other refactors. I tried to only change things when I could delete chunks of code (e.g. `immediate_postdominator()` in `if_statements.py`) or if the change helped preserve previous behavior/test output (e.g. the indenting changes).

I don't love how big this PR got -- but I think even though I technically added lines, the code is a bit more straightforward? I also don't think my changes are incompatible with the direction that #76 is headed.

Other refactors:

- Postdominators & loops can now be calculated on the FlowGraph at construction. Previously it was "postdominator within this interval" because there was not the common exit. This moved a lot of code out of `if_statements.py` and into `flow_graph.py`.

- Added `Node.children()` - it's very easy for a node to say what it's children are, rather than manually doing the `.add_parent()` bookkeeping. (The `.add_parent()` stuff could be removed in a future PR, and `parents` could be calculated in `compute_relations`, but the ordering change would introduce output changes, so I kept that out of this PR for clarity)

- `Node.emit_goto` is now only used for very special cases (certain asm instructions, or the user manually annotates the source). Rather than pre-compute which nodes need goto, we just use a goto if the node was already emitted while building the Body.

- The `emit_node()` function now contains the logic for checking if a node has been already been emitted and a goto should be used instead.

- Added `SwitchStatement` to `if_statements.py`.

- Indenting is now done at formatting time, instead of being determined while constructing `Body`s. This makes it possible to shift `Body` segments around after construction without recalculating the indents.

- Similiarly, empty return elision is done after building the `Body`s, instead of during the `build_*` functions. Like indenting, this is to accomidate re-writes.

Notes on test diffs:
- I added a "terminal andor" test case inspired by OOT's `func_809E80D8`. The `
429dbfb` commit has the output of the test on master, where you can see it failed to combine the conditions.
- I'd consider the output of the `switch` test to be improved: it largely just moves the `case` blocks out of the ends of `if` statements.
- The `multi-switch` test is a doozy. afaict it's semantically equivalent output, but it's still not great looking. (There is slightly less nesting though?).
- `loop-selfassign-phi` has an infinite loop, so the flow graph is not reducible, and now mips2c automatically will fall back to the "no ifs"/"goto" mode. (There are only 12 functions across OOT/MM/Papermario that are non-reducible, and they also appear to be infinite loops.)
- There were 10 tests that changed in total (counting g/o2 pairs separate) -- I think the rest are relatively inconsequential.


ethteck:
Not sure if this is appropriate, but I tested this out on a kind of nasty function in paper mario (func_802B62A4_E25174)

the results seem to be an improvement, but there are some new sections that seem not great:

```            case 16:
            case 19:
                if ((s32) temp_s0_3 < 2) {
                    goto block_30;
                }
                goto block_31;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
                goto block_32;
```

src/if_statements.py
```diff
@@ -118,45 +143,80 @@ def format(self, fmt: Formatter) -> str:
...
+            # which reduces indentation to make the output more readable
+            self.statements.append(attr.evolve(if_else, else_body=None))
+            if if_else.else_body is not None:
+                self.extend(if_else.else_body)
+            return
+
         self.statements.append(if_else)
 
+    def add_switch(self, switch: SwitchStatement) -> None:
+        self.add_statement(switch)
```
simonlindholm, line 202: (in preparation for future changes?)

src/if_statements.py
```diff
@@ -182,167 +261,72 @@ def label_for_node(context: Context, node: Node) -> str:
...
+            emit_goto(context, node, body)
+            return False
     else:
-        body.add_statement(LabelStatement(max(indent - 1, 0), context, node))
-        body.add_node(node, indent, comment_empty=True)
+        body.add_statement(LabelStatement(context, node))
+        context.emitted_nodes.add(node)
+
+    body.add_node(node, comment_empty=True)
+    if isinstance(node, ReturnNode):  # and not node.is_real():
```
simonlindholm, line 287: ?

src/if_statements.py
```diff
@@ -491,7 +476,7 @@ def get_andor_if_statement(
             # An edge-case of our pattern-matching technology: without
             # this, self-loops match the pattern indefinitely, since a
             # self-loop node's conditional edge points to itself.
-            or next_node.is_loop()
+            or next_node.loop
```
simonlindholm, line 479: Is this equivalent? I thought `.loop` meant loop head...

src/if_statements.py
```diff
@@ -118,45 +143,80 @@ def format(self, fmt: Formatter) -> str:
...
+        Returns True if the body ends in an unconditional jump, which
+        may allow for some syntax transformations.
+        This function is conservative: it only returns True if we're
+        *sure* if the control flow won't continue past the Body boundary.
+        """
+        for statement in self.statements[::-1]:
+            if not statement.should_write():
+                continue
+            return isinstance(statement, SimpleStatement) and statement.is_jump
+        return False
```
simonlindholm, line 218: (It wasn't immediately clear to me that this included ReturnNode's by virtue of them always ending with a `return;` with is_jump = true; maybe worth mentioning in the doc comment)

src/if_statements.py
```diff
@@ -568,82 +553,72 @@ def join_conditions(conditions: List[Condition], op: str) -> Condition:
...
-                        if_body=if_body,
-                        else_body=None,
-                    )
-                )
-            elif isinstance(curr_start, SwitchNode):
-                emit_switch_jump(context, curr_start, body, indent)
-            else:  # ReturnNode
-                add_return_statement(context, body, curr_start, indent, last=False)
+            # through to the next node by index, after writing a goto.
+            print(f">>> GOTO {curr_start.name()}")
```
simonlindholm, line 620: heh

src/if_statements.py
```diff
@@ -654,36 +629,29 @@ def build_flowgraph_between(
...
+            body.add_if_else(build_conditional_subgraph(context, curr_start, curr_end))
+        elif isinstance(curr_start, SwitchNode):
+            body.add_switch(build_switch_between(context, curr_start, curr_end))
+        else:
+            # No branch, but check that we didn't skip any nodes
+            if curr_start.children() != {curr_end}:
+                raise DecompFailure(
+                    f"While emitting flowgraph between {start.name()}:{end.name()}, "
+                    f"skipped nodes while stepping from {curr_start.name()} to {curr_end.name()}."
+                )
```
simonlindholm, line 647: Should probably be a warning, if it can occur legitimately. Some output is better than none.

simonlindholm (commented):
Random comments. Can't say I have a good understanding of the big picture, still...

src/flow_graph.py
```diff
@@ -837,12 +855,46 @@ def __str__(self) -> str:
...
+
+
+@attr.s(eq=False)
+class NaturalLoop:
+    """
+    Loops are defined by "backedges" `tail->head` where `head`
+    dominates `tail` in the control flow graph.
+
+    The loop body is the set of nodes dominated by `head` where there
+    is a path from the node to `tail` that does not contain `head`.
```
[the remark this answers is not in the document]
zbanks: Yes, for conditional breaks/early returns, the BB with the `if` is part of the loop, but the branch with the break/`ReturnNode` is not part of the loop. This definition of a "natural loop" seems to be common in interval/structural analysis. 

Right now this is mostly unused - but you're right you have to be careful when trying to emit the loop body to ensure that all of the exit nodes are emitted. The benefit of this definition is that it makes it easier to identify the loop condition, `break`s, and `continue`s.

src/flow_graph.py
```diff
@@ -764,27 +770,35 @@ class BaseNode:
     dominators: Set["Node"] = attr.ib(init=False, factory=set)
     immediate_dominator: Optional["Node"] = attr.ib(init=False, default=None)
     immediately_dominates: List["Node"] = attr.ib(init=False, factory=list)
+    postdominators: Set["Node"] = attr.ib(init=False, factory=set)
+    immediate_postdominator: Optional["Node"] = attr.ib(init=False, default=None)
+    immediately_postdominates: List["Node"] = attr.ib(init=False, factory=list)
+    loop: Optional["NaturalLoop"] = attr.ib(init=False, default=None)
```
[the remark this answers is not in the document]
zbanks: Correct, this is set on the loop head (so `(node.loop is None) or (node.loop.head is node)` is the invariant).

I'm using the definition of "natural loop" where the header node may have multiple backedges, and that is considered a single loop. You're right it's ambiguous for nested do-whiles, I don't think there's anything we could do to distinguish these two cases:

```c
do {
    do {
        x();
    } while (a);
    y();
} while (b);

// -- indistinguishable from --

do {
    x();
    if (!a) continue;
    y();
} while (b);
```

src/flow_graph.py
```diff
@@ -1056,95 +1137,125 @@ def duplicate_premature_returns(nodes: List[Node]) -> List[Node]:
...
-    while True:
-        # Traverse all edges other than indirect jumps, and collect the targets.
-        reachable: Set[Node] = {nodes[0]}
-        predecessors: Dict[Node, Node] = {}
-        for i, node in enumerate(nodes):
-            fallthrough: Optional[Node]
-            if i + 1 < len(nodes):
-                fallthrough = nodes[i + 1]
-                predecessors[fallthrough] = node
+    # Filter out unreachable nodes (except the entry & terminal nodes)
```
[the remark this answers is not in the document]
zbanks: Yeah, tbh I think I'd lean towards removing this filter entirely? (Though I'd want to check how often the current filter is needed to make the flowgraph reducible)

For now, it's needed so that `translate_graph_from_block` in translate.py can iteratively hit every node.

src/if_statements.py
```diff
@@ -762,29 +730,34 @@ def build_body(context: Context, function_info: FunctionInfo, options: Options)
...
-        elif isinstance(node, BasicNode) and node.is_loop():
-            context.loop_nodes.add(node.successor)
 
     if options.debug:
         print("Here's the whole function!\n")
+
     body: Body
-    if options.ifs:
-        body = build_flowgraph_between(context, start_node, return_node, 1)
+    if options.ifs and is_reducible:
```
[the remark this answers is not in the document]
zbanks: There are 15 examples across the oot/mm/papermario corpuses which are non-reducible, they seem to mostly be infinite loops (e.g. `Idle_ThreadEntry`, `crash_monitor_thread`) -- and there's a known path for making those graphs reducible in the future.

The goal of this PR is to make it so that `build_flowgraph_between` and all of the subsequent analysis can assume that the flowgraph is properly reducible. If that doesn't hold, it falls back to the "dumb" --gotos-only mode. 

It's theoretically possible to do the interval/structural analysis "around" the irreducible sections - but we'd need to be much more careful about the calculation & interpretation of `dominators`, `immediate_postdominators`, etc. in a way that I don't think is worth the complexity?

src/flow_graph.py
```diff
@@ -1056,95 +1137,125 @@ def duplicate_premature_returns(nodes: List[Node]) -> List[Node]:
...
+                    changes = True
+
+        # Compute immediate dominator, and the inverse relation
+        for n in nodes:
+            immediately_dominates(n).clear()
+        for n in nodes:
+            doms = dominators(n).difference({n})
+            # If `n == entry` or the flow graph is not reducible, `doms` may be empty.
+            # TODO: Infinite loops could be made reducible by introducing
+            # branches like `if (false) { return; }` without breaking semantics
```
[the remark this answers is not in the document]
zbanks: Related to my comment above - this PR makes it so we try to only do analysis on reducible flowgraphs where `dominators`, `immediate_postdominators`, etc. are all well-defined. 

So although it's a slight regression to leave these (rare) cases unsupported  - it makes the analysis that *is* done simpler & more robust.

src/flow_graph.py
```diff
@@ -764,27 +770,35 @@ class BaseNode:
...
+
 
 @attr.s(eq=False)
 class BasicNode(BaseNode):
     successor: "Node" = attr.ib()
 
-    def is_loop(self) -> bool:
-        return is_loop_edge(self, self.successor)
+    def children(self) -> Set["Node"]:
+        return {self.successor}
```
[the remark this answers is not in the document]
zbanks: Oh - as a trick to make more graphs reducible? To basically turn the node from `X; goto L; ` into `X; if (1) { goto L; }`?

src/if_statements.py
zbanks: Yeah - this fn & `SwitchStatement` *are* used here, but they could probably be replaced by a plain `Body`.

src/if_statements.py
zbanks: It's not equivalent - but I think it's the intended logic? 
Before, `is_loop()` referred to the tail of the loop: the BB with the backedge to the head node. But here, we don't actually mind incorporating the tail of the loop into the compound if -- we want to avoid bringing the *head* into the compound if. 
(I think an alternative would be just using the parents constraint: `len(next_node.parents) > 1 or next_node in next_node.parents`)

src/if_statements.py
zbanks: my printf debugging is exposed!

src/if_statements.py
zbanks: This shouldn't occur legitimately if our calculation of `immediate_postdominator` isn't buggy, I just wanted to include enough info to help debug the problem - should I maybe update the comment? change to an assert?

matt-kempster: an assert-with-error-message is the precedent in the codebase I think

zbanks: Yeah you're right, "this should be impossible" is better left as an assert I think

src/flow_graph.py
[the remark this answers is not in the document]
simonlindholm: Right, exactly. I don't know how that would interact with if_statements, or if it's there's a good place to set `emit_goto = True` within flow_graph if irreducibility is detected (maybe it's simpler to tell users to annotate manually anyway), but in principle I think that may be a good approach for dealing with irreducibility.

src/flow_graph.py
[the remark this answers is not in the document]
simonlindholm: yeah, I agree it should be rare in practice and not worth trying to detect. Can you add a comment about this being set only for loop heads?

tests/end_to_end/andor_return/irix-o2-out.c
```diff
@@ -0,0 +1,9 @@
+s32 test(s32 arg0, s32 arg1, s32 arg2, s32 arg3) {
+    if ((arg0 != 0) || (arg1 != 0)) {
+        if ((arg2 != 0) || (arg3 != 0)) {
+            return arg0 + arg1;
+        }
+        return arg2 + arg3;
+    }
+    return arg2 + arg3;
```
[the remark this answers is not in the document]
simonlindholm: A TODO sounds good! I think the existing duplicate return node emission might have been an earlier attempt at doing what `duplicate_premature_returns` does now? Or just a bug.

src/if_statements.py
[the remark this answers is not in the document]
simonlindholm: I agree it's not worth the complexity, I'm just wondering if it may work sometimes by accident anyway and it could be worth providing that option to the user. (The definition of a reducible cfg seemed pretty complex to me, with three distinct properties, and I was thinking maybe we only rely on e.g. two of them. It may be a naive question because I don't have intuition about reducible cfg's...)

tests/end_to_end/andor_return/irix-o2-out.c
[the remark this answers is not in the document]
zbanks: Cool, I left the TODO in `emit_node` so this can be revisited: https://github.com/matt-kempster/mips_to_c/pull/114/files#diff-fe6eb70d8a821eb6a7b0aa6194f261ff29d461f0bca93ef30bfb32b061f381e2R281-R282

src/flow_graph.py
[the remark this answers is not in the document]
zbanks: Adding TODO

src/if_statements.py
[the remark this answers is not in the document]
zbanks: I'm going to try to explain it better in the comments, but I think the big picture I was going for is that if the flow graph is reducible, now all of the `build_flowgraph_between`, `build_conditional_subgraph`, `build_switch_between` is pretty straightforward interval analysis with well-defined entry & exit nodes. 

I was able to simplify the logic in these functions because there are fewer edge cases -- but at the cost of not supporting irreducible graphs. The path forward for irreducible graphs is to make them reducible, rather than adding special cases to `build_*`.

simonlindholm (approved):

zbanks:
@matt-kempster I'll wait to merge this until you have a chance to re-review it (if you want to)

src/flow_graph.py
```diff
@@ -1056,95 +1159,129 @@ def duplicate_premature_returns(nodes: List[Node]) -> List[Node]:
...
+                dominators(n).update(nodes)
+
+        changes = True
+        while changes:
+            changes = False
+            for n in nodes:
+                if n == entry:
+                    continue
+                nset = dominators(n)
+                for p in parents(n):
```
matt-kempster, line 1218: nit (just noticed this was in the original code): this would be difficult to debug in pdb due to the use of `n` and `p`. Maybe add a TODO to change these variable names to `node` and `parent` respectively?

src/if_statements.py
```diff
@@ -167,6 +234,25 @@ def get_lone_if_statement(self) -> Optional[IfElseStatement]:
                 ret = statement
         return ret
 
+    def elide_empty_returns(self) -> None:
```
matt-kempster, line 237: should we expand this to also elide duplicate returns and/or warn about unreachable post-return statements?

src/if_statements.py
```diff
@@ -698,54 +666,54 @@ def emit_successor(node: Node, cur_index: int) -> None:
...
                 IfElseStatement(
                     block_info.branch_condition,
-                    1,
                     if_body=if_body,
                     else_body=None,
                 )
             )
             emit_successor(node.fallthrough_edge, i)
+        else:
+            assert isinstance(node, TerminalNode)
```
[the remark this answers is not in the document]
matt-kempster: agreed

src/if_statements.py
```diff
@@ -762,29 +738,34 @@ def build_body(context: Context, function_info: FunctionInfo, options: Options)
...
+
+    # Check no nodes were skipped: build_flowgraph_between should hit every node in
+    # well-formed (reducible) graphs; and build_naive explicitly emits every node
+    unemitted_nodes = (
+        set(context.flow_graph.nodes)
+        - context.emitted_nodes
+        - {context.flow_graph.terminal_node()}
+    )
+    for node in unemitted_nodes:
+        body.add_comment(f"bug: did not emit code for node #{node.name()}")
```
matt-kempster, line 760: I wonder if we should expand this to also include the contents of said block?

src/if_statements.py
```diff
@@ -182,167 +268,74 @@ def label_for_node(context: Context, node: Node) -> str:
...
+
+    For other nodes that were already emitted, instead emit a goto.
+    Since nodes represent positions in assembly, and we use phi's for preserved
+    variable contents, this will end up semantically equivalent. This can happen
+    sometimes when early returns/continues/|| are not detected correctly, and
+    this hints at that situation better than if we just blindly duplicate the block
+    """
+    if node in context.emitted_nodes:
+        # TODO: Treating ReturnNode as a special case and emitting it repeatedly
+        # hides the fact that we failed to fold the control flow. Maybe remove?
```
matt-kempster, line 287: Certainly we should emit a comment at the very least indicating duplicate emission?

matt-kempster (approved):
Looks absolutely great, thanks for working on this.

src/if_statements.py
matt-kempster: actually I think this is more appropriate for the `emit_node` discussion

src/flow_graph.py
zbanks: nbd to just fix it now - I'l rename them to `node` & `parent` (and `c` -> `child` for consistency)

src/if_statements.py
zbanks: Sure - not too hard.

src/if_statements.py
zbanks: I think this refactor addressed those issues we were having before where we would sometimes output `return; return; return;` (ex `func_80B8E84C` in MM) -- that stopped happening even before I added `elide_empty_returns`.

PR #115, opened 2021-06-22
3 cleanup tasks (flowgraph/tests)

zbanks:
In order by commits. The tests were re-blessed after each change, so you can see the results of each code change.

- Fix `run_tests.py` to find rodata/data for Paper Mario
    - Now papermario tests pass that depend on rodata, like `func_8013AA9C`, don't error out. (This doesn't show up in the repo though)
- `if (A) { B; } else { goto C; }` --> `if (!A) { goto C; } B;`
    - Could be `goto C;` or `return C;` -- as long as the body ends in a jump.
    - Output structure changes in some cases; `B` clause unindented by one.
    - The `andor-mixed` test (and `render_models` in Paper Mario) have (imo) more readable flow with `if (...) { goto ...; }` up front. The `goto`s are all now _forward_ jumps in the C source.
    - I pushed some logic out of `Body.add_if_else()` into `IfElseStatement.format()` because it requires the body to be built: the checks for the `sub_if` work "outside-in" so the outermost `IfElseStatement` needs to be built & processed first.
         - One minor downside is that now this logic comes *after* `elide_empty_returns()` is applied -- but I think this isn't too bad in practice.
- In `flow_graph.py`, calculate `BaseNode.parents` from `.children()` and remove `.add_parent()`
    - Assignment of phis depended on the order of `.parents`, so some suffixes got swapped around in the test outputs (e.g. `phi_a0_3`/`phi_a0_4`)
    - The `.parents` list is now sorted, which should make it more robust to future changes

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -182,7 +182,7 @@ def create_project_tests(
         if context_file is not None:
             flags.extend(["--context", str(context_file)])
 
-        # Guess the name of .data/.rodata file(s) for the OOT/MM decomp projects
+        # Guess the name of .data/.rodata file(s) for various decomp projects
         for candidate in [
             # mm code/*.asm
             "code_data_" + asm_name,
@@ -207,6 +207,14 @@ def create_project_tests(
             f = asm_file.parent / candidate
             if f.exists():
                 flags.extend(["--rodata", str(f)])
+        # papermario
+        papermario_data_path = Path(
+            str(asm_file).replace("/nonmatchings/", "/data/")
+        ).parent.parent
+        for f in papermario_data_path.glob("*.data.s"):
+            flags.extend(["--rodata", str(f)])
+        for f in papermario_data_path.glob("*.rodata.s"):
+            flags.extend(["--rodata", str(f)])
 
         test_path = asm_file.relative_to(asm_dir)
         name = f"{name_prefix}:{test_path}"
--- a/src/flow_graph.py
+++ b/src/flow_graph.py
@@ -778,9 +778,6 @@ class BaseNode(abc.ABC):
     # i.e. there is an invariant `(node.loop is None) or (node.loop.head is node)`
     loop: Optional["NaturalLoop"] = attr.ib(init=False, default=None)
 
-    def add_parent(self, parent: "Node") -> None:
-        self.parents.append(parent)
-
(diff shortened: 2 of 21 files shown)


simonlindholm:
If/else rearranging sounds dangerous -- won't this frequently cause blocks to be emitted in the wrong order? (Especially on IDO, whose block order matches C outside of rare cases involving switches.)

zbanks:
> If/else rearranging sounds dangerous -- won't this frequently cause blocks to be emitted in the wrong order? (Especially on IDO, whose block order matches C outside of rare cases involving switches.)

Thanks for bringing that up.
I took a closer look by modifying `run_tests.py` to capture the node output order (temporarily):
```diff
diff --git a/run_tests.py b/run_tests.py
index f594414..c4dea7f 100755
--- a/run_tests.py
+++ b/run_tests.py
@@ -89,6 +89,11 @@ def decompile_and_compare(
 
     final_contents = decompile_and_capture_output(options, test_case.brief_crashes)
 
+    options.debug = True
+    debug_contents = decompile_and_capture_output(options, test_case.brief_crashes)
+    node_order = ["", "// Node Order"] + [line.strip() for line in debug_contents.split("\n") if "// Node" in line] + [""]
+    final_contents += "\n".join(node_order)
+
     if test_options.should_overwrite:
         test_case.output_file.parent.mkdir(parents=True, exist_ok=True)
         test_case.output_file.write_text(final_contents)
```
---
As a baseline, the output order was usually sequential, with the major exceptions being complex switch statements & duplicated `ReturnNode`s (as expected).

I found that my commit `b267608` *usually* didn't change the output order, except when the swapped `else_body` was more than just a return or goto.

I refined the code so that the if/else negation only happens when the else body is *just* a goto/return: not when it *ends* in a goto/return. (If a body is solely a `goto`, I think it's more clear that it's not an actual node?)

I did some spot-checking with the newest version `6322c74eb` + the node ordering test patch above across MM, OOT, & PM: it looks like now the only change in output node order caused by this PR is that duplicated `ReturnNode`s are in different places. (Being duplicated, they already *were* out-of-order, now they're just *differently* out-of-order :smile: )

simonlindholm (approved):

PR #116, opened 2021-06-23
Prevent `Type.unify()` from recursing forever

### LINKED ISSUE #81: Type unification can recurse forever
A Type can be a pointer to itself. This can result in result in infinite recursion during unification.

(I don't have a concrete testcase for this, just theorizing. `a = &a; b = &b; a = b` might work.)

zbanks:
Fixes #81 and [Trello](https://trello.com/c/Z8ZhygMP/51-type-unification-can-recurse-forever).

The motivation for fixing this was the IDO recomp project, trying to run `./mips_to_c.py` on `uopt.s` from that lead to a *bunch* of these errors. After this PR IDO's `uopt.s` decompiles without `RecursionError` -- it only has 4 errors over the whole file.

Between the other projects (OOT, MM, PM), the only file that hit a `RecursionError` was MM's `boot/fault.asm` when run without context.

I made `Type.unify()` just give up if it detects any amount of recursion. In theory, there may be certain cases where it *could* succeed? 
I could forsee needing a more robust solution if we implemented "pure" `Type` representations of structs: but for now, recursive structs like `struct foo { struct foo * x; }` are all handled by the `CType` machinery. 

I ran this change against all of (OOT, MM, PM) with & without context, [the output is here](https://gist.github.com/zbanks/b56c6eb15cdac74ae0751f96ce7a5fef). Most of the diffs were in files without context, where mips2c had inferred some really wacky types. This diff leads to some `? *` -> `void *` or `void **`.

### DIFF
--- a/src/types.py
+++ b/src/types.py
@@ -69,16 +69,29 @@ class Type:
 
     _data: TypeData = attr.ib()
 
-    def unify(self, other: "Type") -> bool:
+    def unify(self, other: "Type", *, seen: Optional[Set["TypeData"]] = None) -> bool:
         """
         Try to set this type equal to another. Returns true on success.
         Once set equal, the types will always be equal (we use a union-find
         structure to ensure this).
+        The seen argument is used during recursion, to track which TypeData
+        objects have been encountered so far.
         """
+
         x = self.data()
         y = other.data()
         if x is y:
             return True
+
+        # If we hit a type that we have already seen, fail.
+        # TODO: Is there a looser check that would allow more types to unify?
+        if seen is None:
+            seen = {x, y}
+        elif x in seen or y in seen:
+            return False
+        else:
+            seen = seen | {x, y}
+
         if x.size is not None and y.size is not None and x.size != y.size:
             return False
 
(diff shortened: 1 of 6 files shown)


src/types.py
```diff
@@ -69,16 +69,29 @@ class Type:
...
+        objects have been encountered so far.
         """
+
         x = self.data()
         y = other.data()
         if x is y:
             return True
+
+        # If we hit a type that we have already seen, fail.
+        # TODO: Is there a looser check that would allow more types to unify?
```
simonlindholm, line 87: I think in principle it should be possible to recursively unify arbitrary types by moving recursive calls to after the union-find merge step, and then adding the revert machinery we talked about. Maybe worth doing just for fun, though I doubt there's any practical value to it.

simonlindholm (approved):
Hah, didn't think this would actually come up in practice.

zbanks:
> Hah, didn't think this would actually come up in practice.

It's funny, I think it mostly comes up unintentionally: mips2c makes a series of inferences that causes this mistake. (e.g. the 1 place this comes up in MM goes away after providing context)
This is sort of a patch to cover up that other weakness?

PR #117, opened 2021-06-24
Implement simple do-while loop functionality

matt-kempster:
An offshoot of #76, this implements simple do-while loop detection and output.

Should I add a new command-line option to disable this? It does add a new layer of nesting, since goto labels don't take up a nest level. I also don't know what the sentiment about this change is - perhaps it just gets in the way? (Nevertheless, it's important for #76, so that's why I'm breaking it off like this.)

I should probably run this on some entire projects to see what happens, also.

There's at least one confusing part (the `secretly=...` code), which I plan to document soon, but haven't found the right words to explain properly.

### DIFF
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -151,11 +151,34 @@ class LabelStatement:
         return "\n".join(lines)
 
 
+@attr.s
+class DoWhileLoop:
+    body: "Body" = attr.ib()
+    condition: Optional[Condition] = attr.ib(default=None)
+
+    def should_write(self) -> bool:
+        return True
+
+    def format(self, fmt: Formatter) -> str:
+        space = fmt.indent("")
+        after_do = f"\n{space}" if fmt.coding_style.newline_after_if else " "
+        cond = format_expr(self.condition, fmt) if self.condition else ""
+        with fmt.indented():
+            return "\n".join(
+                [
+                    f"{space}do{after_do}{{",
+                    self.body.format(fmt),
+                    f"{space}}} while ({cond});",
+                ]
+            )
+
+
 Statement = Union[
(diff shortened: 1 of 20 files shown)


src/if_statements.py
```diff
@@ -617,6 +691,22 @@ def build_flowgraph_between(context: Context, start: Node, end: Node) -> Body:
...
+        if not skip_loop_detection and isinstance(curr_start, ConditionalNode):
+            # breakpoint()
+            do_while_loop = detect_loop(context, curr_start, end)
+            if do_while_loop:
+                # breakpoint()
+                body.add_do_while_loop(do_while_loop)
+                # Move past the loop:
+                if curr_start.loop and curr_start.loop.is_self_loop():
+                    curr_start = curr_start.fallthrough_edge
+                else:
```
zbanks, line 703: It's possible that the conditional at the head of the loop is unrelated to the loop itself, unfortunately:

```diff
diff --git a/tests/end_to_end/loop_with_if/orig.c b/tests/end_to_end/loop_with_if/orig.c
index a42150f..172b179 100644
--- a/tests/end_to_end/loop_with_if/orig.c
+++ b/tests/end_to_end/loop_with_if/orig.c
@@ -9,5 +9,13 @@ int test(int length) {
         }
     }
 
+    do {
+        if (i == 7) {
+            i *= 3;
+        } else {
+            i += 6;
+        }
+    } while (i < length * 2);
+
     return i;
 }
```

2021-06-24-021128_628x447_scrot

Here, Node 7 is the head of the loop, and a conditional, but the conditional is the if statement and is unrelated to the loop itself. (The loop conditional is in node 11)

PR #118, opened 2021-06-24
Detect GlobalSymbol arrays & string constants

zbanks:
This is #112, without separating out the `Initializer` interface.

- Support for `.short` directive in `parse_file(...)`
- Detect global symbols that are actually arrays.
  - Added the `GlobalSymbol.array_dim` field, which defaults to `None` meaning "not an array." Before formatting, if this field gets populated, it's then assumed that the symbol is an array with the given length. (`0` can be used for arrays with unknown lengths).
      - This worked out pretty well, the only spot that really needed to be changed was the `StructAccess.format`
      - `GlobalSymbol.type` is otherwise unchanged.
  - Use heuristic to guess about zero-padding
      - Really bad at the BSS region, but pretty good otherwise
  - Support writing array initializers.
- Remove StringLiteral, do detection at format time
   - The only downside is that now we have to prune string literals during `GlobalInfo.global_decls()`, like we do with floats. (It's sort of tricky to mark which symbols have been completely elided)

---

Here is a somewhat representative diff from MM's `ovl_Bg_Dblue_Elevator_0x80B91F20`:
```diff
-static s8 D_80B929E0;
+static s8 D_80B929E0[2] = {(u8)0, (u8)2};
 static ? D_80B929E3;
-static s8 D_80B929E4 = (u8)0; // array: [4]
+static s8 D_80B929E4[6] = {(u8)0, (u8)1, (u8)2, (u8)3, (u8)4, (u8)5};
 static ? D_80B929EA;
-static InitChainEntry D_80B929EC; // array: [5]
-static f32 D_80B92A08 = 0.001f; // const; array: [2]
+static InitChainEntry D_80B929EC[5];
```


simonlindholm:
> ```diff
> +static s8 D_80B929E4[6] = {(u8)0, (u8)1, (u8)2, (u8)3, (u8)4, (u8)5};
> ```

What's up with the s8 vs u8 there, btw?

zbanks:
> > ```diff
> > +static s8 D_80B929E4[6] = {(u8)0, (u8)1, (u8)2, (u8)3, (u8)4, (u8)5};
> > ```
> 
> What's up with the s8 vs u8 there, btw?

`Literal`s always get formatted as unsigned -- https://github.com/matt-kempster/mips_to_c/blob/master/src/translate.py#L1114-L1122 .

src/translate.py
```diff
@@ -3769,59 +3783,75 @@ def read_pointer() -> Optional[Expression]:
...
+                is_const = (
+                    sym.asm_data_entry is not None
+                ) and sym.asm_data_entry.is_readonly
+
+                sort_order = (
+                    not sym.type.is_function(),
+                    is_global,
+                    is_in_file,
+                    is_const,
+                    name,
```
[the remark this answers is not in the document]
zbanks: Oh that's a great idea. The orderings probably start the same, but they would diverge when decompers rename symbols.

zbanks:
Thanks for the feedback so far - I'm still working on cleaning up ` global_decls()`. 
I think the big thing I need to clarify/fix is that `global_decls()` was written with the assumption that all `GlobalSymbol.array_dim`s were `None` at the start of that function, and it's responsibility was to *set* `array_dim`. 
(This assumption is violated by symbols whose types are available in the context)

tests/end_to_end/global_decls/irix-o2-out.c
```diff
@@ -2,13 +2,17 @@ MIPS2C_UNK extern_fn(struct A *); // extern
 MIPS2C_UNK static_fn(struct A *); // static
 extern f32 extern_float;
 struct A *static_A_ptr = &static_A;
+s32 static_array[3] = {2, 4, 6};
+s32 static_bss_array[3] = {0, 0, 0};
```
simonlindholm, line 6: it would be nice to eliminate this initializer -- while GCC defaults to placing global variables explicitly initialized to zero in bss, IDO doesn't, and only places variables without explicit initializers there.

zbanks: Ooh nice insight. 
(This test didn't make that very obvious because the test is hand-edited to add the `rodata`/`data`/`bss` sections)

src/translate.py
```diff
@@ -3769,59 +3783,75 @@ def read_pointer() -> Optional[Expression]:
...
+        if sym.array_dim is None:
+            return for_type(sym.type)
+        else:
+            elements: List[str] = []
+            for _ in range(sym.array_dim):
+                el = for_type(sym.type)
+                if el is None:
+                    return None
+                elements.append(el)
+            return f"{{{', '.join(elements)}}}"
```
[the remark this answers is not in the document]
zbanks: I refactored this slightly so that when CType arrays are supported, they'll use nearly the same code path.

(For now I'm not supporting CTypes at all - but that shouldn't be too bad to add in a future PR)

zbanks:
I did a pretty big cleanup of the `array_dim` logic in `global_decls()` that I think should be easier to follow now.
I also made it more explicit that we only need to detect `array_dim` when its not provided by the context: which helps reduce the input space a bit.

src/translate.py
```diff
@@ -3832,36 +3871,80 @@ def global_decls(self, fmt: Formatter) -> str:
...
-                        if data_size // type_size > 1:
-                            comments.append(f"array: [{data_size // type_size}]")
+                # Try to guess the symbol's `array_dim` if we have a data entry for it,
+                # and it does not exist in the typemap (`not is_global`) or dim is unknown
+                # (Otherwise, if the dim is provided by the typemap, we trust it.)
+                if sym.asm_data_entry and (not is_global or sym.array_dim == 0):
+                    assert sym.array_dim is None or sym.array_dim == 0
+                    # The size of the data entry is uncertain, because of padding
+                    # between sections. Generally `(max_data_size - data_size) < 16`.
+                    data_size, max_data_size = sym.asm_data_entry.size_range_bytes()
```
simonlindholm, line 3885: rename data_size -> min_data_size, and create a data_size member further down just above where it's used

src/translate.py
```diff
@@ -3832,36 +3871,80 @@ def global_decls(self, fmt: Formatter) -> str:
...
+                        # Uh-oh! The type is too big for our data. (not an array)
+                        comments.append(
+                            f"type too large by {type_size - max_data_size}"
+                        )
+                    elif type_size == max_data_size:
+                        # The type perfectly fits our data: 1 element, or not an array
+                        if sym.array_dim == 0:
+                            sym.array_dim = 1
+                        else:
+                            sym.array_dim = None
```
simonlindholm, line 3902: (this else is silly)

src/parse_file.py
```diff
@@ -47,15 +47,30 @@ class AsmDataEntry:
     data: List[Union[str, bytes]] = attr.ib(factory=list)
     is_string: bool = attr.ib(default=False)
     is_readonly: bool = attr.ib(default=False)
+    is_bss: bool = attr.ib(default=False)
 
-    def size_bytes(self) -> int:
-        size = 0
+    def size_range_bytes(self) -> Tuple[int, int]:
```
simonlindholm, line 52: should probably avoid returning zero for the lower bound here, unless the upper bound is also zero.

src/translate.py
```diff
@@ -3832,36 +3871,80 @@ def global_decls(self, fmt: Formatter) -> str:
...
+                            # How many extra bytes do we need to add to `data_size`
+                            # to make it an exact multiple of `type_size`?
+                            extra_bytes = type_size - (data_size % type_size)
+                            if data_size + extra_bytes <= max_data_size:
+                                # We can make an exact multiple by taking some of the bytes
+                                # we thought were padding
+                                data_size += extra_bytes
+                            else:
+                                comments.append(f"extra bytes: {data_size % type_size}")
+                        if data_size // type_size > 1:
```
simonlindholm, line 3917: or array_dim == 0

src/translate.py
```diff
@@ -3832,36 +3871,80 @@ def global_decls(self, fmt: Formatter) -> str:
...
+                    type_size = (type_size_bits or 0) // 8
+                    if not type_size_bits:
+                        # If we don't know the type, we can't guess the array_dim
+                        pass
+                    elif type_size > max_data_size:
+                        # Uh-oh! The type is too big for our data. (not an array)
+                        comments.append(
+                            f"type too large by {type_size - max_data_size}"
+                        )
+                    elif type_size == max_data_size:
```
simonlindholm, line 3897: we shouldn't need this case

simonlindholm (approved):

zbanks:
Tested that my latest commit didn't introduce any regressions by running `./run_tests.py --project-with-context ../oot`

PR #119, opened 2021-06-25
Refactor CLI (while maintaining compatability)

zbanks:
[Trello ticket](https://trello.com/c/toC3va4m/137-improve-cli-to-support-multiple-input-output-files)

- Add argument groups, to make `--help` easier to read
- New (preferred) format is `./mips_to_c.py -f fn -f fn asm.s asm2.s ...`
- Detect if the last argument is "function-like", to keep compatible (no breaking changes, unless there's a bug)

In discord there was talk of`./mips_to_c.py files+ functions*`; but this PR encourages `./mips_to_c.py files+ (-f function)*` and supports `./mips_to_c.py files+ function?` for backwards compatibility. I did this because it seemed less fragile, but still good UX? I could easily change it to `files+ functions*` though without the `-f` instead, if that's preferred.

I rearranged the args to put them in some top-level groups, but I know the diff is tricky to read. The only change I made to the moved args was to put `help=` last for consistency.

---
New help messages:
```
$ ./mips_to_c.py 
usage: mips_to_c.py [--context C_FILE] [--emit-globals] [--valid-syntax] [-f FN ...] filename [filename ...]
mips_to_c.py: error: the following arguments are required: filename

$ ./mips_to_c.py -h
usage: mips_to_c.py [--context C_FILE] [--emit-globals] [--valid-syntax] [-f FN ...] filename [filename ...]

Decompile MIPS assembly to C.

optional arguments:
  -h, --help            show this help message and exit

Input Options:
  filename              input asm filename(s)
  --context C_FILE      read variable types/function signatures/structs from an existing C file. The file must already have
                        been processed by the C preprocessor.
  -D DEFINED            mark preprocessor constant as defined
  -U UNDEFINED          mark preprocessor constant as undefined

Output & Formatting Options:
  -f FN, --function FN  function index or name to decompile
  --valid-syntax        emit valid C syntax, using macros to indicate unknown types or other unusual statements. Macro
                        definitions are in `mips2c_macros.h`.
  --emit-globals        emit global declarations with inferred types.
  --debug               print debug info inline
  --print-assembly      print assembly of function to decompile
  --allman              put braces on separate lines
  --dump-typemap        dump information about all functions and structs from the provided C context. Mainly useful for
                        debugging.
  --visualize           display a visualization of the control flow graph using graphviz

Analysis Options:
  --stop-on-error       stop when encountering any error
  --void                assume the decompiled function returns void
  --gotos-only          disable control flow generation; emit gotos for everything
  --no-andor            disable detection of &&/||
  --no-casts            don't emit any type casts
  --reg-vars REGISTERS  use single variables instead of temps/phis for the given registers (comma separated)
  --goto PATTERN        emit gotos for branches on lines containing this substring (possibly within a comment). Default:
                        "GOTO". Multiple patterns are allowed.

```

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -1,4 +1,5 @@
 import argparse
+import collections
 import re
 import sys
 import traceback
@@ -39,8 +40,7 @@ def print_exception(sanitize: bool) -> None:
 
 
 def run(options: Options) -> int:
-    # We are relying on Python 3.7 dicts preserving insertion order
-    all_functions: Dict[str, Function] = {}
+    all_functions: collections.OrderedDict[str, Function] = collections.OrderedDict()
     asm_data = AsmData()
     typemap: Optional[TypeMap] = None
     try:
@@ -322,7 +322,8 @@ def parse_flags(flags: List[str]) -> Options:
     assert filenames, "checked by argparse, nargs='+'"
     if filenames[-1] == "all":
         filenames.pop()
-    elif re.match(r"[0-9a-zA-Z_]+", filenames[-1]):
+    elif re.match(r"^[0-9a-zA-Z_]+$", filenames[-1]):
+        # The filename is a valid C identifier or a number
         args.functions.append(filenames.pop())
     if not filenames:
         parser.error("the following arguments are required: filename")
--- a/src/main.py
+++ b/src/main.py
@@ -315,7 +315,6 @@ def parse_flags(flags: List[str]) -> Options:
         newline_after_if=args.allman,
         newline_before_else=args.allman,
     )
-    functions: List[Union[int, str]] = args.functions
     filenames = args.filename
 
     # Backwards compatability: giving a function index/name as a final argument, or "all"
@@ -328,7 +327,7 @@ def parse_flags(flags: List[str]) -> Options:
     if not filenames:
         parser.error("the following arguments are required: filename")
 
-    functions = []
+    functions: List[Union[int, str]] = []
     for fn in args.functions:
         try:
             functions.append(int(fn))
--- a/src/main.py
+++ b/src/main.py
@@ -1,5 +1,4 @@
 import argparse
-import collections
 import re
 import sys
 import traceback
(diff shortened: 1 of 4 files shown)


src/main.py
```diff
@@ -38,20 +39,19 @@ def print_exception(sanitize: bool) -> None:
 
 
 def run(options: Options) -> int:
-    mips_file: MIPSFile
```
simonlindholm, line 41: while Python 3.6 doesn't have ordered dicts in its formal spec, all implementations of it do. I think we rely on that in other places already

src/main.py
```diff
@@ -144,50 +143,151 @@ def run(options: Options) -> int:
...
+    group.add_argument(
+        "filename",
+        nargs="+",
+        help="input asm filename(s)",
+    )
+    group.add_argument(
+        "--rodata",
+        dest="filename",
+        action="append",
+        help=argparse.SUPPRESS,  # For backwards compatability
```
simonlindholm, line 161:
```suggestion
        help=argparse.SUPPRESS,  # For backwards compatibility
```

src/main.py
```diff
@@ -144,50 +143,151 @@ def run(options: Options) -> int:
...
 
 def parse_flags(flags: List[str]) -> Options:
-    parser = argparse.ArgumentParser(description="Decompile MIPS assembly to C.")
-    parser.add_argument("filename", help="input filename")
-    parser.add_argument("function", help="function index or name", nargs="?")
-    parser.add_argument(
-        "--debug", dest="debug", help="print debug info", action="store_true"
+    parser = argparse.ArgumentParser(
+        description="Decompile MIPS assembly to C.",
+        usage="%(prog)s [--context C_FILE] [--emit-globals] [--valid-syntax] [-f FN ...] filename [filename ...]",
```
simonlindholm, line 148: IMO remove --valid-syntax, it's a niche use-case and flags like --void should be just as common

(also I wonder if --emit-globals should be made the default)

simonlindholm (approved):

PR #121, opened 2021-06-30
Use dataclasses instead of attrs

simonlindholm:
Something weird is happening with
```
@dataclass(frozen=True, eq=False)
class RegisterVar(Expression):
    type: Type
    register: Register
```
where if the two fields swap places there's an error
```
Traceback (most recent call last):
  File "./mips_to_c.py", line 2, in <module>
    from src.main import main
  File "./mips_to_c/src/main.py", line 11, in <module>
    from .if_statements import get_function_text
  File "./mips_to_c/src/if_statements.py", line 15, in <module>
    from .translate import (
  File "./mips_to_c/src/translate.py", line 948, in <module>
    class RegisterVar(Expression):
  File "/usr/lib/python3.7/dataclasses.py", line 1002, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
  File "/usr/lib/python3.7/dataclasses.py", line 921, in _process_class
    globals,
  File "/usr/lib/python3.7/dataclasses.py", line 491, in _init_fn
    raise TypeError(f'non-default argument {f.name!r} '
TypeError: non-default argument 'type' follows default argument
```

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -19,7 +19,7 @@ jobs:
         python-version: 3.6
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade mypy dataclasses pycparser coverage
+        python -m pip install --upgrade mypy dataclasses types-dataclasses pycparser coverage
     - name: Run tests
       run: |
         ./run_tests.py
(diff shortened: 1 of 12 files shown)


zbanks (approved):
Looks great!

From my reading of [the docs for 3.7 dataclasses](https://docs.python.org/3.7/library/dataclasses.html#init-only-variables), I didn't see anything that'd give us issues.

PR #122, opened 2021-06-30
Detect nested &&/|| expressions in if statements

zbanks:
This logic supersedes the previous `get_andor_if_statement`, while being generic enough to handle plain single ifs & self-loops.

It doesn't seem to interfere with the do-while loop detection, in the future it could be extended to collapse those conditionals too?

~I think this PR is mostly feature-complete, but I want to continue to work on the documentation & tests. (I feel bad for destroying the super useful diagram in `get_andor_if_statement()` -- but it was no longer very relevant.)~

~One concern I have is that the logic may get carried away, and incorporate too much into the conditional expression. Maybe I need to add a way to limit the size of the conditional expression before merging this PR?~

I updated the PR to avoid putting comma expressions inside the conditional if the outermost conditional operator is `&&`, following the previous behavior. This splitting only happens with `&&` because `if (x && (foo(), y)) ...` can be represented by nested ifs, but representing `if (x || (foo(), y))` as multiple ifs would require `goto` or a temporary variable.


run_tests.py
```diff
@@ -81,10 +81,11 @@ def decompile_and_compare(
             return None, f"{test_case.output_file} does not exist. Skippping."
         original_contents = "(file did not exist)"
 
-    test_flags = ["--sanitize-tracebacks", "--stop-on-error", str(test_case.asm_file)]
+    test_flags = ["--sanitize-tracebacks", "--stop-on-error"]
     test_flags.extend(test_case.flags)
     if test_case.flags_path is not None:
         test_flags.extend(get_test_flags(test_case.flags_path))
+    test_flags.append(str(test_case.asm_file))
     options = parse_flags(test_flags)
```
zbanks, line 89: The arg parser doesn't like `--flag`s in the middle of the positional args, like `filename1 --debug filename2`. (the `test_case.flags` can contain paths to other asm files)

src/translate.py
```diff
@@ -733,11 +733,23 @@ def f64(left: Expression, op: str, right: Expression) -> "BinaryOp":
...
+            and isinstance(self.left, Condition)
+            and isinstance(self.right, Condition)
+        ):
+            # DeMorgan's Laws
+            return BinaryOp(
+                left=self.left.negated(),
+                op={"&&": "||", "||": "&&"}[self.op],
+                right=self.right.negated(),
+                type=Type.bool(),
+            )
```
zbanks, line 751: @simonlindholm or @matt-kempster - do you know if this was *intentionally* unimplemented? Were there cases when you could tell IDO didn't simplify the boolean expression? 

From my testing, in conditionals, it looks like IDO always followed the simplification in order to make the control flow graph work. 

Having this done after constructing the conditional is really convenient: there's less state to pass around as the conditional is constructed.

src/translate.py
```diff
@@ -733,11 +733,23 @@ def f64(left: Expression, op: str, right: Expression) -> "BinaryOp":
             floating=True,
         )
 
-    def is_boolean(self) -> bool:
```
zbanks, line 736: `&&` and `||` are "boolean", but the existing logic only cared about comparisons.

tests/end_to_end/empty-if/manual-out.c
```diff
@@ -1,5 +1,5 @@
 void test(s32 arg0, s32 arg1) {
-    if ((arg0 == 1) || (arg1 != 2)) {
```
zbanks, line 2: This *looks* bad but the underlying assembly is actually ambiguous -- either is fine. (The if block is empty)

tests/end_to_end/doubles/irix-o2-mips1-out.c
```diff
@@ -2,7 +2,7 @@ f64 test(f64 arg0, s32 arg2, f64 arg4) {
     f64 temp_f0;
 
     temp_f0 = (((f64) arg2 * arg0) + (arg0 / arg4)) - 7.0;
-    if (!(temp_f0 < arg4)) {
+    if (temp_f0 >= arg4) {
```
simonlindholm, line 5: this looks wrong... float comparisons can't be negated like this, we're supposed to have logic that makes sure that doesn't happen

src/if_statements.py
```diff
@@ -421,162 +357,220 @@ def gather_any_comma_conditions(block_info: BlockInfo) -> Condition:
...
+        # that are part of the outermost && expression, if it exists.
+        # Complex && conditions are better represented by nested ifs
+        if else_node is end:
+            outermost_has_comma = False
+            node: Node = start
+            while node in chained_cond_nodes:
+                assert isinstance(node, ConditionalNode)
+                if node.conditional_edge is end:
+                    block_info = node.block.block_info
+                    assert block_info
```
simonlindholm, line 426:
```suggestion
                    assert isinstance(block_info, BlockInfo)
```
otherwise block_info will be Any

src/if_statements.py
```diff
@@ -421,162 +357,220 @@ def gather_any_comma_conditions(block_info: BlockInfo) -> Condition:
...
+                    # This node is part of the outermost boolean expression (which might be &&).
```
simonlindholm, line 427: What does "outermost boolean expression" mean exactly? First node of the outermost layer of &&/||, right?

either way I don't think it's necessarily true -- in `if (((a && b) || (c && d)) && e)` c jumps to end, is reachable via conditional edge, but isn't top-level.

src/if_statements.py
```diff
@@ -421,162 +357,220 @@ def gather_any_comma_conditions(block_info: BlockInfo) -> Condition:
...
+                    # Check if it would use a comma expression if included in the conditional.
+                    if node is not start and next(
+                        (True for b in block_info.to_write if b.should_write()), False
+                    ):
```
simonlindholm, line 431:
```suggestion
                    if node is not start and any(b.should_write() for b in block_info.to_write):
```
(may be nice to add a helper for this on block_info)

tests/end_to_end/andor_mixed/orig.c
```diff
@@ -8,7 +8,23 @@ int test(int a, int b, int c, int d) {
```
simonlindholm, line 8: can you add a test for `if (a && (b ? c : d) == e)` to check that nothing bad happens?

simonlindholm (commented):
(halfway through)

tests/end_to_end/andor_mixed/irix-o2-out.c
```diff
@@ -1,21 +1,20 @@
...
+    if ((arg0 != 0) && ((arg1 != 0) || (arg2 != 0)) && ((arg3 != 0) || ((arg0 + 1) != 0))) {
+        phi_v1 = 2;
+    }
+    if (((arg0 != 0) && (arg3 != 0)) || (((arg1 != 0) || (arg2 != 0)) && ((arg0 + 1) != 0))) {
+        phi_v1 = 3;
+    }
+    if ((arg0 != 0) && (arg1 != 0) && ((arg2 != 0) || (arg3 != 0)) && (((arg0 + 1) != 0) || ((arg1 + 1) != 0))) {
+        phi_v1 = 4;
+    }
+    if ((((arg0 != 0) || (arg1 != 0)) && (arg2 != 0)) || ((arg3 != 0) && ((arg0 + 1) != 0)) || ((arg1 + 1) != 0) || ((arg2 + 1) != 0)) {
```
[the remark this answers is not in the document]
zbanks: Yeah - it's not so bad if remove all the ` != 0`s and then use an editor to help match parentheses.

src/if_statements.py
zbanks: You're right, this test is wrong.
I think the condition I actually want to check for comma conditions on is "the set of all nodes that are on all paths from `start` to `if_node` (excluding `start` itself)".

If the outermost layer is `||`, this set will be empty because there are nonoverlapping paths from `start` to `if_node`.

src/if_statements.py
```diff
@@ -355,228 +343,225 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
...
+                    #   }
+                    # ```
+                    # ...which can be re-written into a single if statement with `||`.
+                    # ```
+                    #   if (or_terms || !tail_cond) {
+                    #       if_node;
+                    #   } else {
+                    #       next_node;
+                    #   }
+                    # ```
```
simonlindholm, line 522: This is a very nice comment! Unfortunately I don't really get it because I don't understand what `traverse` recursively does, e.g. why we pass in conditional_edge to it and why the if statement ends up being `if (tail_cond) { next_node; } else { if_node; }`...

simonlindholm (commented):
I don't understand traverse, at all... :(

src/if_statements.py
```diff
@@ -355,228 +343,225 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
...
-            or (
-                next_node.conditional_edge is not bottom
-                and next_node.fallthrough_edge is not bottom
-            )
-            # An edge-case of our pattern-matching technology: without
-            # this, self-loops match the pattern indefinitely, since a
-            # self-loop node's conditional edge points to itself.
-            or next_node.loop
-            or len(next_node.parents) > 1
+        if end not in curr_node.fallthrough_edge.postdominators:
```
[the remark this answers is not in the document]
zbanks: I tried to carefully structure this loop to ensure that `start` was always included, and `end` was always excluded. But it made this loop hard to read (the interleaving of the conditions & assignments is bad). It does handle the case you describe but it's not clear that it does.

I have a refactor of the loop. `end not in curr_node.fallthrough_edge.postdominators` became `end not in curr_node.postdominators or end is curr_node` here.

src/if_statements.py
```diff
@@ -355,228 +343,225 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
...
+            # Shorten the chain by removing the last node, then try again.
+            chained_cond_nodes.pop()
+            continue
+
+        # No changes were made: we found the largest chain that matches the constraints
+
+        def traverse(
+            node: Node, if_node: Node, else_node: Node
+        ) -> Tuple[Node, Condition]:
+            """Iterate through the chain of conditionals, until we hit one of if/else body nodes"""
```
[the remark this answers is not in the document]
zbanks: Here `node` could equal `if_node` if `node == start and chained_cond_nodes = [start]` -- it was the one-node self-loop case.

But in refactoring this into a global fn, I made it so `traverse` isn't used for self loops.

src/if_statements.py
```diff
@@ -355,228 +343,225 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
...
+                    #   if (or_terms || !tail_cond) {
+                    #       if_node;
+                    #   } else {
+                    #       next_node;
+                    #   }
+                    # ```
+                    # We continue iterating because, unlike above, we have not encountered the if/else nodes.
+                    node, tail_cond = traverse(node, node.conditional_edge, if_node)
+                    or_terms.append(tail_cond.negated())
+            return node, join_conditions(or_terms, "||")
```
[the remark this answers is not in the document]
zbanks: Yeah -- both cases happen, and `||` is correct for them.

tests/end_to_end/andor_mixed/irix-o2-out.c
```diff
@@ -1,23 +1,36 @@
...
+        phi_v1 = 5;
+    }
+    if (((arg0 == 0) || (arg1 == 0)) && ((arg2 != 0) && (arg3 != 0) && (((arg0 + 1) != 0) || ((arg1 + 1) != 0)))) {
+        phi_v1 = 6;
+    }
```
zbanks, line 24: Crap - this one is not equivalent!
Original:
`if (((a && b) || (c && d)) && ((a + 1) || (b + 1))) `

simonlindholm: Hah, good catch! I wonder if we should do some fuzz-testing here.

zbanks:
Okay - I totally changed the algorithm. I think it's more readable now, and it's _more_ correct (hopefully *fully* correct :smile: ).

I think the issue I had with this algorithm for building the complex conditional is that I wanted something that knew would succeed. The recursive strategy in the previous version couldn't fail: once it got to the recursive bit, it would always produce a single condition. 
It was much easier to collapse two nodes at a time, and do fixed-point iteration. It's less efficient and may not succeed -- but it's much easier to explain.

src/if_statements.py
```diff
@@ -355,228 +343,240 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
...
+        for node in node_cond_edges:
+            for child in node_cond_edges[node][1:]:
+                if child not in (if_node, else_node):
+                    assert isinstance(child, ConditionalNode)
+                    node_parents[child].append(node)
+
+        # Find the first pair of nodes which form a reducible pair: one will always
+        # be the *only* parent of the other.
+        # Note: we do not include `if_node` or `else_node` in this search
+        for child, parents in node_parents.items():
```
zbanks, line 437: I originally thought that I could make this a bit more efficient by iterating through the nodes _backwards_, so that I wouldn't have to recalculate `node_parents` after every change. But, this produces "right heavy" or irregular `BinaryOp` trees.

There are a few places that assume "left heavy" `BinaryOp` trees, so I didn't think the perf was worth the code complexity.

simonlindholm: The algorithm is O(n^4) in the worst-case (lots of consecutive ConditionalNode's that don't fold into a joined conditions, but with dominators that still join them together into a single group), but it's probably not that bad in practice. Calculating `node_parents` incrementally and fixing up irregular BinaryOp's once we've found a solution (not too hard recursively) would make it O(n^3), but I think it's fair to put that off until it becomes a problem.

Changing the &&-with-comma-conditions logic to construct nested if's instead of just dropping the last node and trying again would also help perf in practice, but not asymptotically.

simonlindholm (approved):
This is so much more understandable.

PR #123, opened 2021-06-30
Cast pointers to `intptr_t` instead of `s32`

zbanks:
From [Discord](https://discord.com/channels/688807550715560050/688851997352263820/859842428126888026)

> Isghj — Today at 1:05 PM
> `SkelAnime_InitSV(globalCtx, temp_s0, (void *)0x6005F48, (void *)0x6000200, (s32) this->limbDrawTbl & ~0xF, (s32) (this + 0x2A0) & ~0xF, 0x16);`
> ???
> have never seen a mask on a limb table pointer?
> Kenix — Today at 1:06 PM
> Interesting. Looks like pointer alignment.
> also use (u)intptr_t rather than cast to s32
> Isghj — Today at 1:06 PM
> oh that was mips2c output
> Kenix — Today at 1:06 PM
> :thumbsup:
> @zbanks Can we change m2c to use intptr_t rather than s32 casts? :smile:

This looks like it was from `ovl_En_Yb_0x80BFA100`, which has the following diff when run with with this patch:
```diff
@@ -39,7 +39,7 @@

     Actor_SetScale(arg0, 0.01f);
     ActorShape_Init(&arg0->shape, 0.0f, func_80BFA350, 20.0f);
     temp_s0 = arg0 + 0x144;
-    SkelAnime_InitSV(arg1, (SkelAnime *) temp_s0, (void *)0x6005F48, (void *)0x6000200, (Vec3s *) ((s32) (arg0 + 0x188) & ~0xF), (Vec3s *) ((s32) (arg0 + 0x2A0) & ~0xF), 0x16);
+    SkelAnime_InitSV(arg1, (SkelAnime *) temp_s0, (void *)0x6005F48, (void *)0x6000200, (Vec3s *) ((intptr_t) (arg0 + 0x188) & ~0xF), (Vec3s *) ((intptr_t) (arg0 + 0x2A0) & ~0xF), 0x16);
     SkelAnime_ChangeAnimDefaultRepeat((SkelAnime *) temp_s0, (void *)0x6000200);
     Collider_InitAndSetCylinder(arg1, arg0 + 0x3B8, arg0, &D_80BFB2B0);
     arg0->colChkInfo.mass = 0xFF;
```
----
I'm not 100% sure this is a good idea - I wanted to open it up for discussion.
It seems pretty benign, but I think the two downsides are:
- it assumes that `uintptr_t` and `intptr_t` are already typedef'd to be `u32` and `s32` respectively
- it doesn't treat these types as part of the `Type` system, they only exist when casting.


zbanks:
Punting for now - maybe we can revisit it when we handle other type synonyms like enum constants.

Also from Simon:
> seems like `uintptr_t` is typically what you want, too, while m2c will prefer `intptr_t`
> it not being integrated into the type system also seems potentially weird

PR #124, opened 2021-07-01
Resolve struct/union CTypes to a canonical name

zbanks:
This PR is a setup for emitting global initializers for structs.

Before this change, if a struct were declared before its typedef, then the `CType`s for `struct Foo` and `Foo` would not be equivalent and would not unify (even with `resolve_typedefs()`). 

As an example:
```c
struct Actor;
typedef struct Actor {
    // ...
} Actor;

// a & b have equivalent types in C, but had incompatible CTypes
struct Actor * a;
Actor *b;
```

This change adds `Struct.type`, which represents the "canonical" CType for the struct, which is preferentially a typedef.
The diff to the OOT tests is [here](https://gist.github.com/zbanks/359461ff8786e7ed67c35a814954734f).


simonlindholm (approved):
This seems fine. I think an alternative fix would be to treat `struct A { ... }` and `struct A` as equal in `equal_types`?

simonlindholm:
on a related note `equal_types` should really compare anonymous structs by object identity, too

zbanks:
I did initially try to put the logic in `equal_types`, but the logic got somewhat complicated and `ca.Struct`-specific (e.g. if the `name`s are available, ignore the `decls`s)

simonlindholm:
I think it would be totally fine to special case `ca.Struct` in `equal_types` fwiw, it's only super generic out of laziness. But eh, not worth spending time on given the merged fix

PR #125, opened 2021-07-02
Website features

zbanks:
- "Visualize" button that renders an SVG (requires graphviz)
    - Do you have `graphviz` installed on the server?
    - @simonlindholm Are you OK with the security implications of this? The library's render operations all shell out to a graphviz executable. I could remove this, or have it disabled in `website.py` unless an environment variable is set?
- Checkbox for `--emit-globals`, enabled by default
- Dropdown for selecting function to decomp, default to `all`. (Before it would decomp the first function defined)
   - There's a regex in JS to pick out function names from the assembly, looking for "glabel". It has a lot of false positives (all variable labels), but I think it's an improvement?

### DIFF
--- a/website.py
+++ b/website.py
@@ -89,15 +89,15 @@ try:
             if "dark" in form:
                 print(
                     """
-    <head>
-    <style>
-    body {
-        background-color: #454545;
-        color: #c0c0c0;
-    }
-    </style>
-    </head>
-    """
+<head>
+<style>
+body {
+    background-color: #454545;
+    color: #c0c0c0;
+}
+</style>
+</head>
+"""
                 )
             print("<body><pre><plaintext>", end="")
             print(res.stdout.decode("utf-8", "replace"))
@@ -167,19 +167,19 @@ label {
 <body>
 <form action="?go" method="post">
 <div class="main">
-<div>
+  <div>
     MIPS assembly:
-</div>
-<div style="flex: 20;">
+  </div>
+  <div style="flex: 20;">
     <textarea name="source"></textarea>
-</div>
-<div style="margin-top: 10px;">
+  </div>
+  <div style="margin-top: 10px;">
     Existing C source, preprocessed (optional):
-</div>
-<div style="flex: 11;">
+  </div>
+  <div style="flex: 11;">
     <textarea name="context"></textarea>
-</div>
-<div style="margin-top: 10px;" id="options">
+  </div>
+  <div style="margin-top: 10px;" id="options">
     <input type="submit" value="Decompile">
     <input type="submit" name="visualize" value="Visualize">
     <label>Function: <select name="functionselect"></select></label>
@@ -202,15 +202,15 @@ label {
     <label><input type="checkbox" name="noifs">Use gotos for everything</label> (to use a goto for a single branch, add "# GOTO" to the asm)
     <label><input type="checkbox" name="usesidebar">Output sidebar</label>
     <label><input type="checkbox" name="dark">Dark mode</label>
-</div>
+  </div>
 </div>
 <div class="sidebar">
-<div>
+  <div>
     Output:
-</div>
-<div style="flex: 1;">
+  </div>
+  <div style="flex: 1;">
     <iframe src="about:blank" name="outputframe"></iframe>
-</div>
+  </div>
 </div>
 <script>
 var formEl = document.getElementsByTagName("form")[0]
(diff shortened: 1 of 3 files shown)


website.py
```diff
@@ -222,6 +243,24 @@
 updateDarkMode();
 darkModeCheckbox.addEventListener("change", updateDarkMode);
 
+function updateFunctions() {
+    var functionSelect = document.getElementsByName("functionselect")[0];
+    var prevValue = functionSelect.value;
+    var options = "<option value='all'>all functions</option>";
+    for (var line of sourceEl.value.split("\\n")) {
+        //var match = line.match(/^\\w*glabel\\w*([A-Za-z0-9_]+)/)
```
simonlindholm, line 253: debug remnants

website.py
```diff
@@ -222,6 +245,24 @@
...
+        var match = line.match(/^\\s*glabel\\s+([A-Za-z0-9_]+)/)
+        if (match) {
+            var name = match[1];
+            options += "<option value='" + name + "' " + (name == prevValue ? "selected" : "") + ">" + name + "</option>";
```
simonlindholm, line 257: can you use document.createElement here instead, to avoid the XSS? (not that it matters in practice)

website.py
```diff
@@ -37,6 +39,13 @@
         cmd.append("--allman")
     if "leftptr" in form:
         cmd.extend(["--pointer-style", "left"])
+    if "globals" in form:
+        cmd.append("--emit-globals")
+    if "visualize" in form:
+        cmd.append("--visualize")
+    function = form["functionselect"].value if "functionselect" in form else "all"
+    if function != "all":
+        cmd.extend(["--function", function])
```
simonlindholm, line 48: sanitize `function` to consist only of [a-zA-Z0-9_] similar to regvars

website.py
```diff
@@ -222,6 +245,24 @@
 updateDarkMode();
 darkModeCheckbox.addEventListener("change", updateDarkMode);
 
+function updateFunctions() {
+    var functionSelect = document.getElementsByName("functionselect")[0];
+    var prevValue = functionSelect.value;
+    var options = "<option value='all'>all functions</option>";
+    for (var line of sourceEl.value.split("\\n")) {
```
simonlindholm, line 252: "\n", surely? or oh, is this because of double quoting from Python? use an `r"""` string in that case.

Instead of splitting by newline you can do `for (let match of sourceEl.value.matchAll(/^\w*glabel\w+([A-Za-z0-9_]+)/mg))` I believe

simonlindholm (commented):
Sound like a nice feature! Running graphviz on the website should be totally fine, I'll go install that

website.py
```diff
@@ -9,15 +9,14 @@
...
 cgitb.enable()
 
-form = cgi.FieldStorage()
-do_visualize = "visualize" in form
-if do_visualize:
-    print(f"Content-Type: image/svg+xml; charset=utf-8")
-else:
-    print(f"Content-Type: text/html; charset=utf-8")
-print()
-sys.stdout.flush()
```
simonlindholm, line 15: The downside of delaying this is that if we hit an exception, cgitb will emit HTML before headers have been printed. Probably needs a try-catch

website.py
zbanks: I think `mips_to_c.py` is OK having garbage here? And we're not running it in a shell, so I don't think it's a security concern? 
(With regvars, it seems nice to strip `$` I guess?)

simonlindholm: Hm, yeah, maybe it's fine. There's a bit of an escaping hazard in that if the function name starts with `-` argparse will interpret it as an parameter, before giving up with an error about `--function` requiring an argument, and that sort of thing has made me cautious. Probably overly paranoid though, seems hard to imagine something actually dangerous coming from it :)

website.py
```diff
@@ -16,299 +16,303 @@ def print_headers(content_type: str) -> None:
...
+        if (!sidebarCheckboxEl.checked) return;
+        try {
+            document.getElementsByName("outputframe")[0].contentDocument.body.innerHTML = "Decompiling...";
+        } catch (e) {}
+    });
+    </script>
+    </form>
+    </body>
+    </html>
+    """
```
simonlindholm, line 313: dedent the HTML

website.py
```diff
@@ -16,299 +16,303 @@ def print_headers(content_type: str) -> None:
...
+        )
+except:
+    # Set the headers for the cgitb traceback
+    print_headers(content_type="text/html")
```
simonlindholm, line 317: should only do this if headers haven't been sent; alternatively, try-except only the subprocess parts

simonlindholm (approved):

PR #127, opened 2021-07-02
Make "emit globals" the default; change option to `--no-emit-globals`

zbanks:
- Now that the default changed, all the end-to-end tests changed.
  - Removed `--emit-globals` from all test flags that had it
  - I made 1 new test that has `--no-emit-globals`, just to make sure the flag works
- I left `--emit-globals` supported for compatibility... but tbh given how new it is, it's probably fine to just delete it entirely?
- Update the website

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -179,7 +179,7 @@ def create_project_tests(
         ):
             continue
 
-        flags = ["--emit-globals"]
+        flags = []
         if context_file is not None:
             flags.extend(["--context", str(context_file)])
 
--- a/src/main.py
+++ b/src/main.py
@@ -198,11 +198,6 @@ def parse_flags(flags: List[str]) -> Options:
         "unusual statements. Macro definitions are in `mips2c_macros.h`.",
     )
     group.add_argument(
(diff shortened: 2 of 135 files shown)


simonlindholm (approved):
I agree that removing `--emit-globals` would be totally fine. (In principle I'd have liked if all flags had no- versions of them, but in practice that seems annoying with argparse.)

We should really avoid emitting a declaration for the function being decompiled...

zbanks:
> We should really avoid emitting a declaration for the function being decompiled...

IMO I like when we emit the function prototypes when there are *multiple* functions being decompiled -- but yeah for a single function it's silly.

simonlindholm:
> IMO I like when we emit the function prototypes when there are multiple functions being decompiled -- but yeah for a single function it's silly.

Right, yeah. I think we should special-case this since I think it's the majority use-case.

Another thing I noticed is that some functions have signature is `(void)` for the definition, but `()` for the declaration, which is silly: I think we should show `()` as `(void)` for functions currently getting decompiled.

PR #128, opened 2021-07-02
Remove integer casts for Literals

zbanks:
(As discussed on Discord.)

Output from running tests against MM w/ context is [here](https://gist.github.com/zbanks/00b946953f767fa8c5a4af8a89abdf64).

One downside that we hadn't talked about yet is that all pointer casts used to be `(void *)`, regardless of pointer type. There are some places where we know it's a pointer, but we don't know the exact type, so it now shows up as `(? *)` or `(?32 *)`. 

I think this is still an improvement? I don't think `*(void *)0xA02FE1C0` is much better than `*(? *)0xA02FE1C0`: neither is valid C, and at least the latter is formatted as `*(MIPS2C_UNK32 *)...` with `--valid-syntax`.

### DIFF
--- a/tests/end_to_end/branch-likely-one-instruction/manual-out.c
+++ b/tests/end_to_end/branch-likely-one-instruction/manual-out.c
@@ -3,8 +3,8 @@ s32 test(); // static
 s32 test(void) {
     s32 phi_a0;
 
-    phi_a0 = *(void *)0x8009A600 == 3;
-    if (*(void *)0x800B0F15 == 5) {
+    phi_a0 = *(s32 *)0x8009A600 == 3;
+    if (*(s8 *)0x800B0F15 == 5) {
         phi_a0 = 1;
     }
     return phi_a0;
--- a/src/translate.py
+++ b/src/translate.py
@@ -1156,11 +1156,7 @@ class Literal(Expression):
         suffix = ""
         if not fmt.skip_casts:
             if self.type.is_pointer():
-                prefix = "(void *)"
-            elif self.type.get_size_bits() == 8:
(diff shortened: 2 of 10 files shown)


tests/end_to_end/branch-likely-one-instruction/manual-out.c
```diff
@@ -3,8 +3,8 @@ s32 test(); // static
 s32 test(void) {
     s32 phi_a0;
 
-    phi_a0 = *(void *)0x8009A600 == 3;
-    if (*(void *)0x800B0F15 == 5) {
+    phi_a0 = *(s32 *)0x8009A600 == 3;
```
simonlindholm, line 6: (huh, how does it know this is a s32 and not a u32?)

simonlindholm (approved):

tests/end_to_end/branch-likely-one-instruction/manual-out.c
zbanks: My understanding of [the code](https://github.com/matt-kempster/mips_to_c/blob/master/src/types.py#L312) is if it's either an `s32` or a `u32`, it calls it an `s32`. The `?32`s come when it's not sure if it's a `f32`, but here there's the `== 3`?

simonlindholm: Ah, right! And 2e08819b4ee6607be377845dca11d73fed1cdac1 (Allow `lw` to be used for any 32-bit type, not just s32) made store type annotations more `?`y, but loads haven't yet gone through that treatment

PR #129, opened 2021-07-03
Mark lw/sw values as 'likely not float'

simonlindholm:
This cleans up the ?'s introduced by 2e08819b4ee6607be377845dca11d73fed1cdac1.


simonlindholm:
Possible follow-up: https://github.com/simonlindholm/mips_to_c/compare/likely-ints...simonlindholm:likely-floats

zbanks (approved):
Looks good! This is a pretty short change!

I think it'd be a lot easier if we could do a pass over all `Type`s after translation and "squash" all of the `likely_kind`s into just `kind` -- I think having `.is_float()`, `is_likely_float()` etc. will start to get a bit complex? But I don't think the architecture is set up to make that easy.

src/types.py
```diff
@@ -311,7 +315,9 @@ def simple_ctype(typename: str) -> ca.TypeDecl:
         size = data.size or 32
         sign = "s" if data.sign & TypeData.SIGNED else "u"
 
-        if (data.kind & TypeData.K_ANYREG) == TypeData.K_ANYREG:
+        if (
+            data.kind & TypeData.K_ANYREG
+        ) == TypeData.K_ANYREG and data.likely_kind & TypeData.K_FLOAT:
```
zbanks, line 320: This was the one part that seems a bit too specific? But I see you change it up in `likely-floats`, so I think it's fine for now.

simonlindholm:
> I think it'd be a lot easier if we could do a pass over all Types after translation and "squash" all of the likely_kinds into just kind -- I think having .is_float(), is_likely_float() etc. will start to get a bit complex?

That's a very good point. Another one for if we ever start doing late type resolution!

PR #130, opened 2021-07-03
Mark lwc1/swc1 values as 'likely float'

simonlindholm:


src/translate.py
```diff
@@ -3781,7 +3781,7 @@ def read_pointer() -> Optional[Expression]:
 
         def for_element_type(type: Type) -> Optional[str]:
             """Return the initializer for a single element of type `type`"""
-            if type.is_int() or type.is_float():
+            if type.is_int() or type.is_likely_float():
```
zbanks, line 3784: Should this be `is_likely_int()`? 
Or even `is_numeric()`? I think we actually care that `kind` is a subset of `INT | FLOAT` -- not that it is likely one or the other? We can leave it up to `Literal.format()` to decide how to format the number?

zbanks (changes requested):

src/translate.py
simonlindholm: I think `kind == (K_INT | K_FLOAT)` won't actually ever happen in practice... we can only get there by starting from K_ANYREG or K_ANY, and anything that removes K_PTR will also remove K_FLOAT. Similarly is_likely_int() practically implies is_int(). Not sure what that implies the best option here is

simonlindholm: oh actually what should be done is to combine int and pointer handling; then we can use a single code path and leave things to Literal.format

simonlindholm:
Made some updates. This also fixes issues with Type.of_size() creating "denormalized" Type's, with e.g. size = 8 but the K_PTR and K_FLOAT bits set. Similarly `likely_kind` wouldn't actually be a subset of `kind` initially. (Any arbitrary successful `unify()` would normalize this.) I've confirmed that every place that constructs a TypeData now creates a normalized one. We may want to assert or normalize in the TypeData constructor, but it's not quite clear how to do that without duplicating code from unify()...

src/translate.py
```diff
@@ -3815,6 +3792,21 @@ def for_element_type(type: Type) -> Optional[str]:
...
 
+            if type.is_reg():
+                size_bits = type.get_size_bits()
+                if not size_bits:
+                    return None
+
+                if size_bits == 32:
+                    ptr = read_pointer()
+                    if ptr is not None:
+                        return as_type(ptr, type, silent=True).format(fmt)
```
zbanks, line 3803: This is much nicer than the "look before you leap" style I had before.

src/types.py
```diff
@@ -476,14 +491,19 @@ def int64() -> "Type":
         return Type(TypeData(kind=TypeData.K_INT, size=64))
 
     @staticmethod
-    def of_size(size: int) -> "Type":
-        return Type(TypeData(kind=TypeData.K_ANY, size=size))
+    def int_of_size(size: int) -> "Type":
+        return Type(TypeData(kind=TypeData.K_INT, size=size))
```
zbanks, line 495: Preventing this from being a `K_CTYPE` is more limiting than the old code (but I think that's OK).

The byte/short/word could be a union, a single-field struct, or a struct with bitfields. So I think this line is responsible for the following diff in OOT's `func_800DB2C0`:
```diff
@@ -14,17 +14,7 @@

     if (temp_v0 < (s32) gAudioContext.gAudioBufferParameters.updatesPerFrame) {
 loop_1:
         temp_v1 = &gAudioContext.gNoteSubsEu[(gAudioContext.gMaxSimultaneousNotes * phi_v0) + arg1];
-        if ((((s32) temp_v1->bitField0.s * 2) < 0) || (temp_v0_2 = phi_v0 + 1, temp_v1->bitField0.s = (struct 
-{
-  vu8 enabled : 1;
-  u8 needsInit : 1;
-  u8 finished : 1;
-  u8 envMixerNeedsInit : 1;
-  u8 stereoStrongRight : 1;
-  u8 stereoStrongLeft : 1;
-  u8 stereoHeadsetEffects : 1;
-  u8 usesHeadsetPanEffects : 1;
-}) ((u8) temp_v1->bitField0.s & 0xFF7F), phi_v0 = temp_v0_2, phi_return = temp_v0_2, phi_return = temp_v0_2, ((temp_v0_2 < (s32) gAudioContext.gAudioBufferParameters.updatesPerFrame) == 0))) {
+        if ((((s32) temp_v1->bitField0.s * 2) < 0) || (temp_v0_2 = phi_v0 + 1, temp_v1->bitField0.s = (u8) temp_v1->bitField0.s & 0xFF7F, phi_v0 = temp_v0_2, phi_return = temp_v0_2, phi_return = temp_v0_2, ((temp_v0_2 < (s32) gAudioContext.gAudioBufferParameters.updatesPerFrame) == 0))) {
 
         } else {
             goto loop_1;
```

*But*, I don't think this is actually a bad regression. I think your change is bringing this constructor to be in line with the rest of the constructors. In the future, we can add deliberate support for small CTypes.

zbanks (approved):
lgtm!

A side effect is that now m2c is confident enough to emit initializers for jump tables. This leads to output like this:
```diff
+extern ? L800DCCE4;
+extern ? L800DCCF8;
+extern ? L800DCD0C;
+extern ? L800DCD20;
+extern ? L800DCD60;
+extern ? L800DCDA0;
+static ? *jtbl_801493D0[6] = {&L800DCCE4, &L800DCD0C, &L800DCD20, &L800DCCF8, &L800DCDA0, &L800DCD60}; // const
```

Do you think we should exclude `jtbl_` & `jpt_` symbols from the global decls? (I don't think these decls are useful, unless IDO supports computed gotos)

src/types.py
simonlindholm: That's some pretty funny output from earlier. Yeah, I agree that we should make this as an intentional change later on. It's similar to how we don't support stack structs right now.

simonlindholm:
> Do you think we should exclude jtbl_ & jpt_ symbols from the global decls? (I don't think these decls are useful, unless IDO supports computed gotos)

Definitely! IDO indeed does not support computed gotos.

PR #131, opened 2021-07-10
Improve return detection

### LINKED ISSUE #11: Improve return detection
- Mark returns as definite if `v0`/`f0` is manually set and not read from, ~~and definitely `void` if both are unset at some return point.~~
- ~~Call `mark_used` on return values (on the return node's return register, really, which might be a phi expr!). Maybe move the logic for handling return values into translate.py, for that purpose.~~

simonlindholm:
Instead of the fake "return" return set whenever v0 or f0 is set, check for v0 and f0 individually which makes the most sense to return.

For deciding which one makes the most sense, we look at whether one of the registers is written without being read. (If both are we default to v0.) Otherwise, we assume the function is void, unless we have a known C type or the new flag --not-void is passed. In that case we pick whichever one has a non-function-return value written to it, with preference to v0. (Later on, a known C type should probably imply which one of the registers to use.)

There's a bunch of new, unfortunate machinery for tracking register writes/reads, that kinda duplicates the existing phi node logic. Guessing this is unavoidable.

Still need to write more tests for this and check how it affects real-world cases, and extend the website with --non-void.

Fixes #11.


tests/end_to_end/sqrt/irix-o2-out.c
```diff
@@ -1,13 +1,10 @@
...
-    f32 temp_f0;
 
     sp4 = arg0;
     spC = sp4;
     spC = (bitwise f32) (0x5F3759DF - ((bitwise s32) sp4 >> 1));
     sp4 = spC;
-    temp_f0 = (1.5f - (arg0 * 0.5f * spC * spC)) * spC;
-    sp4 = temp_f0;
-    return temp_f0;
+    sp4 = (1.5f - (arg0 * 0.5f * spC * spC)) * spC;
```
simonlindholm, line 9: Here's an example of a case where it incorrectly thinks the function is void: IDO emits an unnecessary `swc1 $f0, 4($sp)`, leading m2c to think that `$f0` isn't a return register

src/translate.py
```diff
@@ -1495,20 +1521,25 @@ def __contains__(self, key: Register) -> bool:
 
     def __setitem__(self, key: Register, value: Expression) -> None:
         assert key != Register("zero")
-        self.contents[key] = value
+        self.contents[key] = RegData(value, RegDataMeta())
+
+    def set_raw(self, key: Register, data: RegData) -> None:
+        assert key != Register("zero")
+        self.contents[key] = data
```
zbanks, line 1528: This feels asymmetrical? This is the only function which takes (or returns) a `RegData`, it's not the "complement" to `get_raw(...) -> Optional[Expression]`?

I think I'd either rename this to something that's distinct from `get_raw()`, like `set_regdata()`? Or refactor `get_raw()` to return an `Optional[RegData]` (but that sounds harder).

src/translate.py
```diff
@@ -3397,12 +3509,17 @@ def process_instr(instr: Instruction) -> None:
...
                 fn_target = args.imm(0)
                 if isinstance(fn_target, AddressOf) and isinstance(
                     fn_target.expr, GlobalSymbol
                 ):
-                    pass
+                    typemap = stack_info.global_info.typemap
+                    if typemap:
+                        c_fn = typemap.functions.get(fn_target.expr.symbol_name)
+                        if c_fn and c_fn.ret_type is None:
+                            is_known_void = True
```
zbanks, line 3522: Would it be better to go off of `GlobalInfo.address_of_gsym(...)` here? 
The additional type info would not always be available, but if you were decomping multiple functions at once you could incorporate the inferred type signature from earlier passes?

src/translate.py
```diff
@@ -3689,17 +3805,19 @@ def translate_graph_from_block(
     for child in node.immediately_dominates:
         if isinstance(child, TerminalNode):
             continue
-        new_contents = regs.contents.copy()
+        new_contents = {
+            reg: RegData(data.value, RegDataMeta(inherited=True))
```
zbanks, line 3809: Do we need to throw away the previous values in `RegDataMeta` here? It looks like `function_return=True` could be set.

src/translate.py
zbanks: There is also a lot of code which accesses `RegInfo.contents` directly, like `info.contents.get(reg)`. 
Do you think this is worth encapsulating?

zbanks (changes requested):

src/translate.py
simonlindholm: This is intentional, see the updated comment near the function call handling. We used to get some ERROR(read from unset register v0)'s before due to this being out of sync with output_regs_for_instr (which is used to construct phi nodes).

src/translate.py
simonlindholm: It doesn't matter either way -- for `inherited=True` metas we recompute `function_return` based on parent nodes.

(fwiw I had an earlier patch set which reused the entire old RegData here, in the case where it doesn't create a phi for the register, but it ended up less uniform and harder to understand: it made phi vs non-phi parents into separate cases for no real reason.)

src/translate.py
simonlindholm: Added `get_meta` and `set_with_meta` accessors and removed `set_raw`. There's still some two places that access `contents` directly but it feels better now at least.

PR #132, opened 2021-07-11
Jump table switch statements

zbanks:
- Detect `goto **(&jtbl + (value + (-offset)) * 4);` statements, to make `switch (value)`
    - `offset` is optional, but if it exists, then incorporate it into the case labels.
- Emit case blocks in asm order, rather than jump table order
- Look for a parent `ConditionalNode` to determine `default:` node
    - It'd be nice to combine the `ConditionalNode` with the `SwitchNode` - but I think that's a trickier to get right. (One issue is when `SwitchNode` has `SetPhiStmt`s.) 
- Use `break;` instead of `goto end;`, when appropriate
- Detect some fallthrough spots, mark them with a comment

I checked that this merges cleanly with the other open PRs.


src/if_statements.py
```diff
@@ -600,27 +648,56 @@ def emit_return(context: Context, node: ReturnNode, body: Body) -> None:
 
 def build_switch_between(
     context: Context,
-    start: SwitchNode,
+    switch: SwitchNode,
     end: Node,
```
simonlindholm, line 652: Pre-existing and more of a question than a review comment, but I figure I should ask anyway: when the switch contains an early return, will that result in `end` being a `TerminalNode`? Won't that throw a lot of control flow construction off?

Regardless of answer I think cases with early returns would probably make for good test cases!

src/if_statements.py
```diff
@@ -600,27 +648,56 @@ def emit_return(context: Context, node: ReturnNode, body: Body) -> None:
...
-    return SwitchStatement(jump, body)
+    jump: Optional[Union[Expression, SimpleStatement]]
+    jump = get_block_info(switch).switch_value
+    if not isinstance(jump, SwitchControl):
+        jump = switch_jump(context, switch)
+
+    switch_body = Body(print_node_comment=context.options.debug)
+
+    # Order case blocks by their position in the asm, not by their order in the jump table
+    sorted_cases = sorted(set(switch.cases), key=lambda node: node.block.index)
```
simonlindholm, line 680: are all the `.index`es unique, or can this give non-deterministic results? (oh, why does python make map iteration order deterministic but not set? :()

src/if_statements.py
```diff
@@ -600,27 +648,56 @@ def emit_return(context: Context, node: ReturnNode, body: Body) -> None:
...
+    next_sorted_cases: List[Optional[Node]] = []
+    next_sorted_cases.extend(sorted_cases[1:])
+    next_sorted_cases.append(None)
+    for case, next_case in zip(sorted_cases, next_sorted_cases):
+        if case in context.emitted_nodes or case is end:
+            pass
+        elif (
+            next_case is not None
+            and next_case not in context.emitted_nodes
```
simonlindholm, line 689: why is this check here? If I understand correctly it's harmless, because the `else` codepath is also correct, just with missing `// fallthrough` comments, but it feels off to me

simonlindholm (commented):
> Emit case blocks in asm order, rather than jump table order

Oh, I completely missed that regression when it happened. :( It used to be that the `emit_goto` machinery handled this: we'd emit the whole contents of the switch, and `emit_goto=True` would make it proceed through `break`s. A big advantage of that method is that it would deal naturally even with switches split up between multiple jump tables (with intertwined cases) and (in theory, with improved `emit_goto` marking) initial `beq`'s that are part of the switch. I guess in the new world we'll just have to work towards merging switch jumps that correspond to the same switch? Or could there be value in resurrecting `emit_goto`...?

src/if_statements.py
```diff
@@ -320,15 +338,45 @@ def emit_goto(context: Context, target: Node, body: Body) -> None:
     body.add_statement(SimpleStatement(f"goto {label};", is_jump=True))
 
 
+def add_labels_for_switch(
+    context: Context, node: SwitchNode, default_node: Optional[Node]
+) -> None:
+    assert node.cases, "jtbl list must not be empty"
+    if node not in context.switch_nodes:
```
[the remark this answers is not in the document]
zbanks: Yeah I can return early instead. 
Now `context.switch_nodes` starts empty, and this function is responsible for building it.

src/if_statements.py
zbanks: If `next_case` is already emitted, then nothing will be emitted in the next iteration (to avoid duplicating the node). This is an unusual situation - this means that `next_node` has a parent outside of the `SwitchNode` tree.

Commenting out this check results in this diff in `multi-switch`. Here, node 16 can be reached from either switch 1 or 2 (and was originally emitted in switch 1). The "fallthrough" comment isn't correct here.
```diff
--- 

+++ 

@@ -74,8 +74,7 @@

                     return arg0 * arg0;
                 case 2: // switch 2
                     phi_a0_2 = arg0 - 1;
-                    // Duplicate return node #16. Try simplifying control flow for better match
-                    return (phi_a0_2 + 1) ^ phi_a0_2;
+                    // fallthrough
                 case 6: // switch 2
                 case 7: // switch 2
                     phi_a0_3 = arg0 * 2;
```

src/if_statements.py
zbanks: Yeah, `end` would be the `TerminalNode`. The cases will have to end in `goto` instead of `break;`

This also sort of happens with `ConditionalNode`s as well, but when the if body ends in a jump, the else is elided.

src/translate.py
```diff
@@ -1355,6 +1355,64 @@ def format(self, fmt: Formatter) -> str:
         return self.get_var_name()
 
 
+@dataclass
+class SwitchControl(Expression):
```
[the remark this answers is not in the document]
zbanks: We still need to `.use()` the `control_expr` - so I think it's OK to keep it an Expression?
I'm going to do the `Optional[SwitchControl]` thing though.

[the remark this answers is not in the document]
simonlindholm: The `use()` is at the same place as the SwitchControl is constructed though, so we could just `use()` the underlying switch_value there instead

simonlindholm:
> I guess in the new world we'll just have to work towards merging switch jumps that correspond to the same switch? Or could there be value in resurrecting emit_goto...?

Hm, I suppose an idea here could be: generate case labels early, then when emitting the contents of a switch, loop over all blocks lexically afterwards, and do something like:
```python
while node is not end and end in node.post_dominators:
    if node has case label
        if previous was not break:
            emit "// fallthrough"
        emit that case label
    build_flowgraph_between(node, node.immediate_postdominator)
    if node.immediate_postdominator is end:
        emit "break"
        node = next one lexically
    else:
        node = node.immediate_postdominator
```
I haven't thought this through very deeply though, it probably gets hairy trying to deal with already emitted nodes, and having multiple switch nodes that both try to do this...

zbanks:
Need to play with the merge, not yet ready for review. (Wish I could un-hit that button haha)

zbanks:
> > I guess in the new world we'll just have to work towards merging switch jumps that correspond to the same switch? Or could there be value in resurrecting emit_goto...?
> 
> Hm, I suppose an idea here could be: generate case labels early, then when emitting the contents of a switch, loop over all blocks lexically afterwards, and do something like:
> ...
> I haven't thought this through very deeply though, it probably gets hairy trying to deal with already emitted nodes, and having multiple switch nodes that both try to do this...

I guess I'm holding out for the "merging switch jumps that correspond to the same switch" part. (For now, I think its big blocker is the phi machinery, which inserts statements between the jumps in a way that's hard to translate to simple C.)

I think the issue with situations like `multi-switch` is that the nodes in the flow graph end up in a structure that's very difficult to emit in lexical order. In `multi-switch`, the order of nodes `14, 16, 20, 21` bounce back-and-forth between switches 1 & 2 in a way that you'd need to start from a dominating node (like node 0) instead of the SwitchNode to emit correctly, I think?

src/if_statements.py
```diff
@@ -320,14 +336,49 @@ def emit_goto(context: Context, target: Node, body: Body) -> None:
...
+        # There is only one switch in this function (no need to label)
+        context.switch_nodes[node] = 0
+    else:
+        context.switch_nodes[node] = len(context.switch_nodes) + 1
+    switch_index = context.switch_nodes[node]
+
+    # Determine offset
+    offset = 0
+    switch_control = get_block_info(node).switch_control
+    if isinstance(switch_control, SwitchControl):
```
simonlindholm, line 357: `assert switch_control is not None`

src/translate.py
```diff
@@ -1354,6 +1355,71 @@ def format(self, fmt: Formatter) -> str:
...
+    jump_table: Optional[GlobalSymbol] = None
+    offset: int = 0
+
+    def format(self, fmt: Formatter) -> str:
+        index = format_expr(self.control_expr, fmt)
+        if not self.jump_table:
+            return index
+        if self.offset:
+            index = f"{index} - {self.offset}"
+        return f"(&{format_expr(self.jump_table, fmt)})[{index}]"
```
simonlindholm, line 1370: might be misleading to include the offset in the output here, since it's already been added to each case

I think it's fair to give up on compilability in the --gotos-only case and print something like `switch (expr); // switch N`; then we can remove this format method altogether. (maybe with the jump table in a comment if you think it's useful)

simonlindholm (approved):

src/translate.py
zbanks: Yeah, I'll give up compatibility with `--gotos-only` then. :+1:

PR #135, opened 2021-07-18
Improve phi assignment by picking an earlier/smaller set of nodes for SetPhiStmt

zbanks:
In `assign_phis()`, rather than emit the `SetPhiStmt` at the end of each of parents, try to find a different set of nodes that has equivalent behavior, but will likely produce nicer code.

- Prefer picking *earlier* nodes when possible. This helps avoid getting assignments inside complex conditionals (see `andor` tests), and can result in similar phi nodes "bunching up" in a way that's easier to refactor (see `phi_a0_2`-`5` in `multi-switch` test).
- If two or more of `phi.node.parents` share a dominator, we can do the assignment *once* in the dominator, as long as the value of the register at that node is the same. This reduces the number of statements.
- I originally thought loops would be tricky. I was worried we needed to avoid moving phi assignments out of loops, because that's not equivalent. I found some places where the phis got moved out of loops, but it turned out that moving the phi out made it *more* correct. ([`Mio0_Decompress`, in `mio0.c`](https://gist.github.com/zbanks/d628c6a4bfb3c9bfec8d4246ba6ac4bb)) I couldn't manage to reproduce this behavior in an end-to-end test.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -3160,6 +3160,35 @@ def reg_always_set(
     return True
 
 
+def pick_phi_assignment_nodes(
+    reg: Register, nodes: List[Node], uw_expr: Expression
+) -> List[Node]:
+    """
+    As part of `assign_phis()`, we need to pick a set of nodes where we can emit a
+    `SetPhiStmt` that assigns the phi for `reg` to `uw_expr`.
+    The final register state for `reg` for each node in `nodes` unwraps to `uw_expr`,
+    so the best case would be finding a single dominating node for the assignment.
+    """
+    # Find the set of nodes which dominate *all* of `nodes`, sorted by number
+    # of dominators. (This puts "earlier" nodes at the beginning of the list.)
+    dominators = sorted(
+        set.intersection(*(node.dominators for node in nodes)),
+        key=lambda n: len(n.dominators),
+    )
+
+    # Check the dominators for a node with the correct final state for `reg`
+    for node in dominators:
+        raw = get_block_info(node).final_register_states.get_raw(reg)
+        if raw is None:
+            continue
+        if early_unwrap(raw) == uw_expr:
+            return [node]
+
+    # We couldn't find anything, so fall back to the naive solution
+    # TODO: In some cases there may be a better solution (e.g. one that requires 2 nodes)
+    return nodes
+
+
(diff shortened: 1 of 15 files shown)


src/translate.py
```diff
@@ -3168,11 +3197,17 @@ def assign_phis(used_phis: List[PhiExpr], stack_info: StackInfo) -> None:
         phi = used_phis[i]
         assert phi.num_usages > 0
         assert len(phi.node.parents) >= 2
-        exprs = []
+
+        # Group parent nodes by the value of their phi register
+        equivalent_nodes: Dict[Expression, List[Node]] = {}
```
simonlindholm, line 3202: `= defaultdict(list)` if you want to avoid the `if expr not in equivalent_nodes` check

tests/end_to_end/andor_assignment/irix-o2-out.c
```diff
@@ -62,9 +62,7 @@ s32 test(s32 arg0, s32 arg1, s32 arg2, s32 arg3) {
             phi_v1_3 = phi_v1_6 + 5;
         }
     }
-    phi_v1_4 = phi_v1_3;
-    phi_v1_7 = phi_v1_3;
-    if ((phi_s0_2 != 0) && (phi_t0_2 != 0) && (sp24 = phi_v1_3, (func_00400090(phi_s0_2 + phi_t0_2) != 0)) && (arg3 != 0)) {
+    if ((phi_s0_2 != 0) && (phi_t0_2 != 0) && (sp24 = phi_v1_3, phi_v1_4 = phi_v1_3, phi_v1_7 = phi_v1_3, (func_00400090(phi_s0_2 + phi_t0_2) != 0)) && (arg3 != 0)) {
```
simonlindholm, line 65: do you know why this change happened? I don't understand it

simonlindholm (approved):

tests/end_to_end/andor_assignment/irix-o2-out.c
zbanks: The function call `func_00400090` saves/restores the register `v1`. [When the register is restored, it is wrapped in an `EvalOnceExpr`](https://github.com/matt-kempster/mips_to_c/blob/master/src/translate.py#L3421-L3435) which is  no longer unwrapped.

PR #136, opened 2021-07-19
Detect and combine "if guards" with switches

zbanks:
Most switches are proceeded by a `ConditionalNode` which performs a bounds check to ensure that the jump table access is valid. This check is pretty consistent across IDO/GCC. 

It's straightforward to combine this check with the `switch (...)` statement itself: if the conditional branch indicates the default case. (If the conditional branch is to `end`, that indicates that there is *no* default case).

This change was enabled by the earlier phi PR. There used to frequently be phi assignment expressions at the top of `SwitchNode`s, which makes it hard to combine the if+switch. Now those are almost entirely gone, so this heuristic is drastically more successful.

---

I tested this across OOT, MM, & PM (without context). It was able successfully combine 1045 instances, and only failed at 13. (Note: there ~180 functions which are non-reducible: these fall back to gotos-only mode, so they aren't tested here.) The issues were:
- 5 instances where the condition couldn't be parsed (e.g. the condition was partially turned into a temp)
- 4 instances where there were phi statements in the SwitchNode
- 3 instances like `switch (x & 7)`, which doesn't need a guard expression (cool!)
- 1 instance where the SwitchNode control expression wasn't parsed at all, and already had a "cannot parse jump table" comment

### DIFF
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -670,20 +670,23 @@ def build_switch_between(
     Output the subgraph between `switch` and `end`, but not including `end`.
     The returned SwitchStatement starts with the jump to the switch's value.
     """
+    switch_cases = switch.cases[:]
+    if default is end:
+        default = None
+    elif default is not None:
+        switch_cases.append(default)
+
     switch_index = add_labels_for_switch(context, switch, default)
 
     jump = get_block_info(switch).switch_control
     assert jump is not None
 
     switch_body = Body(print_node_comment=context.options.debug)
-    cases = switch.cases[:]
-    if default is not None and default is not end:
-        cases.append(default)
 
     # Order case blocks by their position in the asm, not by their order in the jump table
     # (but use the order in the jump table to break ties)
     sorted_cases = sorted(
-        set(cases), key=lambda node: (node.block.index, cases.index(node))
+        set(switch_cases), key=lambda node: (node.block.index, switch_cases.index(node))
     )
     next_sorted_cases: List[Optional[Node]] = []
     next_sorted_cases.extend(sorted_cases[1:])
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -381,6 +381,28 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
         emit_goto(context, target, body)
 
 
+def is_switch_guard(node: Node) -> bool:
+    """Return True if `node` is a ConditionalNode for checking the bounds of a
+    SwitchNode's control expression. These can usually be combined in the output."""
+    if not isinstance(node, ConditionalNode):
+        return False
+    cond = get_block_info(node).branch_condition
+    assert cond is not None
+
+    switch_node = node.fallthrough_edge
+    if not isinstance(switch_node, SwitchNode):
+        return False
+    switch_block_info = get_block_info(switch_node)
+    assert switch_block_info.switch_control is not None
+
+    # The SwitchNode must have no statements, and the conditional
+    # from the ConditionalNode must properly check the jump table bounds.
+    return (
+        not switch_block_info.statements_to_write()
+        and switch_block_info.switch_control.matches_guard_condition(cond)
+    )
+
+
 def gather_any_comma_conditions(block_info: BlockInfo) -> Condition:
     branch_condition = block_info.branch_condition
     assert branch_condition is not None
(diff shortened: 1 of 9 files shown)


tests/end_to_end/switch-with-if/irix-o2.s
```diff
@@ -4,10 +4,19 @@
...
+.word .case2
 .word .case2
 
 glabel jtbl_4001B8
 .word .case3
+.word .case3
+.word .case3
+.word .case3
+.word .case4
+.word .case4
```
zbanks, line 19: I updated this test to accurately reflect the `orig.c`

tests/end_to_end/multi-switch/irix-o2-out.c
```diff
@@ -19,16 +19,12 @@ s32 test(s32 arg0) {
...
-        if ((u32) (arg0 - 0x65) < 7U) {
-            switch (arg0) { // switch 1
-            case 107: // switch 1
-                phi_a0 = arg0 + 1;
-                // Duplicate return node #24. Try simplifying control flow for better match
-                D_410210 = phi_a0;
-                return 2;
-            }
-        } else {
-            goto block_23;
```
zbanks, line 31: This looks bad ("where did `goto block_23` go? shouldn't there be a `default:` somewhere?) -- but the `default: // switch 1` is down at line 70, unchanged.

src/translate.py
```diff
@@ -1418,7 +1454,7 @@ def from_expr(expr: Expression) -> "SwitchControl":
                 offset = -offset_lit.value
 
         # Check that it is really a jump table
-        if not symbol_name_is_jump_table(jump_table.symbol_name):
+        if jump_table.asm_data_entry is None or not jump_table.asm_data_entry.is_jtbl:
```
zbanks, line 1457: I just added this fn, but I think it's better to just use the `is_jtbl` attribute that's already there! :facepalm:

src/if_statements.py
```diff
@@ -788,10 +809,25 @@ def build_flowgraph_between(
...
+            default_node = curr_start.conditional_edge
+            # is_switch_guard checked that switch_node has no statements to write,
+            # so it is OK to mark it as emitted
+            context.emitted_nodes.add(switch_node)
+            body.add_switch(
+                build_switch_between(context, switch_node, default_node, curr_end)
+            )
+        elif isinstance(curr_start, ConditionalNode):
             body.add_if_else(build_conditional_subgraph(context, curr_start, curr_end))
         elif isinstance(curr_start, SwitchNode):
```
simonlindholm, line 829: nit: move this case up before the ConditionalNode one

src/if_statements.py
```diff
@@ -381,6 +381,28 @@ def emit_goto_or_early_return(context: Context, target: Node, body: Body) -> Non
...
+        return False
+    switch_block_info = get_block_info(switch_node)
+    assert switch_block_info.switch_control is not None
+
+    # The SwitchNode must have no statements, and the conditional
+    # from the ConditionalNode must properly check the jump table bounds.
+    return (
+        not switch_block_info.statements_to_write()
+        and switch_block_info.switch_control.matches_guard_condition(cond)
+    )
```
simonlindholm, line 403: can you check that `switch_node`'s only parent is `node`, too?

src/translate.py
simonlindholm: heh, fair enough. Should probably inline the function at its only usage site again, so it's closer in sync with the error message

simonlindholm (approved):

src/translate.py
```diff
@@ -1373,6 +1372,43 @@ class SwitchControl:
...
+        elif left_expr != late_unwrap(self.control_expr):
+            return False
+
+        right_expr = late_unwrap(cmp_expr.right)
+        return (
+            self.jump_table is not None
+            and self.jump_table.asm_data_entry is not None
+            and self.jump_table.asm_data_entry.is_jtbl
+            and isinstance(right_expr, Literal)
+            # Allow inexact matches, as long as it is less than the detected jump table length
```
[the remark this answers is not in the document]
zbanks: I think this was from padding, which is easy to strip out. I have a stricter version that I'm re-testing now.

zbanks:
Changes in `d097e94` had no diffs in OOT/MM/PM :heavy_check_mark:

PR #138, opened 2021-07-21
Add fake edges to the flow graph to break infinite loops and make the graph reducible

zbanks:
This is sort of a hack? I definitely want feedback on this, maybe it's *too* hacky?

The idea is to find nodes which cannot reach the TerminalNode (i.e. infinite loops) and add a "fake" edge to the TerminalNode. This edge is used by `compute_relations(...)`, but then is ignored by `build_flowgraph_between(...)`. 

This change fixed 190 functions across OOT/MM/PM (183 of them from PM). 
There are only 4 functions left in PM, plus 2 end-to-end tests, which still have the "Flowgraph is not reducible..." warning. These functions cannot be made reducible by adding new edges (they would need to split nodes).

### DIFF
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -832,6 +832,9 @@ def build_flowgraph_between(
             body.add_switch(build_switch_between(context, curr_start, None, curr_end))
         elif isinstance(curr_start, ConditionalNode):
             body.add_if_else(build_conditional_subgraph(context, curr_start, curr_end))
+        elif isinstance(curr_start, BasicNode) and curr_start.fake_successor:
+            assert curr_start.fake_successor is curr_end
+            curr_end = curr_start.successor
         else:
             # No branch, but double check that we didn't skip any nodes.
             # If the check fails, then the immediate_postdominator computation was wrong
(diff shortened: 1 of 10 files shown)


src/if_statements.py
```diff
@@ -930,7 +933,6 @@ def build_body(context: Context, options: Options) -> Body:
         if options.ifs and not is_reducible:
             body.add_comment(
                 "Flowgraph is not reducible, falling back to gotos-only mode. "
```
simonlindholm, line 935:
```suggestion
                "Flowgraph is not reducible, falling back to gotos-only mode."
```

src/if_statements.py
```diff
@@ -832,6 +832,9 @@ def build_flowgraph_between(
             body.add_switch(build_switch_between(context, curr_start, None, curr_end))
         elif isinstance(curr_start, ConditionalNode):
             body.add_if_else(build_conditional_subgraph(context, curr_start, curr_end))
+        elif isinstance(curr_start, BasicNode) and curr_start.fake_successor:
+            assert curr_start.fake_successor is curr_end
```
simonlindholm, line 836:
```suggestion
        elif isinstance(curr_start, BasicNode) and curr_end == curr_start.fake_successor:
```
maybe?

simonlindholm (approved):
Seems reasonable to me. Infinite loops are kinda similar to early returns.

PR #139, opened 2021-07-23
Add an option to emit underscores for unknown variables

Kelebek1:


src/main.py
```diff
@@ -229,6 +229,13 @@ def parse_flags(flags: List[str]) -> Options:
         choices=["left", "right"],
         default="right",
     )
+    group.add_argument(
+        "--unk-underscore",
+        dest="unknown_underscore",
+        help="control whether to output an underscore on unknown variables. I.e unk2AA vs unk_2AA",
```
simonlindholm, line 235:
```suggestion
        help="emit unk_X instead of unkX for unknown struct accesses",
```

src/main.py
```diff
@@ -229,6 +229,13 @@ def parse_flags(flags: List[str]) -> Options:
...
+        action="store_true",
+        default=False,
```
simonlindholm, line 237:
```suggestion
```
(it's the default when you have `action="store_true"`)

simonlindholm (approved):

PR #140, opened 2021-07-23
Rename mips2c_macros to make tab completion easier

Kelebek1:
This change is just to make tabbing to the main mips_to_c scripts easier. Currently hitting m, mi, mip, or mips will have tab completion stuck due to the macros file beginning with the same. Needing to press tab and then an extra _ is kinda awkward. I assume there's some reason that the file can't just be named "macros.h", so I've made it m2c_macros instead. This at least allows typing "mi" to get tab completion for the main script.


Kelebek1:
Closing as no one's really in favour. `./` can limit to executables anyways.

PR #142, opened 2021-08-10
Add an option to output switch case labels in hex

Kelebek1:
Adds `--hex-case` to options. This is sometimes needed in decomp PRs, where some numbers make more sense in hex.

### DIFF
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -154,7 +154,10 @@ class LabelStatement:
         lines = []
         if self.node in self.context.case_nodes:
             for (switch, case) in self.context.case_nodes[self.node]:
-                case_str = f"case {case}" if case is not None else "default"
+                case_num = None
+                if case is not None:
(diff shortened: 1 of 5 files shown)


src/if_statements.py
```diff
@@ -154,7 +154,10 @@ def format(self, fmt: Formatter) -> str:
         lines = []
         if self.node in self.context.case_nodes:
             for (switch, case) in self.context.case_nodes[self.node]:
-                case_str = f"case {case}" if case is not None else "default"
+                case_num = None
+                if case is not None:
+                    case_num = f"0x{case:X}" if fmt.coding_style.hex_case else f"{case}"
+                case_str = f"case {case_num}" if case is not None else "default"
```
simonlindholm, line 160:
```suggestion
                if case is not None:
                    case_num = f"0x{case:X}" if fmt.coding_style.hex_case else f"{case}"
                    case_str = f"case {case_num}" 
                else:
                    case_str = "default"
```

simonlindholm (approved):

PR #143, opened 2021-09-01
Check `phi.replacement_expr` in `PhiExpr.propagates_to()`

zbanks:
This seems to fix the issue mzx was having in Discord with function `func_800CA22C` from MM, where `phi_a0_2` was uninitialized.

It looks like there was a similar bug in the e2e test for `andor_assignment` - so there's already test coverage for this fix.

I don't have a good enough understanding of the phi machinery to succinctly explain why this fixes the problem: but, I don't think this change invalidates any existing comments. If you have any suggestions for updating them to be more clear/explicit let me know. 
This problem predates my earlier PR #135 that changed where phi assignments were handled.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -1348,7 +1348,7 @@ class PhiExpr(Expression):
         admittedly a bit sketchy, in case the phi is in scope here and used
         later on... but we have that problem with regular phi assignments as
         well."""
-        if self.used_by is None:
+        if self.used_by is None or self.replacement_expr is not None:
             return self
         return self.used_by.propagates_to()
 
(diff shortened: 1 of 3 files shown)


simonlindholm:
Without this check we're both saying:

- if `x` and `y` are phis, the only usage of `y` is in `x = y`, and `y` is set from either `a` or `b`, emit `x = a`/`x = b` instead of `y = a`/`y = b`, and elide the `x = y` store
- if `a` and `b` represent the same value, don't emit any `y = a`/`y = b` sets, and expand `y` arbitrarily to `a`

These conflict: we're removing both the `x = y` store and the forwarded `x = a` store. This PR disables forwarding when doing same-value-based expansion, which seems right to me.

Btw, the second affected test shows that we could do same-value-expansion on `x` after we've done so for `y` to improve the output further, and that would eliminate the need for this PR for that test. But it's still needed in general (e.g. for the first affected test).

simonlindholm (approved):

PR #144, opened 2021-09-24
Remove K_CTYPES; represent structs/arrays in Type

zbanks:
This is a big PR :disappointed: -- but I think it's worth it?

I originally tried to eliminate diffs in the test output, but ended up with the following changes. For the most part, they improve the output (and a few outstanding bugs get fixed!)

- `StructAccess.late_field_name()` handles cases where `offset=0` slightly better, which lets `elide_casts_for_store()` decide to remove some extraneous casts.
    - This is responsible for almost all of the changes in the e2e tests. If you view the diff and exclude commit `a9d0b8b7`, it can be easier to spot the other changes.
- The output of `dump_typemap()` is shorter, because `c_type.Struct`s no longer contain all of their "deep" fields.
- Many big struct accesses (> `0x10000` bytes) are fixed.
    - Example in OOT's `BossGanon_Update`: `arg1->unk10B09` is now `arg1->envCtx.unk_E2[3]`
- The new `get_field()` is stricter about having the `target_size` match when provided. 
    - I ran into a diff in `Sram_VerifyAndLoadAllSaves` where I was getting `*phi_s0` instead of `phi_s0->entranceIndex` because the load was expecting an `s16`. This diff goes away if I change the `SaveContext` struct in the context and replace `s32 entranceIndex;` with `s16 entranceIndex; s16 _padding;`.
    - I'm encountering similar diffs in the audio code with the first fields in `struct SequencePlayer` and `struct SequenceChannel`. It can't recognize reading the first byte of these structs, unless the unions are updated to include a byte-wide type at offset 0.

I've primarily been testing using the OOT corpus. [Here's the diff from `master`](https://gist.github.com/zbanks/cbdd1c0f93efa7129227c1613b7db8fa).

The end-to-end tests run at about the same speed, but running against OOT with context seems to be faster. Running with `--filter=Ganon2` took 11.5s on `master`, but 8s on this branch. (The intent of this PR isn't to make performance improvements, but I wanted to check for a regression.)


src/types.py
```diff
@@ -1,39 +1,99 @@
...
-    var_size_align,
 )
 from .error import DecompFailure
 from .options import Formatter
 
+AccessPath = List[Union[str, int]]
+
+
+@dataclass(eq=False)
+class TypeUniverse:
```
zbanks, line 26: This is intended to be a *mutable* set of `Type`s and contains information about derived types. This also completely abstracts `TypeMap` away; `translate.py` no longer references `typemap` at all. 

It keeps a reference to the `TypeMap`, which solely represents the information in the context file.
For example, when running tests, each test runs with a fresh `TypeUniverse` object, but they share `TypeMap`s. 

I don't love the name though :smile: -- not sure if anyone has any better ideas. One option is to rename `c_types.TypeMap` to something like `ContextTypeMap`/`CTypeMap`/`Context`, so this class could take the `TypeMap` name.

src/types.py
```diff
@@ -599,164 +823,96 @@ def unify_with_args(self, concrete: "FunctionSignature") -> bool:
...
+@dataclass(eq=False)
+class StructDeclaration:
+    # TODO: Document
+    size_bits: int
+    align_bits: int
+    tag_name: Optional[str]
+    typedef_name: Optional[str]
+    fields: List[StructField]
+    has_bitfields: bool
+    is_union: bool
```
zbanks, line 842: I decided to measure everything here in bits instead of bytes, even though there isn't really support for bitfields yet. I figured this would make it easier to add support for bitfields in the future? Although it's confusing, I tried to compensate by being as explicit as possible in the names (e.g. `target_size_bits`).

Is this a good call? Or is it too confusing for something we don't really use?

src/types.py
```diff
@@ -239,67 +311,139 @@ def get_function_pointer_signature(self) -> Optional["FunctionSignature"]:
...
+        """If self is a struct, return the StructDeclaration"""
+        if self.is_struct():
+            data = self.data()
+            assert data.struct is not None
+            return data.struct
+        return None
+
+    _GetFieldResult = Tuple[AccessPath, "Type", Optional[int]]
+
+    def get_field(
```
zbanks, line 324: Some high-level comments about the new `get_field`:
- Replaces both the old `get_field()` & `find_substruct_array()`
- Recursively traverses struct fields/arrays, rather than using a precomputed set of fields.
- The return type is kind of awkward, still considering how best to refactor that.

src/translate.py
```diff
@@ -1012,37 +1009,61 @@ class StructAccess(Expression):
...
+
         if (
             self.field_name is None
-            and self.stack_info.global_info.typemap
             and not self.has_late_field_name
+            # TODO: The previous behavior only tried to resolve the late field name
+            # if a typemap was available; however, even without a typemap, we can
+            # resolve zero-offset cases like `*(int *) x`.
+            # The following line is a hack to preserve this old behavior.
+            and self.stack_info.global_info.universe.typemap.var_types
```
zbanks, line 1040: This is the bit I mention in the description. It's honestly a bit of a hack.

src/translate.py
```diff
@@ -2236,38 +2245,31 @@ def add_imm(source: Expression, imm: Expression, stack_info: StackInfo) -> Expre
         # Pointer addition (this may miss some pointers that get detected later;
         # unfortunately that's hard to do anything about with mips_to_c's single-pass
         # architecture.
-        if isinstance(imm, Literal):
+        # If the imm is exactly a multiple of 0x10000, then it is likely that we are
+        # constructing a large offset (see deref)
+        if isinstance(imm, Literal) and imm.value % 0x10000 != 0:
```
zbanks, line 2250: If a struct has a `char big[1000]` field that crosses the `0x10000`-byte boundary, the new `get_field()` is willing to return an access to the exact element at offset `0x10000`.

This PR ends up *fixing* some big struct accesses, like this one from OOT:
```diff
[INFO] Output of ../oot/asm/non_matchings/overlays/actors/ovl_Boss_Ganon2/func_808FD5F4.s changed! Diff:
--- 

+++ 

@@ -705,10 +705,10 @@

         }
         if (phi_v1_16 >= 0xE5U) {
             arg1->envCtx.unk_E1 = 1;
-            arg1->unk10B08 = 0xFF;
-            arg1->unk10B07 = 0xFF;
-            arg1->envCtx.unk_E2 = 0xFF;
-            arg1->unk10B09 = 0x64;
+            arg1->envCtx.unk_E2[2] = 0xFF;
+            arg1->envCtx.unk_E2[1] = 0xFF;
+            arg1->envCtx.unk_E2[0] = 0xFF;
+            arg1->envCtx.unk_E2[3] = 0x64;
             if (arg0->unk398 == 0xEA) {
                 arg1->envCtx.unk_E1 = 0;
                 arg0->unk39C = 0x18U;
```

src/types.py
```diff
@@ -599,164 +846,106 @@ def unify_with_args(self, concrete: "FunctionSignature") -> bool:
...
+        struct_type = universe.get_struct_for_ctype(ctype)
+        if struct_type is not None:
+            return struct_type
+
+        struct = parse_struct(ctype, universe.typemap)
+
+        typedef_name: Optional[str] = None
+        if id(ctype) in universe.typemap.struct_typedefs:
+            typedef = universe.typemap.struct_typedefs[id(ctype)]
+            assert isinstance(typedef, ca.TypeDecl) and isinstance(
```
simonlindholm, line 906: we could make the type of struct_typedefs more precise to get rid of the first of these isinstance checks

src/types.py
```diff
@@ -599,164 +846,106 @@ def unify_with_args(self, concrete: "FunctionSignature") -> bool:
...
+    """Representation of a C struct or union"""
+
+    @dataclass(eq=False)
+    class StructField:
+        type: Type
+        offset_bits: int
+        name: str
+
+    size_bits: int
+    align_bits: int
```
simonlindholm, line 860: I think bytes would be better, since structs are always byte-aligned. For fields it does make more sense to use bit measurements. Although... I suspect if we introduce bitfields we might want those to live entirely within StructDeclaration and not use Type's with size not divisible by 8? The Type would instead mirror the bitfield's underlying type, or just be "int". That way we prevent those weird types from leaking and accidentally ending up behind pointers etc., and bitfield accesses go through special paths that don't benefit from a uniform representation anyway. I think it's fair to say that we will never auto-detect bitfields, which simplifies the story a bit. Given this train of thought maybe all our usage of bit-based sizes should go away...

simonlindholm (commented):
I've read through types.py, still have translate.py to go

simonlindholm:
Some misc comments on the oot diff, not too important:

```diff
-                D_80130638 = 0U;
+                D_80130638 = 0;
```
what's up with this? The latter admittedly looks nicer, but it feels like it may be unintentional?

```diff
-    if (gAudioContext.unk_2624 != 0) {
+    if (gAudioContext.unk2624 != 0) {
```
unk2624 is a u32 ("sampleDmaListSize1"), why is it no longer getting detected?

```diff
-    u8 (*temp_t0)[9];
+    u8 *temp_t0;
```
interesting, I wonder what happened there

the changes in func_8010C39C are weird

```diff

-            sp3C.unk0 = (s32) arg0->home.pos.x;
+            sp3C.unk0 = arg0->home.pos.x;
             sp3C.unk4 = (s32) arg0->home.pos.y;
```
curious that it fixed just the first one

zbanks:
Thanks for the comments so far, they're super useful, I appreciate it.
Let me respond to each diff separately:

----

> Some misc comments on the oot diff, not too important:
> 
> ```diff
> -                D_80130638 = 0U;
> +                D_80130638 = 0;
> ```
> 
> what's up with this? The latter admittedly looks nicer, but it feels like it may be unintentional?

In `StoreStmt.format`, `elide_casts_for_store` is triggered when `isinstance(dest, StructAccess) and dest.late_has_known_type()`, and in the new version it's succeeding at resolving the type.

----

> 
> ```diff
> -    if (gAudioContext.unk_2624 != 0) {
> +    if (gAudioContext.unk2624 != 0) {
> ```
> 
> unk2624 is a u32 ("sampleDmaListSize1"), why is it no longer getting detected?

I'm admittedly running from an old commit in OOT, my context had `char unk_2624[0x210];` (so the old version is quite misleading). 
If I change the struct to have `s32 sampleDmaListSize1; char unk_2624[0x20C];`, then this line becomes `if (gAudioContext.sampleDmaListSize1 != 0) {`
If I change the struct to have `char unk_2624[0x210/4];`, then this line becomes `if (gAudioContext.unk_2624[0] != 0) {`

----

> 
> ```diff
> -    u8 (*temp_t0)[9];
> +    u8 *temp_t0;
> ```
> 
> interesting, I wonder what happened there

The typing is technically correct either way, but in this case I think the old version is better. I'll circle back after updating the PR with the other comments, in case it changes. 
(Including some more context here, for my reference later:)
 ```diff
 extern u8 D_80131C00[][9];
 ...
-    u8 (*temp_t0)[9];
+    u8 *temp_t0;
 ...
-                        temp_t0 = &D_80131C00[phi_t3_2];
+                        temp_t0 = D_80131C00[phi_t3_2];
```

----

> 
> the changes in func_8010C39C are weird
> 
> ```diff
> -            sp3C.unk0 = (s32) arg0->home.pos.x;
> +            sp3C.unk0 = arg0->home.pos.x;
>              sp3C.unk4 = (s32) arg0->home.pos.y;
> ```
> 
> curious that it fixed just the first one

Agreed it's kinda weird. All values here are floats, and they are just being straight-copied without touching fp regs. So I think the new version is slightly more correct?

It looks like `&sp3C` is detected to have type `f32 *`, so `get_deref_field()` is able to determine that `(&sp3C)[0]` has type `f32` (good). This is printed as `sp3C.unk0` because `has_nonzero_access` is True for this var.
However, `get_deref_field()` does not try to treat pointers as arrays (partially to match old behavior; partially because it has a lot of false positives; the `if self.is_array() or index == 0:` check is what prevents this). As a result, the types of `sp3C.unk4` & `.unk8` are left unknown, and are incorrectly inferred to be `s32`s.
Removing the `if self.is_array() or index == 0:` check causes all 3 lines to be re-written without casts, e.g. `sp3C[1] = arg0->home.pos.y;`. (But, this change causes too many problems elsewhere)

File: src/types.py (hunk 7 above)
Comment (zbanks) [reply src/types.py:860]: This is interesting. I was not feeling particularly excited about supporting bitfields, but I was hoping to not make it *harder* :smile: 

I didn't think too much about it, but I was approximately guessing that bitfields could be represented as say, `Type.int_of_size(3)`, like you mentioned.
I hadn't thought about how these types might leak out though -- that's an excellent point.

I think you made a good argument for using bytes instead, I'll try to make that change.

File: src/types.py (hunk 6 above)
Comment (zbanks) [reply src/types.py:906]: Good call.

If I make the following change, it gets rid of the first check:
```diff
-    struct_typedefs: Dict[Union[str, int], CType] = field(default_factory=dict)
+    struct_typedefs: Dict[Union[str, int], TypeDecl] = field(default_factory=dict)
```
Is there a trick I don't know to encode the `isinstance(typedef.type, ca.IdentifierType)` check?

src/types.py
```diff
@@ -599,164 +846,106 @@ def unify_with_args(self, concrete: "FunctionSignature") -> bool:
...
+                break
+            field_size_bits = field.type.get_size_bits()
+            assert field_size_bits is not None
+            if field.offset_bits + field_size_bits < offset_bits:
+                continue
+            fields.append(field)
+        return fields
+
+    @staticmethod
+    def from_ctype(ctype: Union[ca.Struct, ca.Union], universe: TypeUniverse) -> Type:
```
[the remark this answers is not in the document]
zbanks: Yeah, I admit that is an awkward part of this API.

I ended up in this situation to make it easier to construct self-referential structs (like linked lists).
The code ends up inserting the `Type` object that *will* be returned into the universe before the `StructDeclaration` fields are fully populated. This makes self-reference easy, since it makes the struct `Type` available in the universe while parsing the fields.

I'll try to revisit this.

src/types.py
```diff
@@ -239,67 +320,164 @@ def get_function_pointer_signature(self) -> Optional["FunctionSignature"]:
...
+        """
+        Locate the field in self at the appropriate offset (in bits), and optionally
+        with an exact target size (also in bits).
+        The target size can be used to disambiguate different fields in a union, or
+        different levels inside nested structs.
+
+        The return value is a tuple of `(field_path, field_type, remaining_bits)`.
+        If no field is found, the result is `(None, Type.any(), None)`.
+        If `remaining_bits` is nonzero, then there was *not* a field at the exact
+        offset provided; the returned field is at `(offset_bits - remaining_bits)`.
```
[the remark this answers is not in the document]
zbanks: Oh, so if it fails, just return `offset_bits`? That's a good idea.

src/types.py
simonlindholm: I don't think there is, no, other than changing the pycparser typing stubs to allow "ca.TypeDecl[ca.IdentifierType]" or something

src/types.py
simonlindholm: Yeah, `Type.int_of_size(3)` was my thinking too up until I wrote that comment. I'm also not too excited about bitfields, they are a bit too rare and complex to be worth the trouble

src/types.py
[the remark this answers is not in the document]
simonlindholm: my thinking was 0, but `offset_bits` also makes sense. Just feeling like the value isn't going to be used if the first member is None and making it Optional just adds noisy `is not None` asserts. (Tuple[Optional[Tuple[.., ..]], ..] seems technically more correct but less ergonomic)

src/types.py
```diff
@@ -239,67 +320,164 @@ def get_function_pointer_signature(self) -> Optional["FunctionSignature"]:
...
+            return [], self, 0
+
+        return NO_MATCHING_FIELD
+
+    def get_deref_field(
+        self, offset_bits: int, *, target_size_bits: Optional[int]
+    ) -> _GetFieldResult:
+        """
+        Similar to `.get_field()`, but treat self as a pointer (or array) and find
+        the field in the pointer's target.
```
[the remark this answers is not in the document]
zbanks: Yeah you're totally right. The whole math around `index` isn't needed, because `self` should always be a pointer (never an array). 

I should be able to shrink this function a lot. 

(`get_deref_field()` is the rough equivalent of the old `get_field()` because it does the dereference, whereas the new `get_field()` does not dereference and so is used to traverse the structs/arrays)

src/types.py
zbanks: Changed to bytes; renamed fields :+1:

zbanks:
I resolved most of the comments left so far, except:
- Merging/refactoring `TypeUniverse` + `GlobalInfo`. I think this is a good idea, is a bit tricky (`GlobalInfo` depends on `translate.py` stuff; `TypeUniverse` is needed by `Type`). I need to think the best way to handle it.
- `id(ctype)`. It looks like we'd have to subclass `Node` to get this to work? (I think the current hash is based on the id?)

Commit `a345379` (latest) and `a9d0b8b` (previously reviewed) have the same test output (e2e + OOT).
I identified two changes that affect the test output that I want to fix by the end of the PR and marked them `TODO(@zbanks)`.

src/types.py
```diff
@@ -107,52 +184,57 @@ def unify(self, other: "Type", *, seen: Optional[Set["TypeData"]] = None) -> boo
...
+            assert array_dim is not None
         if x.ptr_to is not None and y.ptr_to is not None:
             if not x.ptr_to.unify(y.ptr_to, seen=seen):
                 return False
         if x.fn_sig is not None and y.fn_sig is not None:
             if not x.fn_sig.unify(y.fn_sig, seen=seen):
                 return False
+        if x.struct is not None and y.struct is not None:
+            if not x.struct.unify(y.struct, seen=seen):
+                return False
```
simonlindholm, line 228: may also want to check that array_dim are the same

src/types.py
```diff
@@ -239,67 +325,172 @@ def get_function_pointer_signature(self) -> Optional["FunctionSignature"]:
...
+        """
+        NO_MATCHING_FIELD: Type.GetFieldResult = (None, Type.any(), offset)
+
+        target = self.get_pointer_target()
+        if target is None:
+            return NO_MATCHING_FIELD
+
+        # Assume the pointer is to a single object, and not an array.
+        size = target.get_size_bytes()
+        if offset < 0 or (size is not None and offset >= size):
```
simonlindholm, line 434: offset < 0 will be checked later by get_field fwiw, and I suspect offset >= size implicitly also is. So I think this check may be redundant

src/types.py
```diff
@@ -239,67 +325,172 @@ def get_function_pointer_signature(self) -> Optional["FunctionSignature"]:
...
+                nonlocal output
+                assert upto >= position
+                if upto > position:
+                    output.append(upto - position)
 
-        return output
+            for field in data.struct.fields:
+                if field.offset < position:
+                    # Overlapping fields, e.g. from unions
+                    continue
```
simonlindholm, line 482: removing this and adding an explicit check for union vs struct seems clearer

src/types.py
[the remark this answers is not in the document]
simonlindholm: I see. Would be useful to have a comment about this if revisiting doesn't work out. (I feel like the same trick should work by inserting the StructDeclaration before it's fully constructed, but I'm probably missing some reason why it gets involved)

simonlindholm (commented):
Responding to comments on types.py, will look at translate.py next

src/types.py
```diff
@@ -133,26 +210,26 @@ def unify(self, other: "Type", *, seen: Optional[Set["TypeData"]] = None) -> boo
...
         if sign != TypeData.ANY_SIGN:
             assert kind == TypeData.K_INT
-        if x.ctype_ref is not None and y.ctype_ref is not None:
-            assert typemap is not None
-            x_ctype = resolve_typedefs(x.ctype_ref, typemap)
-            y_ctype = resolve_typedefs(y.ctype_ref, typemap)
-            if not equal_types(x_ctype, y_ctype):
-                return False
+        if kind == TypeData.K_ARRAY:
+            assert array_dim is not None
```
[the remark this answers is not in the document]
zbanks: Yeah, that's my bad. I removed some assertions and saw that it ran smoothly -- I didn't check if it was creating types with `array_dim=0` :disappointed: 

I think I've fixed it, the change should be in the next commit. The change introduced a single diff in the OOT corpus. (The decl for `dynTable` in my context file is `u8 (*dynTable)[][2];`)
```diff
[INFO] Output of ../oot/asm/non_matchings/code/audio_seqplayer/Audio_SequenceChannelProcessScript.s changed! Diff:
--- 
+++ 
@@ -357,7 +357,7 @@
                     case 182: // switch 1
-                        temp_s6->unk19 = (s8) *(arg0->dynTable + temp_s6->unk19);
+                        temp_s6->unk19 = (s8) arg0->dynTable[0][0][temp_s6->unk19];
                         goto loop_7;
```

src/translate.py
```diff
@@ -1007,42 +1004,71 @@ class StructAccess(Expression):
...
+        output = ""
+        for p in path:
+            if isinstance(p, str):
+                if output or not deref:
+                    output += "."
+                else:
+                    output += "->"
+                output += p
+            else:
+                assert isinstance(p, int)
```
simonlindholm, line 1029: introduce+use static_assert_unreachable instead

src/translate.py
```diff
@@ -1007,42 +1004,71 @@ class StructAccess(Expression):
...
+        return output
 
     def dependencies(self) -> List[Expression]:
         return [self.struct_var]
 
-    def late_field_name(self) -> Optional[str]:
+    def make_reference(self) -> bool:
+        self.late_field_path()
+        if self.field_path and self.field_path[-1] == 0:
+            del self.field_path[-1]
```
simonlindholm, line 1039:
```suggestion
            self.field_path.pop()
```

src/translate.py
```diff
@@ -1007,42 +1004,71 @@ class StructAccess(Expression):
...
     stack_info: StackInfo = field(compare=False, repr=False)
     type: Type = field(compare=False)
-    has_late_field_name: bool = field(default=False, compare=False)
+    checked_late_field_path: bool = field(default=False, compare=False)
+
+    @staticmethod
+    def access_path_to_field_name(path: Optional[AccessPath], deref: bool) -> str:
+        """
+        Convert a path like `["foo", 3, "bar"]` into a field name like `.foo[3].bar`.
+        If deref is True, the initial `.` is replaced by `->`.
```
simonlindholm, line 1016: if deref is true and path is empty or has an integer as first component, this fails. Should that be asserted against or signalled via return value?

src/translate.py
```diff
@@ -4255,10 +4273,9 @@ def make_arg(offset: int, type: Type) -> PassedInArg:
         assert offset % 4 == 0
         return PassedInArg(offset, copied=False, stack_info=stack_info, type=type)
 
-    if typemap and function.name in typemap.functions:
-        fn_type = type_from_ctype(typemap.functions[function.name].type, typemap)
-        fn_decl_provided = True
-    else:
+    fn_type = universe.get_function_type(function.name)
```
simonlindholm, line 4276: get_function_type -> get_known_function_type maybe? fn_decl_provided expects to use "is defined in context" as a heuristic for something, but `get_function_type` sounds like it may make function signatures up. (I do wonder if that knowledge of types can't move from GlobalInfo to TypeUniverse without too much pain.)

src/translate.py
```diff
@@ -2741,16 +2765,20 @@ def function_abi(fn_sig: FunctionSignature, *, for_call: bool) -> Abi:
         only_floats = False
 
     for ind, param in enumerate(fn_sig.params):
-        size, align = param.type.get_size_align_bytes()
+        param_type = param.type
+        if param_type.is_array():
+            # Array arguments decay into pointers
```
simonlindholm, line 2770:
```suggestion
            # Array parameters decay into pointers
```

src/translate.py
```diff
@@ -2741,16 +2765,20 @@ def function_abi(fn_sig: FunctionSignature, *, for_call: bool) -> Abi:
...
+            param_type = param_type.reference()
+        size, align = param_type.get_size_align_bytes()
```
simonlindholm, line 2772: we should rename get_size_align_bytes -> get_parameter_size_align_bytes btw, to ensure it's not used by accident (it contains a sketchy "or 32" that only makes sense for function calls, if even then)

src/translate.py
```diff
@@ -4000,16 +4023,12 @@ def address_of_gsym(self, sym_name: str) -> AddressOf:
...
-                sym.array_dim = dim
+            ctype_type = self.universe.get_var_type(sym_name)
+            if ctype_type is not None:
                 sym.type_in_typemap = True
-                type.unify(ctype_type)
-                type = ctype_type
-        return AddressOf(sym, type=type)
+                sym.type.unify(ctype_type)
+
+        return AddressOf(sym, type=sym.type.reference())
```
simonlindholm, line 4031: pre-existing, but it's a bit weird that we decay array to pointer here for the type but not for the value. iirc there is some reason for this, but it doesn't easily come to mind at 2 AM :p, and we should really have some comments that explains the setup for arrays. I'll try to remember it when I next continue reviewing...

simonlindholm (commented):
Some preliminary comments on translate.py, more later

src/translate.py
zbanks: Added an assert, clarified the docstring, and made `StructAccess.format()` super explicit to avoid this case. 
(This case wasn't supposed to happen before, but the checks I had in format weren't specific enough)

src/translate.py
```diff
@@ -1062,27 +1091,34 @@ def format(self, fmt: Formatter) -> str:
...
-        else:
-            needs_deref = True
+        # Rewrite `x[0].y` to `x->y` by dropping the initial `0` in field_path
+        if (
+            field_path
+            and len(field_path) >= 2
+            and field_path[0] == 0
+            and isinstance(field_path[1], str)
+        ):
+            field_path = field_path[1:]
```
simonlindholm, line 1103: While I've convinced myself that this is correct, it feels sketchy to strip a part of the field path like this. The reason it's correct is that access_path_to_field_name does the semantically wrong thing when the path starts with a number: the natural thing to do given e.g. [1, "foo"] would be to output `x->[1].foo` (which is invalid C but the natural reading is `(*x)[1].foo`), but instead it strips the `->` and outputs `x[1].foo`. This stripping is done precisely when it's not done here, so the same amount of stripping happens. Not sure off hand how to best structure this, a comment may be helpful if nothing else.

src/translate.py
```diff
@@ -2232,42 +2259,34 @@ def add_imm(source: Expression, imm: Expression, stack_info: StackInfo) -> Expre
         # addiu $reg1, $reg2, 0 is a move
         # (this happens when replacing %lo(...) by 0)
         return source
-    elif source.type.is_pointer():
+    elif source.type.is_pointer_or_array():
         # Pointer addition (this may miss some pointers that get detected later;
         # unfortunately that's hard to do anything about with mips_to_c's single-pass
         # architecture.
```
simonlindholm, line 2265:
```suggestion
        # architecture).
```

src/translate.py
```diff
@@ -2232,42 +2260,34 @@ def add_imm(source: Expression, imm: Expression, stack_info: StackInfo) -> Expre
...
-        if isinstance(imm, Literal):
+        # If the imm is exactly a multiple of 0x8000, then it is likely that we are
+        # constructing a large offset (see deref)
+        if isinstance(imm, Literal) and imm.value % 0x8000 != 0:
```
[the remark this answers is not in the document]
simonlindholm: hm, why doesn't the 0x7FFF case come up here? an example of it is in mm asm/overlays/ovl_Boss_07_0x809F4980.asm

zbanks:
I realized that the `TypeUniverse`/`TypeMap` isn't actually needed by `Type` -- it's only needed when constructing the `Type`. So in `df5d7728` I did a refactor to merge `TypeUniverse` + `GlobalInfo`. 

The second recent commit, `437453b`, reworks the field path logic and I think addresses all of your open comments about them. Now all non-`None` `StructAccess.field_path`s start with an `int`.

src/translate.py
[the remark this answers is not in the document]
zbanks: You're right, I decided to factor this out into `Literal.likely_partial_offset()`.
I think it wasn't an issue there because there wasn't a field at exactly 0x7FFF? (it's much more common for there to be a field at 0x8000?)

src/translate.py
```diff
@@ -1038,12 +1043,11 @@ def access_path_to_field_name(path: Optional[AccessPath], deref: bool) -> str:
...
 
-    def make_reference(self) -> bool:
-        self.late_field_path()
-        if self.field_path and self.field_path[-1] == 0:
-            self.field_path.pop()
-            return True
-        return False
+    def make_reference(self) -> Optional["StructAccess"]:
+        field_path = self.late_field_path()
+        if field_path and field_path[-1] == 0:
```
simonlindholm, line 1048: `and len(field_path) >= 2`

src/types.py
```diff
@@ -24,77 +25,25 @@
...
-class TypeUniverse:
+class GlobalTypeInfo:
     """
-    Mutable shared state for Types, currently maintaining the set of available
-    struct types.
-    Unlike TypeMap, which is immutable and can be reused across tests/threads,
-    a new TypeUniverse instance should be created for each isolated run.
-    However, if the object is reused between functions, it may help provide
-    cross-function type resolution.
+    Mutable shared state for Types, implemented by GlobalInfo in translate.py.
```
simonlindholm, line 30: Hrm, this circular reference to translate.py actually rather bothers me -- it makes types.py no longer self contained. I wonder if the right split may actually be info about structs in types.py, and info about variables in translate.py.

tests/end_to_end/function-pointer2/irix-o2-out.c
```diff
@@ -1,5 +1,5 @@
-s32 bar(f32); // extern
-extern s32 (*glob2)(f32);
```
simonlindholm, line 2: (it'd be nice to avoid parameter names in types. but that's pre-existing ofc)

src/translate.py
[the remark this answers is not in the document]
zbanks: I screwed up my testing of the last commit: it actually introduced a diff in OOT:
https://gist.github.com/zbanks/2e20069ef8f2d0e72b6b63c074a4c23d

All of the diffs from that change appear to be related to the `MessageContext` struct and `addiu ..., 0x7fff` instructions. The output seems completely improved.

src/types.py
zbanks: Refactored again in `2b420a6` -- I think it's converging on a good solution? :smile:

tests/end_to_end/function-pointer2/irix-o2-out.c
zbanks: This is actually an easy fix, I could do it after this PR? (it'll have a big diff)

src/translate.py
```diff
@@ -2583,38 +2599,48 @@ def array_access_from_add(
...
-            field_name=field_name,
+            field_path=field_path,
             stack_info=stack_info,
-            type=ptr_type if array_dim is not None else new_type,
+            type=field_type,
         )
 
-    if ptr and array_dim is None:
-        ret = AddressOf(ret, type=ptr_type)
+    if ptr:
```
simonlindholm, line 2641: This is an interesting diff -- it shows that we previously *did* avoid AddressOf wrapping for arrays like I was thinking in my earlier comment. I feel like the difference should be visible if the result is dereferenced? Admittedly the type juggling logic in this function loses me completely, so it may for all I know be equivalent or more correct. We could push looking into this to future work.

A disadvantage of hiding `&` in formatting of AddressOf-with-array-type is that it covers up bugs with this pretty well!

src/translate.py
```diff
@@ -2623,9 +2649,9 @@ def handle_add(args: InstrArgs) -> Expression:
     rhs = args.reg(2)
     stack_info = args.stack_info
     type = Type.intptr()
-    if lhs.type.is_pointer():
+    if lhs.type.is_pointer_or_array():
         type = Type.ptr()
-    elif rhs.type.is_pointer():
+    elif rhs.type.is_pointer_or_array():
```
simonlindholm, line 2654: hm, can array actually happen here? I feel like all expressions in registers should have decayed types. not quite assert-worthy though, and I guess it's fine to handle it like pointers

src/translate.py
```diff
@@ -2741,16 +2767,20 @@ def function_abi(fn_sig: FunctionSignature, *, for_call: bool) -> Abi:
         only_floats = False
 
     for ind, param in enumerate(fn_sig.params):
-        size, align = param.type.get_size_align_bytes()
+        param_type = param.type
+        if param_type.is_array():
+            # Array parameters decay into pointers
```
simonlindholm, line 2772: I wonder if a `.decay()` method on Type would be useful

simonlindholm (approved):

src/translate.py
zbanks: Yeah I think you're right; it's in a register, an array shouldn't be possible.
I'll leave a comment about this, in case it comes up in a later refactor? (I think leaving it as `.is_pointer_or_array()` will also help later refactors, even if it's not needed?)

simonlindholm: Some kind of assert-only-in-debug-mode would be useful here. If it comes up we'd want to know about it since it's a probable bug, but at the same time it's a complex invariant so it's hard to be 100% sure about a hard assert, and there's an easy fallback path. Maybe too much effort to be worth introducing machinery for, though.

src/translate.py
zbanks: I had been going back-and-forth on this :+1:

src/translate.py
zbanks: Yeah - it's a good catch.
The previous version of the code seems tenuous to me -- if the inner field is an array (`array_dim is not None`), then the returned expression doesn't depend on `ptr`? 

I think the new version is more consistent -- though at the cost of decaying arrays.

src/translate.py
```diff
@@ -1062,6 +1067,9 @@ def late_field_path(self) -> Optional[AccessPath]:
                 self.field_path = field_path
                 self.type.unify(field_type)
 
+                # Check invariant on self.field_path
+                self.__post_init__()
```
zbanks, line 1071: Is this weird? (I could make the invariant check a separate fn?)

simonlindholm: Looks rather goofy. I don't mind terribly but probably best to put it in a separate function yeah

PR #146, opened 2021-10-02
Avoid parameter names in function pointer types

zbanks:
Follow up from https://github.com/matt-kempster/mips_to_c/pull/144#discussion_r720491490 -- this is what you were asking for, right?

Diff from OOT/e2e:
https://gist.github.com/zbanks/f9382bf2d5d42cb1a1d0da0c7f4f4fad

### DIFF
--- a/src/types.py
+++ b/src/types.py
@@ -527,8 +527,17 @@ class Type:
 
         if data.kind == TypeData.K_PTR:
             if data.ptr_to is None:
-                return ca.PtrDecl(type=simple_ctype("void"), quals=[])
-            return ca.PtrDecl(type=data.ptr_to._to_ctype(seen, fmt), quals=[])
+                target_ctype = simple_ctype("void")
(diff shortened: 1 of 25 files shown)


simonlindholm:
I think names in actual function declarations are fine, it's just function pointer types I feel like they don't belong in.

zbanks:
Latest commit `335804` has a diff in only 1 e2e test, and has zero diffs in the OOT corpus. 

Though FWIW, I had trouble finding function pointers in the OOT corpus. I only found one by grepping, and it didn't have parameter names:
```
tests/project/oot_ctx/non_matchings/code/ucode_disas/UCodeDisas_Disassemble.c
1417:                ((? (*)(UCodeDisas *, u16)) sp370.unk4)(arg0, sp370.unk2);
```

simonlindholm (approved):

PR #151, opened 2021-10-04
Use a struct to represent the stack

zbanks:
The `StackInfo.get_unique_type("stack", ...)` mechanism is replaced by keeping track of the types on a `StructDeclaration`, which can optionally be provides as part of the context.

Adding the `exact` kwarg to `Type.get_field()`/`.get_deref_field()` does two things. First, it makes the usage more ergonomic, now that most callers can ignore the `remaining_offset` component of the return value. Second, when `exact=True`, this is an opportunity to insert a new field into the struct, if the struct has a hole at the given offset. (As of this PR this is only enabled for the stack struct!)

This PR has zero diffs in the e2e tests or OOT corpus, with the exception of the newly-added test. 

---

This changeset provides most of the groundwork for allowing structs on the stack, and for detecting the types of unknown regions of structs provided in the context. To get there, these are the next steps after this PR:
1. Try to add fields to any struct, not just the context struct. This introduces a few diffs in the OOT corpus, so I'd like to review it separately (maybe combined with 2?). This is marked as `TODO` in types.py.
2. When adding fields, use `Type.any()` instead of `Type.any_reg()`. This seems to mostly work, but introduces a lot of diffs and can end up with overlaps. (With the expectation that overlaps are fixed by providing better types in the context!)
3. Add a CLI arg to mark "holes" in context structs. This could maybe be done with fields matching a provided prefix, or fields with a certain typedef (the latter would be preferred, if projects could switch their `UNK_TYPE` macros to typedefs)
4. Add a CLI arg to include the "deduced" structs in the output. This would include both the stack struct, as well as any structs from the context that have holes that were filled in. This would make it drastically easier to specify the stack struct by giving the user a template to work from.


zbanks:
As of `6dca952`, the only existing e2e test that changed is [`doubles/irix-g-mips1`](https://github.com/matt-kempster/mips_to_c/pull/151/files#diff-fd806aef0beb0e23cfb107dbba43a002e5c6f21d82570f0e2c9e1aa41a868906) -- the test where mips2c has a hard time detecting the `double`. (I think the test output is marginally better for omitting the fake `sp4` decl)

PR #152, opened 2021-10-06
Allow multiple `--context` files; add caching/`--no-cache`

zbanks:
Parse multiple `--context` files in sequence. Had to add a slight hack with `CParser()` to preserve a small bit of state, which is stored in `TypeMap`. The order of files matters.

Added caching, where `ctx.c` is cached in a pickle file at `ctx.c.m2c`. A hash of the inputs to `build_typemap` is stored in the `TypeMap` itself and used to validate that the cache is still valid.

On my computer, loading a huge context, like from MM, takes about 5 seconds on cache miss, and about 250ms on cache hit. (Most of that time seems to be in unpickling; not file I/O or computing the hash. Both times are just for `build_typemap()` and not the decomp steps)

---
Things I am unsure about & would appreciate feedback:

Caching might seem like surprising behavior because it creates a file, so I left it disabled by default. Also unpickling is theoretically a security hazard (even if the situation would be contrived here). Maybe caching could be enabled by default, and disabled with `--no-cache`?

Is `ctx.c` --> `ctx.c.m2c` a good naming scheme? I also considered `.ctx.c.m2c` (so it would be hidden), `ctx.c.m2c_cache` (explicit, but verbose), or `ctx.c.m2c.c` (so that editors would know how to syntax highlight). 
I think it's better to store the cache alongside the context to avoid accidentally building up a giant cache folder that the user isn't explicitly aware of? Happy to rewrite this bit though.

Because the caches depend on all of the files before them, it's possible to thrash by repeatedly switching between something like:
- `./mips_to_c.py --context a.c --context b.c ...`
- `./mips_to_c.py --context b.c --context a.c ...`

I don't think this is a major issue though? I can't think of a reason for wanting to do this.

### DIFF
--- a/src/c_types.py
+++ b/src/c_types.py
@@ -84,7 +84,9 @@ class TypeMap:
     VERSION: ClassVar[int] = 1
     _empty: ClassVar["TypeMap"]
 
-    source_hash: str
+    cparser_scope: CParserScope = field(default_factory=dict)
+    source_hash: Optional[str] = None
+
     typedefs: Dict[str, CType] = field(default_factory=dict)
     var_types: Dict[str, CType] = field(default_factory=dict)
     functions: Dict[str, Function] = field(default_factory=dict)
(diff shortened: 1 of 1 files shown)


zbanks:
It was premature of me to request a review, I have some bugs to work out in the latest change! :grimacing:

PR #162, opened 2021-10-29
Add "array-like" heuristic to `Type.get_deref_field()`

zbanks:
Array-like pointers will use an array access when trying to deference an offset outside of the bounds of the target type. Types that are not array-like will use the standard old behavior, like `->unk_123`.

- When building the `TypeMap`, track which structs are `extended_by_other_structs`. Pointers to structs are considered array-like iff they do not have this marker.
    - This flag is set if there is at least one field at offset 0 in _another struct_ with this struct's type, **but** there are no fields at other offsets with this struct's type.
    - This detects the OOP-style polymorphism that we see with `Actor` in OOT/MM: `Actor` & `DynaPolyActor` are marked as `extended_by_other_structs`.
    - This does not detect "internal" base structs -- cases where you would use the `offsetof` macro to calculate the address of the outer struct.
- ~Pointers to ints, floats, other pointers, and functions are always considered "array-like"~
   - ~If the pointer is unknown (`ptr_to=None`) or the `kind` could be one of the "bad" types, it's not array-like.~
   - For now, pointers-to-primitives are not considered array-like

----

[Some regression test output here](https://gist.github.com/zbanks/e0af3f6c6a87cae0c9e5aaf00424f0fc). This output is with a hack to treat `Gfx` types as *not* array-like just to keep the diff size under control, and a bit easier to scan. The raw diff is about 7x larger. `Gfx` diffs look like the following:
```diff
[INFO] Output of ../oot/asm/non_matchings/code/z_kankyo/func_80074FF4.s changed! Diff:
     temp_v1 = sp34->polyOpa.p;
-    sp34->polyOpa.p = temp_v1 + 8;
+    sp34->polyOpa.p = &temp_v1[1];
     temp_v1->words.w0 = 0xFA000000;
     temp_v1->words.w1 = (arg1 << 0x18) | (arg2 << 0x10) | (arg3 << 8) | arg4;
     temp_v1_2 = sp34->polyOpa.p;
-    sp34->polyOpa.p = temp_v1_2 + 8;
+    sp34->polyOpa.p = &temp_v1_2[1];
     temp_v1_2->words.w1 = 0;
```

I think most of the changes in the link above are good, but here's some that are worse:

- Many of the functions that use `gAudioContext` [are worse](https://gist.github.com/zbanks/e0af3f6c6a87cae0c9e5aaf00424f0fc#file-arraylike_nogfx-diff-L123-L156) because they try to do `(&gAudioContext)[1]`. This happens because I'm working from an older version of the OOT repo, which has a `struct AudioContext` that is too small. 

- There's a [weird diff](https://gist.github.com/zbanks/e0af3f6c6a87cae0c9e5aaf00424f0fc#file-arraylike_nogfx-diff-L7760-L7761) with `gSaveContext[-1]` that was originally written `&gSaveContext - 3 + temp_s2`.


simonlindholm (commented):
I'm rather worried about the false positives here... The diff shows things like

```diff
-                        if ((sp74->unk_138C & 0x20) == 0) {
+                        if ((sp74[0x271].center.z & 0x20) == 0) {
```

```diff
-        phi_s1 += 0x64;
+        phi_s1 = (DynaLookup *) &phi_s1[0xC].wall;
```

```diff
-    phi_s1 = arg0->unk_1450 + (temp_v0 * 4);
+    phi_s1 = arg0[0x1B1].y + (temp_v0 * 4);
```

```diff
-        temp_v1->unk_-54 = 0;
+        temp_v1[-1].elements[0xE].p1.x = 0;
```

and without seeing the real offset it's hard to fix these errors up manually.

zbanks:
> I'm rather worried about the false positives here... The diff shows things like

Most (all?) of these seem to be fixed from other changes to `master`. I'll pull & regenerate the diff and take a close look for other false positives. Even so, your point still stands that false positives are always going to be much worse to fix up manually than the raw offsets.

zbanks:
New diff output:
- [MM, without `Gfx`](https://gist.github.com/zbanks/cd518a8cbd4a19b5b94278a99aa45a97) (most diffs are related to the `Gfx` struct, but aren't very interesting to look at)
- [MM, all](https://gist.github.com/zbanks/333c7154dbfe37ebe146d8207f9b085b)
- [OOT, all](https://gist.github.com/zbanks/f5fc789eb5787585ce6e870ea9f87d29)
- [PM, all](https://gist.github.com/zbanks/a2649ab9ae3657951d2168acc35ddfd7)

PR #164, opened 2021-11-01
Simplify gcc's MULT_HI expressions into divs/mods

zbanks:
While working on this, I also found that GCC will use `sllv`/`srlv`/`srav` instructions for optimized constant muls/divs, whereas IDO seems to only use the `sll`/`srl`/`sra`. Adding `fold_mul_chains()` to these instructions does not affect the OOT/MM regression tests, so I think this change should be safe.

---

There is [one diff in MM](https://gist.github.com/zbanks/21b5cdbbd46c68090c2c7f9022a8877a#file-multihi_ootmm-diff) from the `%` detection, which looks like a `divmod` macro?

[Diffs in PM](https://gist.github.com/zbanks/21b5cdbbd46c68090c2c7f9022a8877a). 

There are [100 `MULT_HI`](https://gist.github.com/zbanks/21b5cdbbd46c68090c2c7f9022a8877a#file-multhi_pm_leftovers-c) instances left in PM, which primarily involve phis or (incorrect) extra function arguments. Two examples of these:
```c
tests/project/papermario_us_ctx/nonmatchings/battle/action_cmd/power_shock/func_802A9310_42D220.c
127:            temp_t0 = ((s32) MULT_HI((s16) temp_v1_2, temp_s0_5) >> 5) - ((s32) (temp_v1_2 << 0x10) >> 0x1F);
136:                temp_v0_3 = ((s32) MULT_HI(phi_a1, temp_s0_5) >> 5) - (phi_a1 >> 0x1F);

tests/project/papermario_us_ctx/nonmatchings/battle/action_cmd/dizzy_shell/func_802A96DC_42684C.c
24:        func_80268770(sp10, sp14, (s16) (gActionCommandStatus.unk_44 / 0x64), MULT_HI((s16) gActionCommandStatus.unk_44, 0x51EB851F));
26:        func_8026880C(sp10, sp14, (s16) (gActionCommandStatus.unk_44 / 0x64), MULT_HI((s16) gActionCommandStatus.unk_44, 0x51EB851F));
```


tests/end_to_end/gcc-division/gcc-o2-out.c
```diff
@@ -0,0 +1,93 @@
...
+    func_00400090(x - ((phi_a0_6 >> 8) << 8));
+    func_00400090(x % 0x168);
+    func_00400090(x % 0x3E8);
+    func_00400090(y >> 1);
+    func_00400090(y / 3);
+    func_00400090(y >> 2);
+    func_00400090(y / 5);
+    func_00400090(y / 6);
+    temp_hi = y / 7;
+    func_00400090((u32) (temp_hi + ((u32) (y - temp_hi) >> 1)) >> 2);
```
zbanks, line 69: The `y / 7` codegen from GCC is super weird. I didn't think this was worth supporting.

simonlindholm: Makes sense. Still, it's a bit sad that we're outputting code that's actively incorrect, and will confuse people if/when it comes up (I assume it's not just 7 but some small proportion of numbers?). OTOH, it's nice that we're doing *something* in this case, because MULT_HI's are empirically very hard for people to parse, and it's not that hard to figure out you need to remove the weird sign bit correction thing. Ideally we'd emit a "partially detected division" comment when there's no sign bit correction maybe. But I understand if you feel that's too much effort.

src/translate.py
```diff
@@ -2608,6 +2627,137 @@ def format_f64_imm(num: int) -> str:
...
+        isinstance(left_expr, BinaryOp)
+        and expr.op == ">>"
+        and isinstance(right_expr, Literal)
+    ):
+        divisor_shift += right_expr.value
+        expr = left_expr
+        left_expr = early_unwrap_ints(expr.left)
+        right_expr = early_unwrap_ints(expr.right)
+
+    # Remove inner error term: MULT_HI(x, N) + x --> MULT_HI(x, N)
```
simonlindholm, line 2717: This looks really weird to me. It MULT_HI(x, N) + x == MULT_HI(x, N + 2^32), I would have thought...

tests/end_to_end/gcc-division/orig.c
```diff
@@ -0,0 +1,64 @@
...
+    func_00400090(y % 7);
+    func_00400090(y % 8);
+    func_00400090(y % 9);
+    func_00400090(y % 10);
+    func_00400090(y % 100);
+    func_00400090(y % 255);
+    func_00400090(y % 256);
+    func_00400090(y % 360);
+    func_00400090(y % 1000);
+}
```
simonlindholm, line 64: I wouldn't mind seeing this extended even further, at least for say signed int division. Maybe 11..33, 257, numbers closer to int max. And s8/s16 to test that branch.

tests/end_to_end/gcc-division/orig.c
```diff
@@ -0,0 +1,64 @@
+void func_00400090(unsigned int x) {
+}
+
+void test(volatile int x, volatile unsigned int y) {
```
simonlindholm, line 4: (why volatile?)

simonlindholm (approved):

tests/end_to_end/gcc-division/orig.c
zbanks: Added more test cases. I had to split them into separate functions, because GCC started putting the magic constants on the stack.

src/translate.py
```diff
@@ -2608,6 +2627,137 @@ def format_f64_imm(num: int) -> str:
...
+    # Use `early_unwrap_ints` instead of `early_unwrap` to ignore Casts to integer types
+    # Although this discards some extra type information, this function largely ignores
+    # sign/size information to stay simpler. The result will be made with BinaryOp.int()
+    # regardless of input types.
+    expr = original_expr
+    left_expr = early_unwrap_ints(expr.left)
+    right_expr = early_unwrap_ints(expr.right)
+    divisor_shift = 0
+
+    # Fold `/` with `>>`: ((x / N) >> M) --> x / (N << M)
```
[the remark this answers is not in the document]
zbanks: Added a warning about this - I think that's a good compromise for now?

src/translate.py
zbanks: I added some more comments for this step. Because `MULT_HI` is signed, if `N`'s highest bit is set, then it represents `(s32) ((u32) N - 2^32)` -- so adding back the `2^32` is sort of like calculating `(s32) x * (u32) N`.

tests/end_to_end/gcc-division/gcc-o2-out.c
```diff
@@ -0,0 +1,373 @@
...
+    temp_hi_5 = MULTU_HI(arg0, ((s32) &D_8421085 | 0));
+    func_00400090((u32) (temp_hi_5 + ((u32) (arg0 - temp_hi_5) >> 1)) >> 4);
+    func_00400090(arg0 >> 5);
+    func_00400090(arg0 / 0x21);
+    func_00400090(arg0 / 0x64);
+    func_00400090(arg0 / 0xFF);
+    func_00400090(arg0 / 0x168);
+    func_00400090(arg0 / 0x3E8);
+    func_00400090(arg0 / 0x2710);
+    func_00400090((u32) MULTU_HI((arg0 >> 5), ((s32) &D_A7C5AC5 | 0)) >> 7);
```
simonlindholm, line 294: looks like the "add test" script misidentified this as a symbol instead of a constant

zbanks: Yeah - I was going to leave this alone? In case we want to to fix this kind of error in the future? (Or, I could manually edit the `.s` if you don't like that)

simonlindholm: Depends on if we have a concrete plan on actually fixing the import step to disassemble .o files correctly -- if not I'd prefer to manually fix up the .s. I've been treating the import process as a single-time thing, with e.g. manual fixes to make jump tables work, removing the IDO -g version if not relevant, etc., and I don't think existing tests have these kind of broken import artifacts.

zbanks: Ah - I meant if we want to do something about it on the m2c side. (Warn that the symbol is fake, or try to convert it to a literal)
I'll just edit the .s though.

simonlindholm: ah, I see! I think it's fine to leave that to manual cleanup, mis-identified literals should be rare given a good disassembly, particularly so combined with division

PR #169, opened 2021-11-06
Use base 10 for formatting RHS of / and %

zbanks:
[Here's the diff for MM + OOT](https://gist.github.com/zbanks/037a3e987445e164b03ad67b157c75fd). 

I didn't do it for `*` because I think a lot of `*`s are undetected array accesses? And I think it's easier to manually match hex constants to struct sizes because struct sizes are usually documented in hex? I'm guessing this will be controversial point though :) 
This at least presents a pattern that could be tweaked with some formatting options later?

The `gcc-division` tests are a bit harder to read... but because they admittedly have intentionally-difficult divisors.


simonlindholm (approved):
Judging by the oot/mm diff we could consider keeping hex for powers of two... but I think this is good, even powers of two are imo often more readable as decimal.

PR #170, opened 2021-11-06
Detect substruct pointers at offset 0 in as_type()

zbanks:
Last PR of the day! :sweat_smile: Some more code I had stashed. 

This closes [this ticket on Trello](https://trello.com/c/R2aPq4AG/63-pass-this-actor-instead-of-actor-this).  When casting to a pointer type, `as_type` checks if it can get the pointer type from a subfield on `expr`. This leads to improvements like:

```diff
-    Actor_SetScale((Actor *) this, 0.01f);
+    Actor_SetScale(&this->actor, 0.01f);
```

[Full diff for MM + OOT](https://gist.github.com/zbanks/6bea1d89695a4e4e2bc28eb0a0c47acd). Nothing too surprising...
There are a few functions which whose deduced types were changed from `void func_80C17468(void **arg0);` to `void func_80C17468(GraphicsContext **arg0);` which helped produce better code.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -167,9 +167,29 @@ def current_instr(instr: Instruction) -> Iterator[None]:
 
 def as_type(expr: "Expression", type: Type, silent: bool) -> "Expression":
     type = type.weaken_void_ptr()
+    ptr_target_type = type.get_pointer_target()
     if expr.type.unify(type):
         if silent or isinstance(expr, Literal):
             return expr
+    elif ptr_target_type is not None:
+        ptr_target_type_size = ptr_target_type.get_size_bytes()
+        field_path, field_type, _ = expr.type.get_deref_field(
+            0, target_size=ptr_target_type_size
+        )
+        if field_path is not None and field_type.unify(ptr_target_type):
+            expr = AddressOf(
+                StructAccess(
+                    struct_var=expr,
+                    offset=0,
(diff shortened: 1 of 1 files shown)


src/translate.py
```diff
@@ -1120,11 +1140,15 @@ class StructAccess(Expression):
     offset: int
     target_size: Optional[int]
     field_path: Optional[AccessPath] = field(compare=False)
-    stack_info: StackInfo = field(compare=False, repr=False)
+    stack_info: Optional[StackInfo] = field(compare=False, repr=False)
```
simonlindholm, line 1143: (this is pretty ugly, we should probably make this a thread-local at some point...)

simonlindholm (approved):
Nice, that's quite a big improvement

src/translate.py
zbanks: For fun, I tried putting `has_nonzero_access` in `TypeData`, so that `record_struct_access()` & `has_nonzero_access()` are done on `Type`s instead of using `StackInfo` & `Expression`s. 

As expected, it changed a lot of `*x` into `x->unk_0`, but maybe some of these would be better written as `x[0]`? There's also some changes around `GlobalSymbol`s, those may need additional work too? [Full diff on OOT+MM](https://gist.github.com/zbanks/19923b394923b1b11083e1f4338a292c).

PR #171, opened 2021-11-08
Add --passes/-P for running multiple passes over the input

zbanks:
This generally improves type inference, but causes mips_to_c to run slower. More than 2 seems to not make a difference for single-function decomps.

`--passes=1` is the old behavior, `--passes=2` is the new default behavior and is about 50% slower than before (this trend is approximately linear afaict).

[Diffs from OOT, MM, PM](https://gist.github.com/zbanks/5f6fa0dfb278aa6674e684ba0e451504). Sorry that they are too large for github to display inline. There are a few that are worse where there was some incorrect struct type inference: this isn't great, but it should be relatively straightforward to identify & correct manually?

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -8,7 +8,7 @@ from typing import Dict, List, Optional, Union
 from .c_types import TypeMap, build_typemap, dump_typemap
 from .error import DecompFailure
 from .flow_graph import visualize_flowgraph
-from .if_statements import get_function_text, function_format_pass
+from .if_statements import get_function_text
 from .options import CodingStyle, Options
 from .parse_file import AsmData, Function, parse_file
 from .translate import (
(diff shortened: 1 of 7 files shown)


src/translate.py
```diff
@@ -4372,6 +4373,16 @@ def address_of_gsym(self, sym_name: str) -> AddressOf:
 
         return AddressOf(sym, type=sym.type.reference())
 
+    def flow_graph(self, function: Function) -> FlowGraph:
```
simonlindholm, line 4376: might be cleaner to pass this as an argument to translate_to_ast rather than memoize it

src/main.py
```diff
@@ -109,59 +109,68 @@ def run(options: Options) -> int:
...
-        return_code = 1
 
-    for index, (function, function_info) in enumerate(zip(functions, function_infos)):
-        if index != 0:
-            print()
-        try:
-            if options.print_assembly:
-                print(function)
-                print()
+    for pass_number in range(options.passes):
```
simonlindholm, line 113: (if options.passes <= 0 we'll crash by referring to return_code while it's uninitialized. I bet someone's going to do it. :p)

not sure if the "is_final_pass" thing is actually cleaner than the loop up to options.passes-1 that you had before.

simonlindholm (approved):

src/main.py
zbanks: I'll revert to the loop option.

PR #172, opened 2021-11-08
Support "weak" stack var types which may change

zbanks:
This handles compiler-managed variables whose types may change in different blocks. I'm calling these "weak" but maybe there's a better term ("compiler-managed" is too long).

[Diff for MM](https://gist.github.com/zbanks/eb653931eddba526b6a827c452cbc7e5). Some of the diffs in the e2e tests are false-positives, because the tests are for weirder edge cases. But, it seems to fix some of the weird type inferences like in [`EnHorse`](https://gist.github.com/zbanks/eb653931eddba526b6a827c452cbc7e5#file-weakstack_mm-diff-L3705).

It seems like these variables are low on the stack? Maybe it's worth only using this logic in a user-configured region, but I'm not sure what the interface would look like.


simonlindholm (commented):
I like the approach (and diff), though I'm a bit vary of putting this state in StackInfo, where control-flow-sibling blocks can overwrite state for each other. That is, `field_type = self.weak_stack_var_types[location]` could read a type from a sibling instead of a parent. Would it be practical to put it in RegInfo? Phi nodes would be a problem, of course, so maybe in practice it's not practical and this is what we should do... Definitely deserves some thinking and a comment, at least.

(Re ovl_Player_Actor I'm not super happy about how it's adding an overlapping struct member, but that's not the fault of this PR, so eh)

zbanks:
> I like the approach (and diff), though I'm a bit vary of putting this state in StackInfo, where control-flow-sibling blocks can overwrite state for each other. That is, `field_type = self.weak_stack_var_types[location]` could read a type from a sibling instead of a parent. Would it be practical to put it in RegInfo? Phi nodes would be a problem, of course, so maybe in practice it's not practical and this is what we should do... Definitely deserves some thinking and a comment, at least.

You're right -- I think the most correct way to implement would require something like PhiNode for stack vars? 

Here's an example from `doubles/irix-g-out.c`. Although it doesn't involve compiler-managed stack vars, it demonstrates one issue:
```c
    if ((sp8 < arg4) || (sp8 == arg4) || (sp8 > 9.0)) {
        sp0 = 5.0;
    } else {
        sp0 = 6.0;
    }
    D_410180 = sp0;
```

If I move the weak tracking into RegInfo, then there's a phi-like situation when handling the `D_410180 = sp0;` block: *something* should check that the two branches used compatible types for `sp0`.

---

I added some caveats to the comments, and I think it's worth committing this imperfect implementation? But let me know if I should expand on the comment more? 
Otherwise, I'm not particularly excited about the idea of adding a new phi system to manage stack vars: that seems like a lot of complexity?

src/translate.py
```diff
@@ -397,6 +399,34 @@ def get_stack_var(self, location: int, *, store: bool) -> "Expression":
...
+                # Keep track of the type used at `location`
+                self.weak_stack_var_types[location] = field_type
+            elif location in self.weak_stack_var_types:
+                # This is a load from `location` we previously stored to
+                # Check if the `field_type` is compatible with the last stored type
+                if not self.weak_stack_var_types[location].unify(field_type):
+                    # The types weren't compatible. Mark this `location` as "weak" and use
+                    # the previous stored type instead of the one from `get_deref_field()`
+                    self.weak_stack_var_locations.add(location)
+                    field_type = self.weak_stack_var_types[location]
```
simonlindholm, line 429: (fwiw this line could be dedented without change in behavior, not sure if it makes the code any more understandable)

All this logic is kinda hard to follow, in part probably because the stored type isn't passed in for stores. Maybe worth refactoring? The detailed comments do help a lot.

simonlindholm (commented):
>  Otherwise, I'm not particularly excited about the idea of adding a new phi system to manage stack vars: that seems like a lot of complexity?

Agreed; it's a lot of complexity for little gain. It's also unclear if "phis but for types" can even be a thing. If we could find some phi-less middle ground though, like "inherit stack types from one of the parent nodes arbitrarily" (to at least avoid effects from siblings) or "inherit stack types from dominator" maybe that'd be an improvement?

The comments you added look great to me.

Still not super happy about the code here, but I'm totally fine approving this on the basis of "it improves output" if you don't want to make further changes.

src/translate.py
zbanks: Dedenting this actually helped refactor this whole block. Is this clearer to follow?

simonlindholm: Yeah, it's a good improvement. I do still feel the "in part probably because the stored type isn't passed in for stores" thing. It's weird that types are added to `weak_stack_var_locations` on the read/write *following* a write, so e.g. if you have `sp0 = x; sp0 = y;` (without a read afterwards) where `x` and `y` have different types, the "compiler-managed" comment wouldn't show up, and type unification is generally delayed by one step compared to what you'd naively expect. In practice it does work, of course.

simonlindholm (approved):
Re the discussion of phi nodes etc., I think theoretically what should happen is if our phi is read from, its type should be unified with all write types in predecessor blocks. Turning that into something more practical, we could have the stack type state be inherited from an *arbitrary* already-processed predecessor -- all predecessors ought to have the same types for stack vars that we read from, so it doesn't matter which one we pick. It does need dominator tree traversal order to be such that each block has at least one predecessor processed before it (except for the entry node), but I believe that's possible to arrange.

Up to you if you feel like implementing that; approving regardless.

src/translate.py
simonlindholm: (I wonder if `store: Optional[Type]` instead of `: bool` would be useful for other things too... probably not very)

zbanks: Ah, I misunderstood before. 
This does seem like a reasonable thing to change, and yeah `store: Optional[Type]` seems like a good way to implement this. (I think this will also help with structs on the stack, because you'll know store size in `get_stack_var`?)

zbanks: I'm taking a look at this today -- there's a complication with passing the `store` type in `make_store()`.
`make_store` currently does the following:
```python
    dest = deref(target, args.regs, stack_info, size=size, store=True)
    dest.type.unify(type)
    return StoreStmt(source=as_type(source_val, type, silent=is_stack), dest=dest)
```

This code is very sensitive to the order that the 3 types (`type`, `dest.type`, `source_val.type`) are unified.

If I change `store=True` to `store=type`, that only passes the "plain" type through, e.g. `Type.reg32()` -- it's useful for size, but it won't have any information about the pointer target.

If I put a `source_val.type.unify(type)` before the `deref(..., store=type)` call, that adds the information I need to `type`, but is too "early"?

The best I have so far is this, although it requires adding a `Type.copy()` method:
```python
    reg_type = type.copy()
    source_val.type.unify(type)
    dest = deref(target, args.regs, stack_info, size=size, store_type=type)
    dest.type.unify(reg_type)
    return StoreStmt(source=as_type(source_val, reg_type, silent=is_stack), dest=dest)
```

zbanks: If I make the change above, and move the type unification so it happens on the store instead of the following load, there are a lot of diffs in the output that look worse. 
Most of these diffs have look like the following:

```diff
- // 968fee6, with unification delayed until the following load
+ // a modified version with unification on the store

 void EnGeg_Init(Actor *thisx, GlobalContext *globalCtx) {
     ? sp34;
-    ColliderCylinder *sp30;                         /* compiler-managed */
+    ColliderCylinder *sp30;
     s32 temp_v0;
     EnGeg *this = (EnGeg *) thisx;
 
     sp34.unk_0 = (s32) D_80BB405C.unk_0;
     sp34.unk_4 = (s32) D_80BB405C.unk_4;
     if ((gSaveContext.weekEventReg[0x3D] & 1) != 0) {
         Actor_MarkForDeath(&this->actor);
         return;
     }
     ActorShape_Init(&this->actor.shape, 0.0f, func_800B3FC0, 0.0f);
     SkelAnime_InitSV(globalCtx, &this->unk_144, &D_06011AC8, &D_06012DE0, &this->unk_24C, &this->unk_2B8, 0x12);
     sp30 = &this->unk_18C;
     Collider_InitCylinder(globalCtx, &this->unk_18C);
     Collider_SetCylinder(globalCtx, &this->unk_18C, &this->actor, &D_80BB3E9C);
     sp30 = &this->unk_1D8;
-    Collider_InitSphere(globalCtx, &this->unk_1D8);
-    Collider_SetSphere(globalCtx, &this->unk_1D8, &this->actor, &D_80BB3E70);
+    Collider_InitSphere(globalCtx, (ColliderSphere *) &this->unk_1D8);
+    Collider_SetSphere(globalCtx, (ColliderSphere *) &this->unk_1D8, &this->actor, &D_80BB3E70);
     CollisionCheck_SetInfo2(&this->actor.colChkInfo, &D_80BB3ED4, &D_80BB3EC8);
     if (this->actor.update != 0) {
         temp_v0 = Object_GetIndex(&globalCtx->objectCtx, 0xA1);
         this->unk_248 = temp_v0;
         if (temp_v0 < 0) {
             Actor_MarkForDeath(&this->actor);
         }
     }
     Effect_Add(globalCtx, &this->unk_4DC, 4, 0U, (u8) 0, &sp34);
     this->actor.draw = NULL;
     this->unk_4E0 = 0x64;
     this->actor.draw = EnGeg_Draw;
     this->actionFunc = func_80BB217C;
 }
```

In this example, the types of `this->unk_1D8` and `this->unk_1C8` aren't known ahead of time. 
This is a clear example where `sp30` is a compiler-managed variable. In the second store, `sp30 = &this->unk_1D8;`, it's better to assume that `sp30` has an unknown type until more type information for the variable is available. (The load of `sp30` comes after the `Collider_InitSphere` call and before the `Collider_SetSphere` call.)

simonlindholm: Interesting. Let's keep the existing logic then.

zbanks: > It's weird that types are added to `weak_stack_var_locations` on the read/write _following_ a write, so e.g. if you have `sp0 = x; sp0 = y;` (without a read afterwards) where `x` and `y` have different types, the "compiler-managed" comment wouldn't show up, and type unification is generally delayed by one step compared to what you'd naively expect. 

I do want to clarify that in my refactors in `75951fb` then `1f8d587` slightly improve this situation. 
Now, types are added to `weak_stack_var_locations` on the _access_ following a write, regardless if it's a read or write. I also added a final pass in `1f8d587` to check the last write, in case there are no more accesses after the final write. 

This handles the `sp0 = x; sp0 = y;` case. It is still weird that the check is deferred, but it is more consistent.

PR #173, opened 2021-11-10
Attempt to fix some number formatting

Kelebek1:
Right now, literals are not properly signed in Python. Values are currently only read as uint, so a 16-bit value of 0xFFFF is only checked as 65535 and not the intended -1, so decimal printing is skipped. This PR attempts to fix it by skipping the elide when the `--no-casts` is already set, which preserves type information into printing.

This PR also adds some spacing between braces for data, `{0x600DE68, 1.0f, 0, -1, 2, 0},` -> `{ 0x600DE68, 1.0f, 0, -1, 2, 0 },`.


src/translate.py
```diff
@@ -1320,6 +1320,7 @@ def format(self, fmt: Formatter) -> str:
 class Literal(Expression):
     value: int
     type: Type = field(compare=False, default_factory=Type.any)
+    elide_cast: bool = False
```
zbanks, line 1323:
```suggestion
    elide_cast: bool = field(compare=False, default=False)
```

This way `Literal(1, elide_cast=True) == Literal(1, elide_cast=False)`, [ref](https://docs.python.org/3/library/dataclasses.html).

src/translate.py
```diff
@@ -1335,15 +1336,21 @@ def format(self, fmt: Formatter, force_dec: bool = False) -> str:
...
+
         if force_dec:
             value = str(self.value)
         else:
-            value = fmt.format_int(self.value)
+            value = fmt.format_int(
+                self.value,
+                str(self.type),
+                self.type.is_signed(),
+                self.type.get_size_bits(),
```
zbanks, line 1352: I think you can use just `self.type.is_signed()` & `self.type.get_size_bits()` -- I don't think you also need to check `str(self.type)`? 

I think it would also be better to move the negative number logic into this function. I don't think `format_int()` needs to change at all?

src/translate.py
```diff
@@ -2171,6 +2178,8 @@ def elide_casts_for_store(expr: Expression) -> Expression:
         return elide_casts_for_store(uw_expr.expr)
     if isinstance(uw_expr, Literal) and uw_expr.type.is_int():
         # Avoid suffixes for unsigned ints
+        if uw_expr.elide_cast:
+            return Literal(uw_expr.value, type=expr.type, elide_cast=True)
         return Literal(uw_expr.value, type=Type.intish())
```
zbanks, line 2183: Generally (always?), `late_unwrap(expr).type` is equivalent to `expr.type`. 
```suggestion
        if uw_expr.elide_cast:
            return expr
        return Literal(uw_expr.value, type=Type.intish())
```
Alternatively, this change would make it so you don't need to set `elide_cast=True` in `GlobalInfo.initializer_for_symbol(...)`
```suggestion
        return replace(uw_expr, elide_cast=True)
```

zbanks (changes requested):

src/translate.py
Kelebek1: I thought it would have to go into format_int, since the change is specifically about formatting ints. Also seemed like I'd just be duplicating code otherwise.

src/translate.py
Kelebek1: With the replace, I do still need the elide in `initializer_for_symbol` otherwise casts still get included when not using `--no-casts`

```
ActorAnimationEntryS D_80AD57E0[0xA] = {
    { (AnimationHeader* )0x600DE68, 1.0f, 0, -1, 2, 0 },
    { (AnimationHeader* )0x600EE98, 1.0f, 0, -1, 2, 0 },
    { (AnimationHeader* )0x600FD34, 1.0f, 0, -1, 0, 0 },
    { (AnimationHeader* )0x60030EC, 1.0f, 0, -1, 2, 0 },
    { (AnimationHeader* )0x6003D78, 1.0f, 0, -1, 2, 0 },
    { (AnimationHeader* )0x600D52C, 1.0f, 0, -1, 0, 0 },
    { (AnimationHeader* )0x6000A44, 1.0f, 0, -1, 0, 0 },
    { (AnimationHeader* )0x6001EF4, 1.0f, 0, -1, 0, 0 },
    { (AnimationHeader* )0x6002224, 1.0f, 0, -1, 0, 0 },
    { (AnimationHeader* )0x6002CB0, 1.0f, 0, -1, 0, 0 },
};```

Comment (simonlindholm) [reply src/translate.py]: That output is exactly what I'd expect without --no-casts. Int->pointer casts are sketchy and should be pointed out, not least to suppress compiler warnings.

File: src/translate.py (hunk 2 above)
Comment (zbanks) [reply src/translate.py]: I think the role of `format_int` (and `format_hex`) is to convert a bare Python number to a string. `Literal.format()`s is converting a typed C literal: it's responsible for looking at the type information and deciding how to use it. For example, adding the `U` suffix is done in `Literal.format` and not as an argument to `format_int`.
I think doing it in `Literal.format()` will also keep the size of the change much smaller: `format_int()` won't need to change at all, and you won't need to pass a handful of boolean arguments around.

As for s8/s16/s32/s64, you can check `type.is_int() and type.is_signed()` instead of doing string comparison.

File: src/translate.py (hunk 3 above)
Comment (Kelebek1) [reply src/translate.py]: This is specifically for data decomp with known structs. Should literals still have a cast when using directly within known data? Code itself isn't affected at all, no casts are lost there, this is only for data.

Comment (simonlindholm) [reply src/translate.py]: Yes, even for data the cast is sketchy (indicates that either the type is wrong, or it should be passing a symbol reference rather than a literal, e.g. via undefined_syms.txt), and there will be compiler warnings if you don't cast

File: tests/end_to_end/global_decls/irix-o2-allglobals-out.c [hunk 4]
```diff
@@ -7,22 +7,30 @@ MIPS2C_UNK static_fn(struct A *);                   /* static */
...
+s32 static_array[3] = { 2, 4, 6 };
 struct A static_bss_A;
 s32 static_bss_array[3];
 s32 static_int;
-s32 unused_static_array[3] = {8, 0xA, 0xC};
-s32 static_ro_array[3] = {7, 8, 9};                 /* const */
-s32 unused_static_ro_array[3] = {0xA, 0xB, 0xC};    /* const */
+s32 unused_static_array[3] = { 8, 0xA, 0xC };
+s32 static_ro_array[3] = { 7, 8, 9 };               /* const */
+s32 unused_static_ro_array[3] = { 0xA, 0xB, 0xC };  /* const */
```
simonlindholm, line 33: would be nice to change the test to have one of these arrays be s16, one s32, one u32, with one negative member in each. (involves changing the orig.c file and also the .s files manually)

src/translate.py
```diff
@@ -4442,7 +4453,7 @@ def for_type(type: Type) -> Optional[str]:
 
                 value = read_uint(size)
                 if value is not None:
-                    expr = as_type(Literal(value), type, True)
+                    expr = as_type(Literal(value, elide_cast=True), type, True)
```
simonlindholm, line 4456: is this elide_cast=True needed? I feel like elide_casts_for_store ought to cover it

src/translate.py
```diff
@@ -1335,15 +1336,25 @@ def format(self, fmt: Formatter, force_dec: bool = False) -> str:
...
+
         if force_dec:
             value = str(self.value)
         else:
-            value = fmt.format_int(self.value)
+            size_bits = self.type.get_size_bits()
+            v = self.value
+            if (
+                self.type.is_signed()
+                and v > 0
```
simonlindholm, line 1352: `and v < 2**size_bits` for safety (if it's out of range something has gone badly wrong, but we don't want to make it worse by changing literals in valid ways)

src/translate.py
```diff
@@ -1335,15 +1336,25 @@ def format(self, fmt: Formatter, force_dec: bool = False) -> str:
...
+                and size_bits
+                and v & (1 << (size_bits - 1))
+            ):
+                v = self.value - (1 << size_bits)
```
simonlindholm, line 1356:
```suggestion
                v -= 1 << size_bits
```

simonlindholm (approved):

tests/end_to_end/global_decls/irix-o2-allglobals-out.c
Kelebek1: Could you check if I did this correctly?

simonlindholm: Looks good to me.

PR #174, opened 2021-11-11
Only declare necessary functions

Kelebek1:
This PR seeks to only print functions which require a forward declaration when printing the global declarations. There's probably a better way to do this and I'm all ears if there's a simple way, but this was what I found.

There are some mypy errors with this, but they can't happen and I don't know how to surpress them.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -131,7 +131,7 @@ def run(options: Options) -> int:
             except:
                 pass
         try:
-            global_info.global_decls(fmt, options.global_decls)
+            global_info.global_decls(fmt, options.global_decls, [])
         except:
             pass
         for info in preliminary_infos:
@@ -166,7 +166,11 @@ def run(options: Options) -> int:
             if type_decls:
                 print(type_decls)
 
-        global_decls = global_info.global_decls(fmt, options.global_decls)
+        global_decls = global_info.global_decls(
+            fmt,
+            options.global_decls,
+            [fn for fn in function_infos if isinstance(fn, FunctionInfo)],
+        )
         if global_decls:
             print(global_decls)
     except Exception as e:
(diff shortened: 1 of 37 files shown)


src/main.py
```diff
@@ -168,7 +169,50 @@ def run(options: Options) -> int:
...
+                        elif isinstance(arg, Macro) and isinstance(
+                            arg.argument, AsmGlobalSymbol
+                        ):
+                            func_name = arg.argument.symbol_name
+                        else:
+                            continue
+
+                        if func_name in global_info.local_functions:
+                            if func_name not in funcs_seen:
+                                funcs_need_declare.add(func_name)
```
simonlindholm, line 198: Can you split the get-symbols-from-function code out into a helper function? `run` is long enough as it is already

src/main.py
```diff
@@ -168,7 +169,50 @@ def run(options: Options) -> int:
...
+
+            new_decls = []
+            for decl in global_decls.splitlines():
+                if "(" in decl and (");") in decl:
```
simonlindholm, line 202: String parsing is not great, and will even fail when function pointers get involved. Can we put this filtering inside global_decls instead?

simonlindholm (commented):
Looks good to me modulo some minor comments; I'll let @zbanks merge since this is more of his area of the code

src/translate.py
```diff
@@ -4582,6 +4623,13 @@ def global_decls(self, fmt: Formatter, decls: Options.GlobalDeclsEnum) -> str:
                 if decls != Options.GlobalDeclsEnum.ALL and sym.initializer_in_typemap:
                     continue
 
+                if (
+                    sym.type.is_function()
+                    and name in self.local_functions
+                    and name not in forward_declares_needed
```
zbanks, line 4629: Can you add `and decls != Options.GlobalDeclsEnum.ALL`, so that the filtering is disabled with `--globals=all`?

zbanks (approved):

PR #178, opened 2021-11-17
Prune overlapping struct fields in preliminary passes

zbanks:
This PR adds a pass that cleans up inferred fields in structs. This pass can change the paths used in `StructAccess`/`LocalVar` expressions, so it's only run when there are multiple passes. 

This primarily solves two problems:

1. Overlapping fields which are actually subfields of an earlier field. 
    - In `Vec3f unk0; f32 unk4; f32 unk8;`, `unk4` is really `unk0.y` and can be deleted.
    - This is an issue I had wanted to fix before allowing stack vars to be structs.
2. Incorrectly inferring large structs. 
    - It was really common to incorrectly infer `Actor unk_144;` in MM structs. This was brought up in #172. 
    - I think most of these errors are caused by missing arguments in function calls, but they're relatively easy to fix while fixing 1. This PR is more of a band-aid for these types of errors.

The strategy I use is a bit odd though, and I'm looking for feedback here. It depends on the new `--passes` feature and does the processing post-hoc (instead of incrementally). 
Is this a bad idea? My added code is relatively simple, but is it making `mips_to_c` too complicated? Is it weird to rely on performing the translation passes & discarding the output? I'm OK completely punting this PR.

[Diff in MM](https://gist.github.com/zbanks/5362785fe83253b50d86e27f71e84d88) & the [remaining overlapping fields](https://gist.github.com/zbanks/5362785fe83253b50d86e27f71e84d88#file-remaining_overlaps_mm). PM had no diffs (it doesn't have a lot of structs), and OOT only had two small diffs.
`En_Horse` in MM still outputs overlaps with the default `--passes=2`, and these don't go away until `--passes=4`.

### DIFF
--- a/src/types.py
+++ b/src/types.py
@@ -105,6 +105,11 @@ class TypePool:
                 decls.append(struct.format(fmt) + "\n")
         return "\n".join(decls)
 
+    def prune_structs(self) -> None:
+        """Remove overlapping fields from all known structs"""
+        for struct in self.structs:
+            struct.prune_overlapping_fields()
+
 
 @dataclass(eq=False)
 class TypeData:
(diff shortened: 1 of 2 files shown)


src/types.py
```diff
@@ -1049,6 +1054,61 @@ def try_add_field(
...
+                    offset, target_size=f2.type.get_size_bytes()
+                )
+                if sub_path is not None and sub_type.unify(f2.type):
+                    # Remove `f2` so that accesses to `f2.offset` go through `field` instead.
+                    # Defer adding `f2` to `fields_to_remove` until we're done processing `field`.
+                    conflicting_fields.add(f2)
+                elif not f2.known and f2.type.is_int() and field.type.is_int():
+                    # Ignore overlapping inferred ints.
+                    # Usually the code is easier to manually fix if we leave the overlap
+                    pass
```
zbanks, line 1097: I could remove this `elif`, there are just 2 cases in MM where it matters (`audio_playback` and `pfsdeletefile`). It's not super clear to me if it appreciably helps, or is worth the complexity?

src/types.py
```diff
@@ -1049,6 +1054,61 @@ def try_add_field(
...
+                else:
+                    # `field` is too big: this usually means that we incorrectly inferred
+                    # it to be a (large) struct. Reset its type to something smaller.
+                    # Although this is unlikely to produce the correct type for this field,
+                    # it helps minimize the damage done to the rest of the struct.
+                    if offset >= 4:
+                        field.type = Type.any_reg()
+                    else:
+                        field.type = Type.int_of_size(offset * 8)
+                    conflicting_fields = set()
```
simonlindholm, line 1107: the fields in conflicting_fields still overlap with this one, do we need to do anything about that?

src/types.py
simonlindholm: Seems fine, worth extending the comment to explain that this is because of load+cast being converted to load of smaller size. Eventually it would be nice to start detecting such loads, I assume this code can go away at that point

simonlindholm (approved):
I think this makes sense. Not an ideal design ofc, but afaict impossible to do single-pass, so...

PR #179, opened 2021-11-19
Use `&x + scale * index` expressions to determine `x`'s type

zbanks:
This is a relatively small change, but I think it improves the output a lot (at least on PM/MM). As I mentioned on discord, I know I have a lot of PRs open right now, but this change seems more useful than the others?

The "weird" part, IMO, is where I create a new struct when `x` is a `GlobalSymbol` and `scale` isn't 1, 2, or 4. This has the downside of potentially being wrong/noisy, because the new struct can't then unify with an existing struct. However, I think it has a net positive on the output? And it hopefully it's easy for the decomper to look at, and quickly identify "these look like the same struct" or "this looks like a `Vec3s`".

The output also has a lot more expressions like `(&sp30)[...]`, because stack variables are not yet allowed to be arrays. The element type is usually correct, but it requires the decomper to manually turn the variable into an array.

This change also seems to help reduce the number of temps, which is always a nice bonus.

The diffs highlight the fact that some "symbols" in MM are written in [`gSaveContext + 0x100` form](https://gist.github.com/zbanks/1ae3c38d1f9a7fd2dffe155fd131bbef#file-mm-diff-L5321), which mips2c assumes is the whole symbol name. [I have a commit to parse these into `StructAccess`es](https://github.com/matt-kempster/mips_to_c/commit/3bf77f6272f4a012ba894a846c03c8505399e2f6) but I'm not sure if I should include that in this PR.

The [diffs are pretty big](https://gist.github.com/zbanks/1ae3c38d1f9a7fd2dffe155fd131bbef), esp. on MM. There were changes on 37% of MM files and 5% of PM files. The diffs are large because there are many new global initializers added. MM's `ovl_Player_Actor` again posed some trouble, I think due to the fn arg issue.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -3002,6 +3002,8 @@ def array_access_from_add(
             inner_type = Type.reg32(likely_float=False)
         elif typepool.struct_field_inference and isinstance(uw_base.expr, GlobalSymbol):
             # Make up a struct with a tag name based on the symbol & struct size.
+            # Although `scale = 8` could indicate an array of longs/doubles, it seems more
+            # common to be an array of structs.
             struct_name = f"_struct_{uw_base.expr.symbol_name}_0x{scale:X}"
             struct = typepool.get_struct_by_tag_name(
                 struct_name, stack_info.global_info.typemap
--- a/src/translate.py
+++ b/src/translate.py
@@ -1315,6 +1315,39 @@ class GlobalSymbol(Expression):
     def format(self, fmt: Formatter) -> str:
         return self.symbol_name
 
+    def potential_array_dim(self, element_size: int) -> Tuple[int, int]:
+        """
+        Using the size of the symbol's `asm_data_entry` and a potential array element
+        size, return the corresponding array dimension and number of "extra" bytes left
+        at the end of the symbol's data.
+        If the extra bytes is nonzero, it's likely that `element_size` is incorrect.
+        """
+        # If we don't have the .data/.rodata entry for this symbol, we can't guess
+        # its array dimension. Jump tables are ignored and not treated as arrays.
+        if self.asm_data_entry is None or self.asm_data_entry.is_jtbl:
+            return 0, element_size
+
+        min_data_size, max_data_size = self.asm_data_entry.size_range_bytes()
+        if element_size > max_data_size:
+            # The type is too big for the data (not an array)
+            return 0, max_data_size
+
+        # Check if it's possible that this symbol is not an array, and is just 1 element
+        if min_data_size <= element_size <= max_data_size and not self.type.is_array():
+            return 1, 0
+
+        array_dim, extra_bytes = divmod(min_data_size, element_size)
+        if extra_bytes != 0:
+            # If it's not possible to make an exact multiple of element_size by incorporating
+            # bytes from the padding, then indicate that in the return value.
+            padding_bytes = element_size - extra_bytes
+            if min_data_size + padding_bytes > max_data_size:
+                return array_dim, extra_bytes
+
+        # Include potential padding in the array. Although this is unlikely to match the original C,
+        # it's much easier to manually remove all or some of these elements than to add them back in.
+        return max_data_size // element_size, 0
+
 
 @dataclass(frozen=True, eq=True)
 class Literal(Expression):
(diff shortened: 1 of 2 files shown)


src/translate.py
```diff
@@ -4582,38 +4665,28 @@ def global_decls(
                     # The size of the element type (not the size of the array type)
                     if element_type is None:
                         element_type = sym.type
+
+                    # If we don't know the type, we can't guess the array_dim
                     type_size = element_type.get_size_bytes()
-                    if not type_size:
```
zbanks, line 4586: The changes in this section have almost effect on the output. The only change is that the sizes in the warning comments now use `format_int`, e.g. `type too large by 0x30`. (I verified this using the MM tests with the changes in `array_access_from_add` commented out.)

This change was for the refactor of `GlobalSymbol.potential_array_dim`, so `array_access_for_add` could use the same array heuristic. (I think also makes this logic a bit easier to follow?)

src/translate.py
```diff
@@ -2934,10 +2967,60 @@ def array_access_from_add(
...
+                struct_name, stack_info.global_info.typemap
+            )
+            if struct is None:
+                struct = StructDeclaration.unknown_of_size(
+                    typepool, size=scale, tag_name=struct_name
+                )
+            elif struct.size != scale:
+                # This should only happen if there was already a struct with this name in the context
+                raise DecompFailure("sizeof(struct {struct_name}) != {scale:#x}")
+            inner_type = Type.struct(struct)
```
zbanks, line 3016: This is the weird part mentioned in the description. I could take it out if it's *too* weird? It checks `typepool.struct_field_inference` so that it's only enabled when `--structs` is turned on.

src/translate.py
```diff
@@ -4582,38 +4667,28 @@ def global_decls(
...
-                                comments.append(f"extra bytes: {data_size % type_size}")
-                        if data_size // type_size > 1 or is_vla:
-                            # We know it's an array
-                            array_dim = max_data_size // type_size
+                    if type_size:
+                        potential_dim, extra_bytes = sym.potential_array_dim(type_size)
+                        if potential_dim == 0 and extra_bytes > 0:
+                            # The type is too big for our data. (not an array)
+                            comments.append(
+                                f"type too large by {fmt.format_int(type_size - extra_bytes)}"
```
simonlindholm, line 4678:
```suggestion
                                f"type too large by {fmt.format_int(extra_bytes)}"
```

src/translate.py
```diff
@@ -1315,6 +1315,39 @@ def format_string_constant(self, fmt: Formatter) -> str:
     def format(self, fmt: Formatter) -> str:
         return self.symbol_name
 
+    def potential_array_dim(self, element_size: int) -> Tuple[int, int]:
+        """
+        Using the size of the symbol's `asm_data_entry` and a potential array element
+        size, return the corresponding array dimension and number of "extra" bytes left
+        at the end of the symbol's data.
+        If the extra bytes is nonzero, it's likely that `element_size` is incorrect.
```
simonlindholm, line 1323:
```suggestion
        If the extra bytes are nonzero, it's likely that `element_size` is incorrect.
```

src/translate.py
```diff
@@ -2934,10 +2967,62 @@ def array_access_from_add(
         index = addend
         scale = 1
 
+    if scale < 0:
+        scale = -scale
+        index = UnaryOp("-", index, type=index.type)
```
simonlindholm, line 2972: (technically type=index.type can be incorrectly, but it doesn't matter ofc)

src/translate.py
```diff
@@ -2934,10 +2967,62 @@ def array_access_from_add(
...
+            struct = typepool.get_struct_by_tag_name(
+                struct_name, stack_info.global_info.typemap
+            )
+            if struct is None:
+                struct = StructDeclaration.unknown_of_size(
+                    typepool, size=scale, tag_name=struct_name
+                )
+            elif struct.size != scale:
+                # This should only happen if there was already a struct with this name in the context
+                raise DecompFailure("sizeof(struct {struct_name}) != {scale:#x}")
```
simonlindholm, line 3017:
```suggestion
                raise DecompFailure(f"sizeof(struct {struct_name}) != {scale:#x}")
```

simonlindholm (approved):

src/translate.py
zbanks: The original was right, though the refactor is confusing. I checked the diff to be sure:

```diff
-static GlobalContext D_801C0EAC;                    /* type too large by 102988; unable to generate initializer */
+static GlobalContext D_801C0EAC;                    /* type too large by 0x1924C; unable to generate initializer */
-static Player D_801C0EB8;                           /* type too large by 3440; unable to generate initializer */
+static Player D_801C0EB8;                           /* type too large by 0xD70; unable to generate initializer */
-static CollisionHeader D_80AB1690;                  /* type too large by 20; unable to generate initializer */
+static CollisionHeader D_80AB1690;                  /* type too large by 0x14; unable to generate initializer */
-static Actor D_8085D5E0;                            /* type too large by 320; unable to generate initializer */
+static Actor D_8085D5E0;                            /* type too large by 0x140; unable to generate initializer */
```

The return value from `potential_array_dim` is analogous to the return of `divmod`. `extra_bytes` represents the number of *unused* bytes: if `potential_dim  == 0`, then the data was only `extra_bytes` large but the type is `type_size`.

PR #180, opened 2021-11-27
Detect inferred switches from nested ifs/switches

zbanks:
- Added a flag `--no-switches` to disable the new behavior.
- Added a test variant to `multi-switch` that runs w/ `--no-switches`. The test output matches the original, before these changes.
- When implicit switches are detected, there is a comment in the output to help the decomper distinguish them from true jumptable switches. 
    - Maybe this comment could be removed once we have confidence there are few/no false positives?
- IDO is a bit more consistent with what kinds of comparisons it will generate, compared to GCC. 
    - Behavior is controlled by the `--compiler` flag.
- `switch_index` values are all screwed up, now that we can both elide or add switches during codegen. 
    - I've tweaked it so that it never uses the same index twice, but it's possible to have 1 switch have index `0` (which disables comments), and have a second switch have index `1` (so only one switch/labels is commented). 
    - It's also somewhat common for the output to have noncontinuous `switch_index`s.
- It looks like GCC may be able to generate some conditions using a pattern like: `if (x >= 5)` / `if (x == 6)` / `if (x < 6)`: all 3 of these conditions are true iff `x == 5`, but the `if (x == 5)` check is implicit.
    - This shows up in functions like PM's `func_802403D4_97BA04` (in what would be the `temp_v1 == 0xD` case). I think it's compiler-generated, but I'm not 100% sure since the fn hasn't been decomp'd.
- I don't like that I have to use `early_unwrap_ints` in `if_statements.py`. I originally tried to factor out the cast in `BinaryOp.scmp(...)`, but I don't think that's the most robust approach either. 
- Diffs (although they're quite hard to read/interpret):
    - [OOT `a871499`](https://gist.github.com/zbanks/db8138ec2ad5a46ba3747ab52ec67e64); ~[OOT `cf61f36`](https://gist.github.com/zbanks/269cc748ea525cc7c1dfb6b979b560ec)~
    - ~[MM `cf61f36`](https://gist.github.com/zbanks/56bda38b098f8d2778d3f971351eabce)~
    - ~[PM `cf61f36`](https://gist.github.com/zbanks/d2e8860c6bfe6b4acf26f16a6a13e0bf)~
    - Overall, from what I can tell, there are fewer `goto`s and levels of nesting. I've spot-checked a bunch of decomp'd OOT functions for false positives/negatives.
    -  The change to `is_empty_goto()` and its use in `build_flowgraph_between()` in `1b0c532` introduced a lot of diffs where extra `return;` statements & `/* Duplicate return node #N. Try simplifying control flow for better match */`  comments were elided.


src/if_statements.py
```diff
@@ -570,6 +602,213 @@ def try_make_if_condition(
...
+    If one is found, return the corresponding SwitchStatement; otherwise, return None.
+
+    Both IDO & GCC can convert switch statements into a tree of comparisons & jump tables.
+    We try to identify the largest contiguous tree of ConditionalNodes & SwitchNodes that
+    all compare the same variable (`var_expr`). SwitchNodes and `(x == N)` ConditionalNodes
+    represent jumps to specific case labels, whereas comparisons like `(x < N)` are
+    primarily used to manage the overall tree depth.
+    """
+    # The start node must either be an if or a switch
+    if not isinstance(start, (ConditionalNode, SwitchNode)):
```
simonlindholm, line 619: if we have a SwitchNode it would also be fine to return and let build_switch_between deal with this, right? since there's no bounds check so we can't have a multi-switch. (no opinion yet on whether it's a good idea)

src/if_statements.py
```diff
@@ -570,6 +602,213 @@ def try_make_if_condition(
...
+    # Nodes that have no statements, and should be marked as emitted if we emit a SwitchStatement.
+    # A ConditionalNode like `if (x == N)` isn't directly emitted; it's replaced by a `case N:` label
+    # This also includes empty BasicNodes & ReturnNodes that jump directly to the end
+    nodes_to_mark_emitted: Set[Node] = set()
+    # Map of label -> node. Similar to SwitchNode.cases, but the labels may not be contiguous
+    cases: Dict[int, Node] = {}
+    # The `default:`-labeled node, if found
+    default_node: Optional[Node] = None
+    # Require at least one ConditionalNode that is not a guard for a SwitchNode
+    has_non_guard_conditional = False
```
simonlindholm, line 662: Can't we have a switch that's split into two jump tables and no other conditional nodes?

simonlindholm (commented):
Started reviewing this, still have most of try_build_implicit_switch to go.

flow_graph has a helpful warning which you can remove now:
```
                    "\n"
                    "(You might need to pass --goto and --no-andor flags as well, "
                    "to get correct control flow for non-jtbl switch jumps.)"
```

src/if_statements.py
```diff
@@ -786,6 +1039,10 @@ def build_flowgraph_between(
                 curr_start = imm_pdom
                 continue
 
+        if curr_start in context.emitted_nodes and is_empty_goto(curr_start, end):
```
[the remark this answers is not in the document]
zbanks: Yeah I should explain this more. It shouldn't result in any missing `emit_gotos` because `curr_start` is already emitted.

These irregular switches end up with a lot of empty `BasicNode`s which just act as gotos to `end`. If we've hit one, we don't need to emit any more goto/return statements: the return value of `build_flowgraph_between` will be terminated with a `}` or `break;`. 

Example function w/ this check removed, where the end of each case block is a `goto block_41;` and all that block does is `break;`:
```diff
[INFO] Output of ../oot/asm/non_matchings/overlays/actors/ovl_Boss_Ganon/func_808E2544.s changed! Diff:
--- 
+++ 
@@ -152,13 +152,14 @@

             arg0->unk_1C2 = 1;
             arg0->speedXZ = 0.0f;
         }
+block_41:
         break;
     case 1:
         Math_ApproachZeroF(arg0 + 0x1CC, 1.0f, 40.0f);
         if (arg0->unk_1CC == 0.0f) {
             Actor_Kill(arg0);
         }
-        break;
+        goto block_41;
     case 10:
         arg0->unk_1C2 = 0xB;
         arg0->unk_1B6 = 0xE;
@@ -252,7 +253,7 @@

                 }
             }
         }
-        break;
+        goto block_41;
     case 12:
         arg0->speedXZ = 20.0f;
         temp_f12_5 = arg0->unk_1F0 - arg0->world.pos.x;
@@ -283,7 +284,7 @@

             arg0->unk_1C2 = 1;
             arg0->speedXZ = 0.0f;
         }
-        break;
+        goto block_41;
     }
     if ((s32) arg0->unk_1C2 >= 0xB) {
         temp_f12_7 = arg0->world.pos.x;
```

Example function w/ this check removed that doesn't have a switch statement, but still benefits from this change
```diff
[INFO] Output of ../oot/asm/non_matchings/code/ucode_disas/UCodeDisas_Disassemble.s changed! Diff:
--- 
+++ 
@@ -1563,9 +1563,11 @@

                 if (sp370.unk_2 == 0) {
                     if (temp_t0_12 != 0) {
                         osSyncPrintf("count_gsDPNoOpQuiet(),", temp_a1_3);
+                        goto block_253;
                     }
                 } else if (temp_t0_12 != 0) {
                     osSyncPrintf("count_gsDPNoOpVerbose(),", temp_a1_3);
+block_253:
                 }
                 arg0->enableLog = (s32) sp370.unk_2;
                 phi_t0_5 = (s32) sp370.unk_2;
```

src/if_statements.py
zbanks: Hmm, I don't know if the compiler could/would do that? But you might be right...

The intent of this check was to avoid handling the case of just a single SwitchNode & it's ConditionalNode guard, because detecting the default node in this function was less accurate than the existing logic. 

I'm going to work on this some more to see if I can make the default node matching better. It would be good to get that working, so that the logic in `build_flowgraph_between` (`elif switch_guard_expr(curr_start) is not None: ...`) could be removed in the future?

simonlindholm: Apparently IDO doesn't:
```c
void foo(void) {
    switch (x) {
    case 0:
    case 1:
    case 2:
    case 3:
    case 4:
    case 100:
    case 101:
    case 102:
    case 103:
    case 104:
        bar();
    }
}
```
results in
```
00000000 <foo>:
   0:	27bdffd8 	addiu	sp,sp,-40
   4:	afb00018 	sw	s0,24(sp)
   8:	3c100000 	lui	s0,0x0
			8: R_MIPS_HI16	x
   c:	8e100000 	lw	s0,0(s0)
			c: R_MIPS_LO16	x
  10:	afbf001c 	sw	ra,28(sp)
  14:	2a010005 	slti	at,s0,5
  18:	1420000b 	bnez	at,48 <foo+0x48>
  1c:	00000000 	nop
  20:	260eff9c 	addiu	t6,s0,-100
  24:	2dc10005 	sltiu	at,t6,5
  28:	10200012 	beqz	at,74 <foo+0x74>
  2c:	00000000 	nop
  30:	000e7080 	sll	t6,t6,0x2
  34:	3c010000 	lui	at,0x0
			34: R_MIPS_HI16	.rodata
  38:	002e0821 	addu	at,at,t6
  3c:	8c2e0000 	lw	t6,0(at)
			3c: R_MIPS_LO16	.rodata
  40:	01c00008 	jr	t6
  44:	00000000 	nop
  48:	2e010005 	sltiu	at,s0,5
  4c:	10200009 	beqz	at,74 <foo+0x74>
  50:	00000000 	nop
  54:	00107880 	sll	t7,s0,0x2
  58:	3c010000 	lui	at,0x0
			58: R_MIPS_HI16	.rodata
  5c:	002f0821 	addu	at,at,t7
  60:	8c2f0014 	lw	t7,20(at)
			60: R_MIPS_LO16	.rodata
  64:	01e00008 	jr	t7
  68:	00000000 	nop
  6c:	0c000000 	jal	0 <foo>
			6c: R_MIPS_26	bar
  70:	00000000 	nop
  74:	8fbf001c 	lw	ra,28(sp)
  78:	8fb00018 	lw	s0,24(sp)
  7c:	27bd0028 	addiu	sp,sp,40
  80:	03e00008 	jr	ra
  84:	00000000 	nop
```
i.e.
```
if (x < 5) {
  if (x < 5) jtbl;
} else {
  if (x - 100 < 5) jtbl;
}
```
which is a bit silly. GCC on the other hand doesn't believe in multiple jump tables: https://decomp.me/scratch/dJz3D

src/if_statements.py
```diff
@@ -570,6 +603,216 @@ def try_make_if_condition(
...
+    Look for irregular switch-like structures from nested ConditionalNodes & SwitchNodes.
+    If one is found, return the corresponding SwitchStatement; otherwise, return None.
+
+    Both IDO & GCC can convert switch statements into a tree of comparisons & jump tables.
+    We try to identify the largest contiguous tree of ConditionalNodes & SwitchNodes that
+    all compare the same variable (`var_expr`). SwitchNodes and `(x == N)` ConditionalNodes
+    represent jumps to specific case labels, whereas comparisons like `(x < N)` are
+    primarily used to manage the overall tree depth.
+    """
+    # The start node must either be an if or a switch
```
simonlindholm, line 619:
```suggestion
    # The start node must either be an if
```
and you can also remove the `start_block_info.switch_control is not None` case below

src/if_statements.py
```diff
@@ -570,6 +603,216 @@ def try_make_if_condition(
...
+    # Need to combine at least 3 nodes into 2 distinct cases, otherwise it could be a plain if/else with ||
+    if (
+        irregular_comparison_count < 2
+        or len(nodes_to_mark_emitted) < 3
+        or len(set(cases.values())) < 2
+    ):
+        return None
+
+    # If this new irregular switch uses all of the other switch nodes in the function,
+    # then we no longer need to add labelling comments with the switch_index
```
simonlindholm, line 786: (since the change to emit structured switches, are switch indices still actually useful?) Would be nice to clean this up so ordering is consecutive, but it seems like low priority.

src/if_statements.py
```diff
@@ -570,6 +603,216 @@ def try_make_if_condition(
...
+                # GCC will freely use other types of comparisons
+                # Examples from PM: func_8026E558, pr_load_npc_extra_anims
+                if cond.op == "==":
+                    if cond.right.value in cases:
+                        return None
+                    cases[cond.right.value] = node.conditional_edge
+                    node_queue.append(node.fallthrough_edge)
+                elif cond.op == "!=" and (
+                    node.block.index > node.conditional_edge.block.index
+                    or context.options.compiler == context.options.CompilerEnum.GCC
```
simonlindholm, line 727: do we end up with false positives if we remove this compiler check?

src/if_statements.py
```diff
@@ -570,6 +603,216 @@ def try_make_if_condition(
...
+                    cases[cond.right.value] = node.fallthrough_edge
+                    node_queue.append(node.conditional_edge)
+                elif cond.op == "<" or (
+                    cond.op in (">=", "<=", ">")
+                    and context.options.compiler == context.options.CompilerEnum.GCC
+                ):
+                    # TODO: GCC (maybe?) can emit a series of `<`/`>=` comparisons that limit
+                    # to a specific case. Ex: `x < 9 && x >= 8` is only true for `x == 8`.
+                    # Detecting these would require tracking & checking all of the bounds.
+                    # For now, ignore those, and assume the compiler isn't emitting nonsense.
```
simonlindholm, line 740: (if it emits e.g. `8 <= x < 16` to cover case 8..15: I wouldn't call that nonsense)

src/if_statements.py
```diff
@@ -570,6 +603,216 @@ def try_make_if_condition(
...
+                    node.block.index > node.conditional_edge.block.index
+                    or context.options.compiler == context.options.CompilerEnum.GCC
+                ):
+                    if cond.right.value in cases:
+                        return None
+                    cases[cond.right.value] = node.fallthrough_edge
+                    node_queue.append(node.conditional_edge)
+                elif cond.op == "<" or (
+                    cond.op in (">=", "<=", ">")
+                    and context.options.compiler == context.options.CompilerEnum.GCC
```
simonlindholm, line 735: same question

simonlindholm (approved):

src/if_statements.py
zbanks: (Just to be clear, the updated code with the `irregular_comparison_count` variable is more accepting of these types of things, and would accept a pair of jump tables with no other nodes. It doesn't seem to come up in practice, but it doesn't hurt)

src/if_statements.py
zbanks: There are still a few situations where the case labels get mangled a bit, primarily if the original source uses `goto`.

src/if_statements.py
zbanks: Ah yeah I should clarify. By "nonsense," I meant something structured like this:
```c
if (x > 10) {
    if (x == 5) {
        foo();
    } else if (x == 15) {
        bar();
    }
} else {
    if (x == 5) {
        baz();
    }
}
```
This would be equivalent to the following, because `foo()` is unreachable. But the current logic would not make a switch statement, because it looks like `x == 5` is repeated.
```c
switch (x) {
    case 5:  baz(); break;  // the `foo();` branch was unreachable
    case 15: bar(); break;
}
```

src/if_statements.py
zbanks: Yeah, an example is [this if/else chain from `z_vimode` in OOT](https://github.com/zeldaret/oot/blob/master/src/code/z_vimode.c#L81-L105) and [this if/else chain in `z_boss_ganon`](https://github.com/zeldaret/oot/blob/826e5b71d11a60ccfbbe1267fbb4c93f7f4b080a/src/overlays/actors/ovl_Boss_Ganon/z_boss_ganon.c#L4622-L4799).

src/if_statements.py
zbanks: Thanks for checking - it looks like this one isn't needed. 
There *might* be a false-positive in MM's `Room_AllocateAndLoad` when I remove this check, but the structure is *very weird*, (removing the other `== GCC` check doesn't help with the outer `if` either)
```diff
     if ((temp_v0_5 != 0) && (temp_v0_5 != -2) && (temp_v0_5 != -7)) {
-        if ((temp_v0_5 == -8) || (temp_v0_5 == -5) || (temp_v0_5 == -4) || ((temp_v0_5 < 0) && (temp_v0_5 != -1) && (temp_v0_5 != -6))) {
-
-        } else {
+        switch (temp_v0_5) {                        /* irregular */
+        case -8:
+        case -5:
+        case -4:
+            break;
+        default:
             phi_v1_3 = temp_v0_5 - 1;
             if (temp_v0_5 < 0) {
                 phi_v1_3 = 2;
             }
+            break;
         }
         phi_a2 = (s32) gSaveContext.respawn[phi_v1_3].roomIndex;
     } else {
         phi_a2 = (s32) ((globalCtx + 0x18000)->unk_85C + ((globalCtx + 0x18000)->unk_B48 * 2))->unk_1;
     }
```

PR #186, opened 2021-12-03
Move &&/CommaConditionExpr handling into Body.add_if_else 

zbanks:
This gives more opportunities for converting comma expressions in `&&` conditions into nested ifs.

The weirdest part about this PR is that we always build the complex `&&` conditional, and then *later* decompose it into separate parts. I think the iteration that splits the conditional apart got a bit messy, so I tried to compensate with comments & am happy to clarify further. (I have a mental image of a [zipper](https://en.wikipedia.org/wiki/Zipper_(data_structure)), but I'm not sure it's the best way to describe what I'm doing.)

The new test is based on the structure of MM's `func_8093BBD0` if statements, but it's changed to use `a && b && c` to ensure that the rewrites in `Body.add_if_else` don't swap any of the terms.

[Diffs for OOT + MM](https://gist.github.com/zbanks/80edd5dbe2ebfb207a66c2660e5d2959). 
I think the only downside of this change is when the comma conditions were fake, like the following:
```diff
-    if (((spD0 & 4) == 0) && (temp_t7 = temp_t3 > 0, (temp_t7 != 0)) && (phi_s1 = (u32) temp_t3, (temp_t7 != 0))) {
+    if ((spD0 & 4) == 0) {
+        temp_t7 = temp_t3 > 0;
+        if (temp_t7 != 0) {
+            phi_s1 = (u32) temp_t3;
+            if (temp_t7 != 0) {
```
But, I think these are pretty minor, and should still be reasonable for the decomper to fix up?

There's also a [weird diff in MM's `z_parameter.c` `func_801210E0`](https://gist.github.com/zbanks/80edd5dbe2ebfb207a66c2660e5d2959#file-oot_mm-diff-L3334-L3357), but I think this is from the irregular switch code being too aggressive. The code with `--gotos-only` makes the `if (temp_v0_6 != 0)` check look handwritten:
```c
// --gotos-only snippet from z_parameter's func_801210E0
    temp_v0_6 = phi_a3_3->unk280;
    if (temp_v0_6 == 0) {
        goto block_94;
    }
    if ((u32) (temp_v0_6 - 1) >= 0x15U) {
        goto block_94;
    }
    switch (temp_v0_6);                             /* switch 2; jump table: jtbl_801DDCA4 */
``

### CODE REVIEW

File: src/if_statements.py [hunk 1]
```diff
@@ -231,17 +231,73 @@ def add_statement(self, statement: Statement) -> None:
...
+        # into nested if's like `if (A1) { if (A2) { if (A3) { B; goto C; } } } D;`
+        # (This can't be done without additional gotos if there is an explicit `else` block)
+        # We perform this expansion if any of the `A1`, `A2`... subconditions are CommaConditionExprs,
+        # to avoid using the comma operator in conditions.
+
+        # Iterate through the outermost `&&` subconditions in `if_else.condition` like a zipper.
+        # We start with the whole condition in `outer_cond`, and iterate by appending the right
+        # side into `inner_cond`. At the start of each iteration, the whole condition would be
+        # represented by `join_conditions(outer_cond, "&&", inner_cond)` -- unless any conditions
+        # are removed.
```
simonlindholm, line 255: Would be useful to state here that outer_cond is originally structured like `(((...) && ...) && ...)` and gets converted into `(... && (... && (...)))`

src/if_statements.py
```diff
@@ -231,17 +231,73 @@ def add_statement(self, statement: Statement) -> None:
...
+    def add_if_else(self, if_else: IfElseStatement, else_jump: bool = False) -> None:
+        if if_else.else_body is not None and not if_else.if_body.ends_in_jump():
+            # Simple case; perform no further rewrites
+            self.statements.append(if_else)
             return
 
-        self.statements.append(if_else)
+        # We now know that we have an IfElseStatement like `if (A) { B; goto C; } else { D; }`
+        # where `D` may be empty. We can rewrite this into `if (A) { B; goto C; } D;`
+        # which reduces indentation to make the output more readable.
```
simonlindholm, line 242: This rewrite only happens at the very end of this function, which is pretty far down. Can you put the middle part in a function? It doesn't need access to the full if_else either, only to `.condition` and `.if_body`, and it may be nicer to have those as individual variables rather than incrementally replace()ing the if_else.

simonlindholm (approved):

src/if_statements.py
zbanks: I moved the bulk of this fn out to `rewrite_if_ands()`.  Is this sort of what you meant? 
I don't love the required precondition that can't be `assert`ed on (no `else_body` to check if it's empty) -- but moving that check into `rewrite_if_ands()` makes `Body.add_if_else()` practically empty, and loses a bit of that encapsulation.

src/if_statements.py
zbanks: Refactored to avoid this, since it seems to affect codegen on IDO.

simonlindholm (approved):

src/if_statements.py
simonlindholm: Yeah, your refactoring matches what I was thinking. IMO the fact that the helper doesn't take in an else block means it's clear from the caller that it cannot possibly emit an else, which makes the assert unnecessary. But true, I can see the argument about it potentially being error-prone.

src/if_statements.py
```diff
@@ -360,6 +309,60 @@ def format(self, fmt: Formatter) -> str:
...
+                new_body.add_statement(SimpleStatement(stmt))
+            new_body.add_if_else(
+                IfElseStatement(
+                    condition=inner_cond,
+                    if_body=if_body,
+                    else_body=None,
+                )
+            )
+            if_body = new_body
+            condition = outer_cond
```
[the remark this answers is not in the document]
zbanks: When I was initially writing this, it started as recursive (which was related to that `else_jump` var I accidentally left behind :sweat_smile: )

Kind of on the fence, but I don't think it'd be significantly clearer?

PR #191, opened 2021-12-12
Cherry-pick changes to parse_file.py from ppc2cpp branch + cleanup

zbanks:
- Cherry-pick'd `069f226` from `ppc2cpp`
    - Removed the `-I` shortarg for `--incbin-dir`, not sure if we really want to use `-I` for that.
- Refactored directive parsing in `parse_file`to avoid manual string slicing.
- Checked that the new `parse_file.py` can be used on the `ppc2cpp` branch without additional modifications.
- All tests still run, no diffs in OOT, MM, or PM.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -269,7 +269,7 @@ def parse_flags(flags: List[str]) -> Options:
         action="append",
         default=[],
         type=Path,
-        help="Search paths for loading .incbin directives in the input asm",
+        help="Add search path for loading .incbin directives in the input asm",
     )
 
     group = parser.add_argument_group("Output Options")
--- a/src/main.py
+++ b/src/main.py
@@ -263,6 +263,14 @@ def parse_flags(flags: List[str]) -> Options:
         default=[],
         help="Mark preprocessor constant as undefined",
     )
+    group.add_argument(
+        "-I",
+        "--incbin-dir",
+        dest="incbin_dir",
+        default=None,
+        type=Path,
+        help="Search path for loading .incbin directives",
+    )
 
     group = parser.add_argument_group("Output Options")
     group.add_argument(
@@ -554,6 +562,7 @@ def parse_flags(flags: List[str]) -> Options:
         print_stack_structs=args.print_stack_structs,
         unk_inference=args.unk_inference,
         passes=args.passes,
+        incbin_dir=args.incbin_dir,
     )
 
 
--- a/src/main.py
+++ b/src/main.py
@@ -265,10 +265,11 @@ def parse_flags(flags: List[str]) -> Options:
     )
     group.add_argument(
         "--incbin-dir",
-        dest="incbin_dir",
-        default=None,
+        dest="incbin_dirs",
+        action="append",
+        default=[],
         type=Path,
-        help="Search path for loading .incbin directives in the input asm",
+        help="Search paths for loading .incbin directives in the input asm",
     )
 
     group = parser.add_argument_group("Output Options")
@@ -561,7 +562,7 @@ def parse_flags(flags: List[str]) -> Options:
         print_stack_structs=args.print_stack_structs,
         unk_inference=args.unk_inference,
         passes=args.passes,
-        incbin_dir=args.incbin_dir,
+        incbin_dirs=args.incbin_dirs,
     )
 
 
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -263,6 +263,13 @@ def parse_flags(flags: List[str]) -> Options:
         default=[],
         help="Mark preprocessor constant as undefined",
     )
+    group.add_argument(
+        "--incbin-dir",
+        dest="incbin_dir",
+        default=None,
+        type=Path,
+        help="Search path for loading .incbin directives in the input asm",
```
simonlindholm, line 271: Would be nice to allow passing this multiple times. (I agree with not using -I for this for now, since it would conflict a bit with C context)

src/parse_file.py
```diff
@@ -204,6 +204,28 @@ def parse_ascii_directive(line: str, z: bool) -> bytes:
...
+        raise DecompFailure(f"Could not parse asm_data .incbin directive: {args}")
+
+    try:
+        with (options.incbin_dir / filename).open("rb") as f:
+            f.seek(offset)
+            data = f.read(size)
+            if len(data) == size:
+                return data
+    except OSError:
+        pass
```
simonlindholm, line 224: would be nice to at least warn about the paths that have failed to be opened, for user awareness

src/parse_file.py
```diff
@@ -382,8 +403,13 @@ def process_label(label: str, *, glabel: bool) -> None:
                             raise DecompFailure(
                                 f"Could not parse asm_data .space in {curr_section}: {line}"
```
simonlindholm, line 404:
```suggestion
                                f"Could not parse asm_data {directive} in {curr_section}: {line}"
```

src/parse_file.py
```diff
@@ -204,6 +204,28 @@ def parse_ascii_directive(line: str, z: bool) -> bytes:
     return b"".join(ret)
 
 
+def parse_incbin(args: List[str], options: Options) -> Optional[bytes]:
+    if options.incbin_dir is None:
+        return None
+    try:
+        filename = args[0].strip("'\"")
```
simonlindholm, line 211: Add a todo about reusing the .ascii parser?

zbanks:
Tested error messages:
```c
/* Unable to find fake.dol in any of 2 search paths */
/* Unable to read 20 bytes from baserom.dol at 0x37853400 (got 0 bytes) */
/* Unable to read 20971520 bytes from baserom.dol at 0x378534 (got 476044 bytes) */
```

src/main.py
```diff
@@ -263,6 +263,14 @@ def parse_flags(flags: List[str]) -> Options:
         default=[],
         help="Mark preprocessor constant as undefined",
     )
+    group.add_argument(
+        "--incbin-dir",
+        dest="incbin_dirs",
+        action="append",
+        default=[],
+        type=Path,
+        help="Search paths for loading .incbin directives in the input asm",
```
simonlindholm, line 272:
```suggestion
        help="Add search path for loading .incbin directives in the input asm",
```

src/parse_file.py
```diff
@@ -204,6 +204,40 @@ def parse_ascii_directive(line: str, z: bool) -> bytes:
...
+        if len(data) != size:
+            print(
+                f"/* Unable to read {size} bytes from {full_path} at {offset:#x} (got {len(data)} bytes) */"
+            )
+            return None
+        return data
+
+    print(
+        f"/* Unable to find {filename} in any of {len(options.incbin_dirs)} search paths */"
+    )
```
simonlindholm, line 237: This warning could be useful even if not passing any incbin_dirs at all, together with a note about which command line flag allows that. Though we might need to deduplicate warnings so it doesn't get spammy.

simonlindholm (approved):

src/parse_file.py
zbanks: I deduped using a global variable, do you think that's an OK compromise? Or should I really keep it isolated to `parse_file` calls?

simonlindholm: We're pretty good at avoiding global variables elsewhere, and this seems like an easy case to handle. E.g. pass in a warning set/dict which the function adds to, which is nice in that it moves prints higher up the call stack too.

PR #192, opened 2021-12-13
Move MIPS instruction handling to a new python file

zbanks:
Happy to take any feedback here, now is a great time to make name changes or do any restructuring.

Potential weird things:
- I put the new `Arch` on `GlobalInfo`, but maybe it should also be in `Options`? Or only in `Options`? (Maybe `Options` should be available in more fns)
- I use a singleton instance of `MipsArch()`, instead of passing the class itself around. This *feels* like it would play better with the type system, but I'm not actually sure?

### DIFF
--- a/src/arch_mips.py
+++ b/src/arch_mips.py
@@ -1,0 +1,480 @@
+from .types import Type
+from .parse_instruction import Register
+from .translate import (
+    Arch,
+    BinaryOp,
+    CmpInstrMap,
+    CommentStmt,
+    ErrorExpr,
+    Expression,
+    InstrMap,
+    InstrSet,
+    Literal,
+    PairInstrMap,
+    StmtInstrMap,
+    StoreInstrMap,
+    UnaryOp,
+    as_f32,
+    as_f64,
+    as_int64,
+    as_intish,
+    as_intptr,
+    as_ptr,
+    as_s32,
+    as_s64,
+    as_type,
+    as_u32,
+    as_u64,
+    fn_op,
+    fold_gcc_divmod,
+    fold_mul_chains,
+    handle_add,
+    handle_add_double,
+    handle_add_float,
+    handle_addi,
+    handle_bgez,
+    handle_conditional_move,
+    handle_convert,
+    handle_la,
+    handle_load,
+    handle_lwl,
+    handle_lwr,
+    handle_ori,
+    handle_sltiu,
+    handle_sltu,
+    handle_sra,
+    handle_swl,
+    handle_swr,
+    imm_add_32,
+    load_upper,
+    make_store,
+    void_fn_op,
+)
+
+
+class MipsArch(Arch):
+    stack_pointer_reg = Register("sp")
+    frame_pointer_reg = Register("fp")
+    return_address_reg = Register("ra")
+    temporary_reg = Register("at")
+
+    base_return_regs = [Register(r) for r in ["v0", "f0"]]
+    all_return_regs = [Register(r) for r in ["v0", "v1", "f0", "f1"]]
+    argument_regs = [Register(r) for r in ["a0", "a1", "a2", "a3", "f12", "f14"]]
+    simple_temp_regs = [
+        Register(r)
+        for r in [
+            "v0",
+            "v1",
+            "t0",
+            "t1",
+            "t2",
+            "t3",
+            "t4",
+            "t5",
+            "t6",
+            "t7",
+            "t8",
+            "t9",
+            "f0",
+            "f1",
+            "f2",
+            "f3",
+            "f4",
+            "f5",
+            "f6",
+            "f7",
+            "f8",
+            "f9",
+            "f10",
+            "f11",
+            "f13",
+            "f15",
+            "f16",
+            "f17",
+            "f18",
+            "f19",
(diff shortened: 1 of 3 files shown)


simonlindholm (commented):
Looks good to me, would be nice to have @matt-kempster's opinion as well. I agree with putting this on GlobalInfo (in principle we could even have it be thread local, but let's not go there unless we need to). Options seems like more of a high level user preference thing while this is more fundamental. I also agree with instance in favor of class.

Might be nice to think about how multiple ABIs within a single ISA will work. parse_instruction.py will also need info about arch but can't use the translate.py Arch for circular dependency reasons, which would be a second reason for an Abi class. I'm thinking Arch will split into several parts, e.g. TranslateArch/InstructionMapping for the instrs_ part, and maybe the sets of registers should go into a single AbiArch class (or do we want a part of that that is shared between abis?), maybe held together by a Target?

(We can leave all of that for a later PR.)

simonlindholm:
Also, we could consider using a subdirectory for arch/

src/arch_mips.py
```diff
@@ -0,0 +1,480 @@
...
+    as_f64,
+    as_int64,
+    as_intish,
+    as_intptr,
+    as_ptr,
+    as_s32,
+    as_s64,
+    as_type,
+    as_u32,
+    as_u64,
```
matt-kempster, line 27: I think it might be a good follow-up to reduce imports by grouping these together in classes or megafunctions like `as()` which takes in an enum. 

Same for the `handle_` family.

src/translate.py
```diff
@@ -51,100 +51,45 @@
     TypePool,
 )
 
+InstrSet = Set[str]
+InstrMap = Dict[str, Callable[["InstrArgs"], "Expression"]]
+StmtInstrMap = Dict[str, Callable[["InstrArgs"], "Statement"]]
+CmpInstrMap = Dict[str, Callable[["InstrArgs"], "Condition"]]
+StoreInstrMap = Dict[str, Callable[["InstrArgs"], Optional["StoreStmt"]]]
+MaybeInstrMap = Dict[str, Callable[["InstrArgs"], Optional["Expression"]]]
+PairInstrMap = Dict[str, Callable[["InstrArgs"], Tuple["Expression", "Expression"]]]
```
matt-kempster, line 60: would it be worth making these generic types e.g. `Mapping` and `Collection`?

src/translate.py
```diff
@@ -51,100 +51,45 @@
...
+
+
+class Arch(abc.ABC):
+    stack_pointer_reg: Register
+    frame_pointer_reg: Register
+    return_address_reg: Register
+    temporary_reg: Register
```
matt-kempster, line 67: so, since `PpcArch` won't have `temporary_reg` (right?), are we going to do `Optional[Register]`? How do we know when we're "over-generifying" the situation?

matt-kempster:
I totally agree about `GlobalInfo`, that seems fine.

I don't think I follow the distinction between the singleton instance vs. a class. What would the difference be? Anyway, the way you have it seems more Pythonic.

I don't agree with using a subdirectory quite yet, as we're likely to only have 2 arches and subdirectories are kind of painful to deal with/can make it harder to explore a codebase. But I can see it becoming useful if we need to do separate translation/parsing/etc steps per architecture and need to resolve circular dependencies.

matt-kempster (approved):
Anyway, looks great!

src/arch_mips.py
zbanks: Yeah I'd be more inclined to use a class as a namespacing technique -- the `as_` fns are already mostly very thin wrappers around `as_type()`.

src/translate.py
zbanks: I decided to convert all `Register(...)` constants in `translate.py` to one of these class members, *except* in the following cases:
- `Register("zero")`, this is already a special value used by `parse_instruction.py`
- Any code related to function arguments, which will need to be replaced next anyway (and probably won't be replaced with a simple list).
- Usages of `Register("condition_bit")`, `Register("hi")`, `Register("lo")` as they relate to `Arch.instrs_float_comp` and `Arch.instrs_hi_lo`. These are used as the implicit destination(s) of certain instructions. I also figured these might get refactored when adding support for PPC, they might fit into the `CASES_IMPLICIT_DESTINATION` paradigm? 

On the other hand, I sort of expect to convert `frame_pointer_reg` and `temporary_reg` into `Optional[Register]` when adding PPC support. (But I chose to wait until it's explicitly needed?)
- `Register("ra")` is used in `overwrite_reg(...)` even if the instruction may not reference `$ra` directly. It's straightforward to convert this code to use an `Optional[Register]` instead.
- `Register("fp")` is related to `StackInfo.uses_framepointer`. I imagine for PPC we could keep `StackInfo.uses_framepointer = False` and `PpcArch.frame_pointer_reg = None`.

I guess it's a bit of a personal taste how much "preemptive refactoring" makes sense to do. I wanted to abstract away enough registers to make sure that I'd have access to the `arch` variable in all the places that need it.

simonlindholm: (Re overwrite_reg it would be equally fine to remove that code, it's already if False'd)

PR #195, opened 2021-12-27
Don't restore registers after a function call with a reference to the stack

zbanks:
The `local_var_writes` dict is used to track the last value written to each location on the stack, and is used to elide register restore statements (i.e. after function calls).

However, if a function is called with a reference to the stack, the function *may* modify the stack and invalidate the value held in this dict.

If this is detected, then clear out the `local_var_writes` dict to disable the register-restore logic.

This implementation is maybe a bit too conservative: it considers expressions like `*(&sp10)` to be a reference. I did [try to fix this](https://github.com/matt-kempster/mips_to_c/commit/142948363e660a6d1b6755de03516daec954d9ca), but I abandoned  it because it had mostly the same output, while adding a lot of complexity & still leaving unhandled edge-cases.

[Diffs](https://gist.github.com/zbanks/c13019078a8907e7e0685bb6702c63c6)
- Original issue is fixed in [`update_actor_shadow`](https://gist.github.com/zbanks/c13019078a8907e7e0685bb6702c63c6#file-pm-diff-L12-L13).
- PM seems mostly improved
- In OOT/MM there are some `spX = spX;` expressions that look like they *were* register restores? [Example](https://gist.github.com/zbanks/c13019078a8907e7e0685bb6702c63c6#file-mm_oot-diff-L11182-L11201), [Matching](https://github.com/zeldaret/oot/blob/master/src/code/z_kankyo.c#L908-L923). Not really sure to what extent these should be considered a regression?

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -3625,6 +3625,19 @@ def translate_node_body(node: Node, regs: RegInfo, stack_info: StackInfo) -> Blo
             if reg in regs:
                 del regs[reg]
 
+    def maybe_clear_local_var_writes(func_args: List[Expression]) -> None:
+        # Clear the `local_var_writes` dict if any of the `func_args` contain
+        # a reference to a stack var. (The called function may modify the stack,
+        # replacing the value we have in `local_var_writes`.)
+        for arg in func_args:
+            if uses_expr(
+                arg,
+                lambda expr: isinstance(expr, AddressOf)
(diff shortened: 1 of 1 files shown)


simonlindholm (approved):
Seems fine. In practice we could probably also get away with just removing restores for the exact addresses taken pointers to, and also not checking containment recursively, if we want a smaller diff, but the results look okay.

PR #197, opened 2021-12-28
Weaken int casts

zbanks:
Edit: I drastically simplified this after opening the PR. Some aspects of this branch were introduced in other PRs, and the scope of this change was reduced to just the "part 2" section below. 

The original description may still be helpful to reference, and is preserved below:

---

There was some discussion of this on Discord, but this change weakens most of the `as_s32`/`as_u32` casts into `as_sint`/`as_uint` as well as uses `silent=True` as the default for these casts.

I broke this change up into 3 separate commits, and tracked the diffs for each separately:
- Part 1: use `silent=True` for `as_s32`/`as_u32` (`0f37596`)
   - I left `as_s64`/`as_u64` as `silent=False` by default because these types are much harder for mips2c to get right?
   - [Diffs for OOT, MM, PM](https://gist.github.com/zbanks/d8655acae883aa80a01c265c7acde0f8)
- Part 2: introduce `Type.signed_int()`/`Type.unsigned_int()` and replace most `as_s32`/`as_u32`'s with `as_sint`/`as_uint` (`1356f1b`)
   - [Diffs for OOT, MM, PM](https://gist.github.com/zbanks/674e4f51171aa3913a36e4dc2f981034)
- ~Part 3: unify the left/right types in the `BinaryOp.[isu]cmp` functions.~ (`447423b`)
   - This helps "undo" some of the precision loss from part 2.
   - With `scmp`/`ucmp`, the left/right should have consistent types.
   - With `icmp`, there's a somewhat-common edge case where a iterating pointer is compared to an address at the *end* of an array, and this address often is a symbol with an unrelated type. 
       - A more sophisticated strategy could be to only do a "shallow" unification, but I don't think we have such a concept yet, and this doesn't seem like a good time to introduce it.
   - With `fcmp` or `f32`, the left/right are already cast to a specific type, so futher unifying them would have no effect.
   - [Diffs for OOT, MM, PM](https://gist.github.com/zbanks/f3ba05509f4e83e429bf4e137aabfe9e)
   - **Edit**: I reverted this change, as discussed on Discord.  It's not providing much benefit, and it's somewhat unrelated to the main goal of this PR.


simonlindholm:
I pushed the unary negation case of this as 4c1eb54ddf3b7194fae1dae0c28ac2101c1a646f (already done by ppc).

Shifts and comparisons I'm less sure about, because signed/unsigned have different behavior there and it may be useful signal... Maybe we can improve is_type_obvious instead, e.g. have it use StructAccess.late_has_known_type or (more annoying) check if UnaryOp/BinaryOp's follow C int promotion rules?

Part 2 looks good to me. (Sorry to get to this so late.)

zbanks:
(blast from the past! trying to clean things up...)

The IR pattern PR introduced `Type.sintish` & `Type.uintish`, so I've incorporated those names. 
I also removed "Part 1", changes to `silent`, so this PR is *just* "Part 2": replacing `as_s32`/`as_u32` with `as_sintish`/`as_uintish`. 

The diffs are _very_ large, but here's a representative sample:
https://gist.github.com/zbanks/451bce4c678dc80fe10016f165c7fff2

simonlindholm:
Every instance of "ish" really, including as_intish and Type.intish. Struck me as not adding much and asymmetric with ptr and intptr now on reread. (I don't care much though.)

zbanks:
Heh, unfortunately we can't have a method called `int` in Type (it confuses mypy at least) -- so I'll leave all the `ish`es for now so that it's consistent with the existing `Type.intish` & `as_intish` (even though it's not consistent with `intptr`)

PR #198, opened 2021-12-29
Match stack args to their slots based on offset

zbanks:
Use the expected `offset` of each function arg to choose the correct arg from `subroutine_args`, and emit an error if it is missing.

Additionally, if the function has `params_known`, it is not necessary to re-unify the parameter types. This was causing frame shift issues when there was a return struct or a >4-byte parameter. This fix causes most of the diffs in the project tests.

[Diffs](https://gist.github.com/zbanks/824e7811836821cca181ed675bbac4c0) (a bit odd that OOT had no diffs, but it's a relatively small corpus)

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -3494,7 +3494,7 @@ def translate_node_body(node: Node, regs: RegInfo, stack_info: StackInfo) -> Blo
 
     to_write: List[Union[Statement]] = []
     local_var_writes: Dict[LocalVar, Tuple[Register, Expression]] = {}
-    subroutine_args: List[Tuple[Expression, SubroutineArg]] = []
+    subroutine_args: Dict[int, Expression] = {}
     branch_condition: Optional[Condition] = None
     switch_expr: Optional[Expression] = None
     has_custom_return: bool = False
@@ -3659,7 +3659,7 @@ def translate_node_body(node: Node, regs: RegInfo, stack_info: StackInfo) -> Blo
                 # About to call a subroutine with this argument. Skip arguments for the
                 # first four stack slots; they are also passed in registers.
                 if to_store.dest.value >= 0x10:
-                    subroutine_args.append((to_store.source, to_store.dest))
+                    subroutine_args[to_store.dest.value] = to_store.source
             else:
                 if isinstance(to_store.dest, LocalVar):
                     stack_info.add_local_var(to_store.dest)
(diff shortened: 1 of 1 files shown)


src/translate.py
```diff
@@ -3808,16 +3814,19 @@ def process_instr(instr: Instruction) -> None:
...
+                func_args.append(arg)
 
             if not fn_sig.params_known:
                 while len(func_args) > len(fn_sig.params):
                     fn_sig.params.append(FunctionParam())
-            for i, (arg_expr, param) in enumerate(zip(func_args, fn_sig.params)):
-                func_args[i] = as_type(arg_expr, param.type, True)
+                # We can only perform this naive zip when `params_known` is False, and
+                # we're assuming that each parameter is "basic" (<=4 bytes, no return struct, etc.)
+                # Otherwise, we would need to duplicate the logic from `function_abi` here.
```
simonlindholm, line 3827: so basically, when params_known = False, a FunctionSignature doesn't represent a real function signature but rather the ABI representation of one. (maybe an observation worth noting in a comment somewhere?)

src/translate.py
```diff
@@ -3755,6 +3755,12 @@ def process_instr(instr: Instruction) -> None:
             for slot in abi.arg_slots:
                 if slot.reg:
                     func_args.append(as_type(regs[slot.reg], slot.type, True))
+                elif slot.offset in subroutine_args:
+                    func_args.append(
+                        as_type(subroutine_args.pop(slot.offset), slot.type, True)
+                    )
+                else:
+                    func_args.append(ErrorExpr(f"Missing stack arg {slot.offset:#x}"))
```
simonlindholm, line 3763: might be nice to extend the text to note that this is likely due to a m2c limitation

simonlindholm (approved):
One thing that stands out in the diff is casts to e.g. Vec3f for each field of Vec3f structs passed as arguments, for which
```py
                slots.append(
                    AbiArgSlot(offset=4 * i, reg=reg2, name=name2, type=param_type)
                )
```
not special-casing size != 4 is to blame. Maybe we should use Type.any() there. (Doing type inference on ABI representation level could have been useful here, even though we have a C function signature!)

Though, admittedly, that cast in the output is probably fine for indicating that something weird is going on with struct passing, so perhaps we should just intentionally leave that unfixed.

simonlindholm:
> Maybe we should use Type.any() there

or None, and resolving that in the caller e.g. by emitting an inline comment about the parameter being part of a >4-byte type

zbanks:
> One thing that stands out in the diff is casts to e.g. Vec3f for each field of Vec3f structs passed as arguments, for which
> 
> ```python
>                 slots.append(
>                     AbiArgSlot(offset=4 * i, reg=reg2, name=name2, type=param_type)
>                 )
> ```
> 
> not special-casing size != 4 is to blame. Maybe we should use Type.any() there. (Doing type inference on ABI representation level could have been useful here, even though we have a C function signature!)
> 
> Though, admittedly, that cast in the output is probably fine for indicating that something weird is going on with struct passing, so perhaps we should just intentionally leave that unfixed.

Yeah -- I think this is relatively rare, and the types should help the user sort things out?
I like the idea of using `None` + comments best, I might do that in a separate PR though.

src/translate.py
zbanks: The way I was picturing it, is when `params_known = False`, we assume that the real function signature resembles the "naive" ABI representation. It's hard to tell if the first argument is a return struct pointer, or if set of 3 `f32`s on the stack are separate arguments or a single `Vec3f`.

I'll try to clarify this, though.

zbanks:
I'm gonna merge this, and handle slot-spanning args in a new PR.

PR #199, opened 2021-12-29
Add comments when args span multiple slots

zbanks:
Follow-up to Simon's suggestion in #198.

For large arguments that span multiple `AbiArgSlot`s, add inline comments indicating their type & offset. This is done using a new `CommentExpr` wrapper. I also changed the argument names to add the `_unk0` if the argument does not fit in one slot, instead of starting with `_unk4`.

Warn when there are unused `subroutine_args` for functions with known parameters by including an `extra?` comment.

[Diffs](https://gist.github.com/zbanks/c26b1658f1050eaf3db48c0592214ce6)
- The `/* extra? */`s for `create_entity` in PM are false positives: the prototype in my context file does not have an ellipsis, unlike the `main` branch.
- I'm not sure how these arguments should interact with `StackInfo.add_known_param(...)` -- what type should be used when an argument spans multiple slots? [Example](https://gist.github.com/zbanks/c26b1658f1050eaf3db48c0592214ce6#file-all-diff-L188-L191): calling `right_unk0` a `u64` was probably more helpful than calling it a `s64`, but I don't know how best to generalize this?

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -32,7 +32,7 @@ from .flow_graph import (
     SwitchNode,
     TerminalNode,
 )
-from .options import Formatter, Options
+from .options import CodingStyle, Formatter, Options
 from .parse_file import AsmData, AsmDataEntry
 from .parse_instruction import (
     Argument,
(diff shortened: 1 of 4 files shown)


zbanks:
> huh, what's up with the missing `temp_a1_3 = temp_t7 + 0xFA;` in the diff

I screwed up the `dependencies` function, it's fixed in the next commit.

PR #200, opened 2021-12-31
Attempt to follow functions that have a definition

Kelebek1:
MM diffs: https://gist.github.com/Kelebek1/d35ad7516b68469d67a53d9a6a7917cd

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -146,7 +146,8 @@ def run(options: Options) -> int:
         preliminary_infos = []
         function_set: Set[str] = {f.name for f in functions.values()}
 
-        while function := get_next_func(function_set, functions, global_info):  # type: ignore
+        function = get_next_func(function_set, functions, global_info)  # type: ignore
+        while function:
             flow_graph = flow_graphs[function.name]
             try:
                 if isinstance(flow_graph, Exception):
@@ -156,6 +157,8 @@ def run(options: Options) -> int:
                 preliminary_infos.append(info)
             except:
                 pass
+            function = get_next_func(function_set, functions, global_info)  # type: ignore
+
         try:
             global_info.global_decls(fmt, options.global_decls, [])
         except:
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -97,71 +116,75 @@ def run(options: Options) -> int:
...
+    function_names: List[str] = list(functions.keys())
+    function = get_next_func(function_names, functions, global_info)
+    while function:
         try:
-            flow_graphs.append(build_flowgraph(function, global_info.asm_data))
+            flow_graph = build_flowgraph(function, global_info.asm_data)
         except Exception as e:
             # Store the exception for later, to preserve the order in the output
-            flow_graphs.append(e)
+            flow_graph = e  # type: ignore
```
zbanks, line 146: We don't use `# type: ignore` in this codebase, `flow_graph` should have type `Union[FlowGraph, Exception]`

src/main.py
```diff
@@ -59,6 +59,25 @@ def print_exception_as_comment(
         print("*/")
 
 
+def get_next_func(
+    names: List[str], functions: Dict[str, Function], globalinfo: GlobalInfo
+) -> Optional[Function]:
```
zbanks, line 64: Maybe this could be turned into an iterator, `yield`'ing `Function`s instead of returning one `Optional[Function]` at a time? 

Can you add a comment describing what this function is doing/why?

src/main.py
```diff
@@ -85,9 +104,9 @@ def run(options: Options) -> int:
         return 0
 
     if not options.function_indexes_or_names:
-        functions = list(all_functions.values())
+        funcs = list(all_functions.values())
```
zbanks, line 107: I think using `funcs` and `functions` to be two separate variables for different things is hard to read. Particularly later on when there's the `for ..., (func, function) in ...`.

I think `funcs` could be `function_names` (or `fn_names`, `func_names`, etc.)?

src/main.py
```diff
@@ -97,71 +116,75 @@ def run(options: Options) -> int:
...
-    for i in range(options.passes - 1):
-        preliminary_infos = []
-        for function, flow_graph in zip(functions, flow_graphs):
+    # Perform x passes to improve type resolution, but discard the results/exceptions
+    for p in range(options.passes):
+        function_names: List[str] = list(functions.keys())  # type: ignore
+
+        function = get_next_func(function_names, functions, global_info)
+        while function:
+            flow_graph = flow_graphs[function.name]  # type: ignore
```
zbanks, line 156: Same comment as above, avoid `type: ignore`

src/main.py
```diff
@@ -3,7 +3,7 @@
 import sys
 import traceback
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, Set
```
zbanks, line 6: I think `Set` is now unused?

zbanks (changes requested):

PR #201, opened 2022-01-09
Refactor `Arch` & `Instruction`, create `ArchAsm` ABC

zbanks:
This PR has no (intentional!) functional changes; it is just moving the methods out of `Instruction` and into the new `ArchAsm` ABC. This is to prepare for incorporating PPC support from the `ppc2cpp` branch.

`Arch` is now a "super-interface" that also implements `ArchAsm`, so `MipsArch` implements *both* the translate-related behaviors *and* the instruction-related functions. 

I opted to pass `arch: ArchAsm` around to a lot of the functions in `flow_graph` -- the alternative would be to add it as a member to `Function`, but that felt inelegant?

### DIFF
--- a/src/parse_instruction.py
+++ b/src/parse_instruction.py
@@ -1,77 +1,11 @@
 """Functions and classes useful for parsing an arbitrary MIPS instruction.
 """
+import abc
 from dataclasses import dataclass, replace
 import string
 from typing import List, Optional, Set, Union
 
 from .error import DecompFailure
-
-LENGTH_TWO: Set[str] = {
-    "neg",
-    "negu",
-    "not",
-    "neg.s",
-    "abs.s",
-    "sqrt.s",
-    "neg.d",
-    "abs.d",
-    "sqrt.d",
-}
-
-LENGTH_THREE: Set[str] = {
-    "slt",
-    "slti",
-    "sltu",
-    "sltiu",
-    "addi",
-    "addiu",
-    "addu",
-    "subu",
-    "daddi",
-    "daddiu",
-    "dsubu",
-    "add.s",
-    "sub.s",
-    "div.s",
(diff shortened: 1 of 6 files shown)


src/parse_instruction.py
```diff
@@ -352,128 +286,55 @@ def derived(
...
+    all_return_regs: List[Register]
+    argument_regs: List[Register]
+    simple_temp_regs: List[Register]
+    temp_regs: List[Register]
+    saved_regs: List[Register]
 
-    def __str__(self) -> str:
-        args = ", ".join(str(arg) for arg in self.args)
-        return f"{self.mnemonic} {args}"
+    @staticmethod
```
simonlindholm, line 306: no need to make this a staticmethod in the parent interface, should be fine for derived classes to keep state

(unless Python requires it)

simonlindholm (approved):

matt-kempster (approved):
I was initially hesitant about code like `arch.is_delay_slot_instruction(last_ins)` because it didn't seem intuitive to me that the `ArchAsm` is the correct source of truth for whether an instruction is a delay slot instruction or not. I was thinking that the code should look more like `last_ins.is_delay_slot_instruction(arch)` or passing `arch` to `Instruction` at initialization time. But the former either leads to ugly `isinstance`-style code where all the possibilities live in a bunch of places, or leads to simple pass-through code that just defers to ArchAsm. And the latter isn't always possible because of fictive/munged/placeholder instructions, and may lead to heterogeneous instructions (and also suffers from the "former" problems too). And to put the nail in the coffin, I started to realize the `ArchAsm` is actually the right place for this information to go, and will help keep us organized if we add more architectures. (All this was discussed offline with @zbanks.)

So, TL;DR, looks great.

PR #202, opened 2022-01-12
More preparation for `ppc2cpp` merge

zbanks:
This is a bunch of small changes which have little-to-no effect on the existing behavior, but can bring the main branch closer to `ppc2cpp`.

- ArchAsm: add `all_regs` & `aliased_regs`
- ArchAsm: make `frame_pointer_reg` optional
- ArchAsm: add additional fns
- Arch: set defaults for `instrs_*`
- Refactor `handle_ori()` to `handle_or()`
- Change `Node` debug reps to prefer `.name()` over `.block.index`
- Emit skipped nodes in stable order
- Use `ArchAsm` for more instruction parsing (`.all_regs`)
- Allow `$` in symbol names
- Allow `;` as a comment separator, like `#`
- Remove trailing space from zero-argument `Instruction`s
- Add project support to `run_tests.py`

----

Only diffs are from `MIPS2C_ERROR`'s when the instruction has zero arguments, like the following:
```diff
[INFO] Output of ../oot/asm/osMapTLBRdb.s changed! Diff:
     MIPS2C_ERROR(unknown instruction: mtc0 $t1, $3);
-    MIPS2C_ERROR(unknown instruction: tlbwi );
+    MIPS2C_ERROR(unknown instruction: tlbwi);
     MIPS2C_ERROR(unknown instruction: mtc0 $t0, $10);
```

### DIFF
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -1360,11 +1360,14 @@ def build_body(context: Context, options: Options) -> Body:
 
     # Check no nodes were skipped: build_flowgraph_between should hit every node in
     # well-formed (reducible) graphs; and build_naive explicitly emits every node
-    unemitted_nodes = (
+    unemitted_nodes = list(
         set(context.flow_graph.nodes)
         - context.emitted_nodes
         - {context.flow_graph.terminal_node()}
     )
+    # NB: Sorting by name isn't perfectly correct ("100" comes before "2"), but this does
+    # create a reproducable output order because names are unique (unlike `n.block.index`).
+    unemitted_nodes.sort(key=lambda n: n.name())
     for node in unemitted_nodes:
         if isinstance(node, ReturnNode) and not node.is_real():
             continue
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -1360,15 +1360,9 @@ def build_body(context: Context, options: Options) -> Body:
 
     # Check no nodes were skipped: build_flowgraph_between should hit every node in
     # well-formed (reducible) graphs; and build_naive explicitly emits every node
-    unemitted_nodes = list(
-        set(context.flow_graph.nodes)
-        - context.emitted_nodes
-        - {context.flow_graph.terminal_node()}
-    )
-    # NB: Sorting by name isn't perfectly correct ("100" comes before "2"), but this does
-    # create a reproducable output order because names are unique (unlike `n.block.index`).
-    unemitted_nodes.sort(key=lambda n: n.name())
-    for node in unemitted_nodes:
+    for node in context.flow_graph.nodes:
+        if node in context.emitted_nodes or isinstance(node, TerminalNode):
+            continue
         if isinstance(node, ReturnNode) and not node.is_real():
             continue
         body.add_comment(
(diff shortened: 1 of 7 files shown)


src/if_statements.py
```diff
@@ -1360,11 +1360,14 @@ def build_body(context: Context, options: Options) -> Body:
...
     # well-formed (reducible) graphs; and build_naive explicitly emits every node
-    unemitted_nodes = (
+    unemitted_nodes = list(
         set(context.flow_graph.nodes)
         - context.emitted_nodes
         - {context.flow_graph.terminal_node()}
     )
+    # NB: Sorting by name isn't perfectly correct ("100" comes before "2"), but this does
+    # create a reproducable output order because names are unique (unlike `n.block.index`).
+    unemitted_nodes.sort(key=lambda n: n.name())
```
simonlindholm, line 1370: could also do
```py
for node in context.flow_graph.nodes:
    if node in context.emitted_nodes or isinstance(node, TerminalNode):
        continue
```

src/parse_instruction.py
```diff
@@ -287,13 +295,15 @@ def derived(
         return Instruction(mnemonic, args, replace(old.meta, synthetic=True))
 
     def __str__(self) -> str:
+        if not self.args:
+            return self.mnemonic
         args = ", ".join(str(arg) for arg in self.args)
         return f"{self.mnemonic} {args}"
 
 
 class ArchAsm(abc.ABC):
```
simonlindholm, line 304: would be nice to move all these classes closer to the top of the file. Good style since they are externally referenced and avoids some quoted type annotations

src/parse_instruction.py
```diff
@@ -336,10 +361,11 @@ def parse_instruction(line: str, meta: InstructionMeta, arch: ArchAsm) -> Instru
...
             filter(
-                None, [parse_arg(arg_str.strip()) for arg_str in args_str.split(",")]
+                None,
+                [parse_arg(arg_str.strip(), arch) for arg_str in args_str.split(",")],
             )
         )
         instr = Instruction(mnemonic, args, meta)
         return arch.normalize_instruction(instr)
-    except Exception as e:
+    except:
```
simonlindholm, line 370:
```suggestion
    except Exception:
```

src/translate.py
```diff
@@ -2410,13 +2410,15 @@ def handle_la(args: InstrArgs) -> Expression:
     return add_imm(var, Literal(target.offset), stack_info)
 
 
-def handle_ori(args: InstrArgs) -> Expression:
-    imm = args.unsigned_imm(2)
-    r = args.reg(1)
-    if isinstance(r, Literal) and isinstance(imm, Literal) and (r.value & 0xFFFF) == 0:
-        return Literal(value=(r.value | imm.value))
+def handle_or(args: InstrArgs, left: Expression, right: Expression) -> Expression:
```
simonlindholm, line 2413: args is unused

simonlindholm (approved):

src/parse_instruction.py
zbanks: Is this to make it explicit that we're handling all exceptions? Whatever it is, I can change it.

matt-kempster: Bare `except` is dangerous: https://www.flake8rules.com/rules/E722.html

PR #203, opened 2022-01-27
Refactor asm pattern matching

simonlindholm:


src/asm_pattern.py
```diff
@@ -0,0 +1,182 @@
...
+        return True
+
+
+@dataclass
+class AsmMatcher:
+    input: List[BodyPart]
+    output: List[BodyPart] = field(default_factory=list)
+    index: int = 0
+
+    def try_match(self, pattern: Pattern) -> Optional[List[BodyPart]]:
```
zbanks, line 152: It might be nice to return a new class for the results, instead of `List[BodyPart]`? 
If there were methods like `.get_instruction(index: int) -> Instruction`, this could remove the `typing.cast` calls? (They would be `assert`s in the getters instead.)

(I've also wanted to the symbolic registers/literals from the `TryMatchState`? The replacement functions could be easier to read with `match.get_register("x")` instead of `match[0].args[1]`...)

src/parse_instruction.py
```diff
@@ -295,15 +296,14 @@ def expect(n: str) -> str:
             value = AsmLiteral(parse_number(arg_elems))
         elif tok == "(":
             # Address mode or binary operation.
-            # There was possibly an offset, so value could be a AsmLiteral or Macro.
-            assert value is None or isinstance(value, (AsmLiteral, Macro))
```
zbanks, line 299: Was there a specific case you wanted to widen the type of `AsmAddressMode.lhs` for?

zbanks (approved):

src/asm_pattern.py
simonlindholm: Agreed! I'll do that

src/parse_instruction.py
simonlindholm: Yes, stores to `N($sp)` and `(N + 4)($sp)` where N is an AsmGlobalSymbol at parse time. Not used yet though.

Issue #187, opened 2021-12-09
Track PPC instructions implemented in ppc2cpp branch


zbanks:
This issue is just used to track progress on the `ppc2cpp` branch by using github's checklist feature. 
This is a list of instructions used in SMS sorted by popularity (with their associated count).

- [ ] `lwz` (118317)
- [ ] `addi` (113645)
- [ ] `stw` (83449)
- [ ] `li` (65901)
- [ ] `bl` (58090)
- [ ] `lfs` (46314)
- [ ] `stfs` (28358)
- [ ] `beq` (24540)
- [ ] `lis` (22538)
- [ ] `b` (22268)
- [ ] `mr` (20079)
- [ ] `bne` (19009)
- [ ] `stb` (15124)
- [ ] `mtlr` (15073)
- [ ] `cmplwi` (13605)
- [ ] `blr` (13176)
- [ ] `lbz` (12208)
- [ ] `cmpwi` (11851)
- [ ] `fmuls` (10326)
- [ ] `stwu` (9580)
- [ ] `mflr` (9289)
- [ ] `lfd` (9093)
- [ ] `fsubs` (7925)
- [ ] `add` (7697)
- [ ] `slwi` (6244)
- [ ] `bge` (6223)
- [ ] `blrl` (5793)
- [ ] `sth` (5679)
- [ ] `clrlwi.` (5630)
- [ ] `stfd` (5143)
- [ ] `fadds` (5022)
- [ ] `fcmpo` (4692)
- [ ] `ble` (4658)
- [ ] `extsb.` (4604)
- [ ] `lhz` (4294)
- [ ] `fmr` (3818)
- [ ] `or.` (3803)
- [ ] `rlwinm` (3503)
- [ ] `cmpw` (3272)
- [ ] `clrlwi` (3189)
- [ ] `blt` (3178)
- [ ] `xoris` (3171)
- [ ] `fmadds` (2823)
- [ ] `rlwinm.` (2768)
- [ ] `cmplw` (2690)
- [ ] `lha` (2316)
- [ ] `lwzx` (2144)
- [ ] `subf` (2110)
- [ ] `mulli` (1770)
- [ ] `ori` (1736)
- [ ] `fctiwz` (1710)
- [ ] `extsh.` (1531)
- [ ] `cror` (1512)
- [ ] `or` (1251)
- [ ] `addis` (1238)
- [ ] `stmw` (1226)
- [ ] `lmw` (1223)
[...]

PR #204, opened 2022-01-30
PPC Support

zbanks:
This is the big PR that merges in the work of the `ppc2cpp` branch to add support for PPC & C++. Progress was primarily tracked in #187.

This branch has one commit on top of `ppc2cpp` which changes the default target arch & makes some small changes to the website. (The default target on `ppc2cpp` is `ppc-mwcc-c++`, but on the `mips_to_c` mainline we want to use `mips-ido-c` as the default.)

Known caveats (including comments marked with :rocket:)
- Not every instruction is (fully) implemented (see #187 & end-to-end tests)
- `PpcArch.function_abi()` is relatively basic, and does not support passing args on the stack.
    - `likely_regs` [comment](https://github.com/matt-kempster/mips_to_c/pull/204#discussion_r796173246)
- Correct NaN handling in float comparisons [comment](https://github.com/matt-kempster/mips_to_c/pull/204#discussion_r796176793)
- Unify `make_storex` with `handle_loadx` [comment](https://github.com/matt-kempster/mips_to_c/pull/204#discussion_r796139741)
- Refactor `output_regs_for_instr` to avoid deviating from `process_instr` [comment](https://github.com/matt-kempster/mips_to_c/pull/204#discussion_r795828854)

[Diffs from OOT/MM](https://gist.github.com/zbanks/a26b6dd75f533f38e1e4f63d306a5cf6) (MIPS-only). The only diffs are related to error messages, and some constant folding (`0x10000 + 0x4908` --> `0x14908`).


simonlindholm (commented):
Partway through.

We need to update the README to mention experimental PPC support.

src/types.py
```diff
@@ -930,6 +943,112 @@ def ctype(ctype: CType, typemap: TypeMap, typepool: TypePool) -> "Type":
...
+            elif term.kind == CxxTerm.Kind.LONG_LONG:
+                type = Type.int_of_size(64)
+            elif term.kind == CxxTerm.Kind.FLOAT:
+                type = Type.f32()
+            elif term.kind == CxxTerm.Kind.DOUBLE:
+                type = Type.f64()
+            elif term.kind == CxxTerm.Kind.LONG_DOUBLE:
+                type = Type.f128()
+            elif term.kind == CxxTerm.Kind.SIGNED:
+                assert type is not None
```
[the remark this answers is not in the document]
zbanks: I had thought these asserts should be guaranteed by `demangle_codewarrior.py` (`SIGNED`, `UNSIGNED`, `ARRAY`, etc. are all "`NONTERMINATING_KINDS`") -- but it's not actually required.

(The `array_dim` for `ARRAY`s *is* guaranteed though)

src/types.py
```diff
@@ -930,6 +943,112 @@ def ctype(ctype: CType, typemap: TypeMap, typepool: TypePool) -> "Type":
             return Type(TypeData(kind=TypeData.K_INT, size_bits=size_bits, sign=sign))
         static_assert_unreachable(real_ctype)
 
+    @staticmethod
+    def demangled_symbol(sym: CxxSymbol) -> "Type":
+        return Type.demangled_type(sym_type=sym.type, sym_name=sym.name)
+
+    @staticmethod
+    def demangled_type(sym_type: CxxType, sym_name: Optional[CxxTerm] = None) -> "Type":
+        # TODO: This function may need to depend on the ABI, so it may need to be moved out of types.py?
```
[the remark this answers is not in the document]
zbanks: I'll clarify that I was thinking about class member functions in particular. (+ constructors/destructors)

But, I think that could be supported by having `FunctionSignature` track that it is a member function, letting `function_abi` be responsible for adding the `this` argument etc.

src/types.py
```diff
@@ -930,6 +943,112 @@ def ctype(ctype: CType, typemap: TypeMap, typepool: TypePool) -> "Type":
...
+                        params_known=True,
+                        is_variadic=is_variadic,
+                        return_type=return_type,
+                    )
+                )
+            elif term.kind == CxxTerm.Kind.QUALIFIED:
+                # TODO: Support C++ classes/namespaces
+                type = Type.any()
+            elif term.kind == CxxTerm.Kind.VOID:
+                pass
```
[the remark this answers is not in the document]
zbanks: Changed, and added `.weaken_void_ptr()` to the pointer/reference case.

This improved some test outputs when functions had a void-returning function pointer as an argument, such as `sms/asm/JSystem/dsptask.s`:
```diff
 /* DspStartWork (long unsigned, void (*) (short unsigned) *) */
-s32 DspStartWork__FUlPFUs_v(u32 arg0, ? (*arg1)(u16)) {
+s32 DspStartWork__FUlPFUs_v(u32 arg0, void (*arg1)(u16)) {
```

src/types.py
```diff
@@ -930,6 +943,112 @@ def ctype(ctype: CType, typemap: TypeMap, typepool: TypePool) -> "Type":
...
+            elif term.kind == CxxTerm.Kind.ELLIPSIS:
+                pass
```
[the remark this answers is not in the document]
zbanks: This *should* be handled by the `FUNCTION` section above, but I'll add a comment explaining it.

tests/elf_file.py
```diff
@@ -0,0 +1,217 @@
...
+            if st_shndx == 0 or st_shndx >= 0xFF00:
+                section = None
+            else:
+                section = elf.sections.get(sec_names[st_shndx])
+
+            symbol = ElfSymbol(offset=st_value, name=sym_name, section=section)
+            elf.symbols[sym_name] = symbol
+            symbols_by_index.append(symbol)
+            if section is not None and (
+                st_value not in section.symbols or not sym_name.startswith("...")
```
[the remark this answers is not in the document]
zbanks: Added the following comment:
```
            # Do not overwrite existing symbols at this address, unless the existing
            # symbol starts with "..." (such as `...bss@0`).
```

I was seeing `...bss@0` as a symbol pointing to the start of the `.bss` section. This would shadow the actual symbol that we cared about at `.bss+0`. (The ordering of these two in the symtab was inconsistent.)

There may be other symbols we'd want to ignore? Also `mips_to_c` was not able to parse the symbol name `...bss@0`, but I'm not sure if that needs to be fixed...

src/flow_graph.py
```diff
@@ -391,23 +399,60 @@ def process(item: Union[Instruction, Label]) -> None:
...
     for item in body_iter:
-        process(item)
+        if arch.uses_delay_slots:
+            process_with_delay_slots(item)
+        else:
+            process_no_delay_slots(item)
 
     if block_builder.curr_label:
         # As an easy-to-implement safeguard, check that the current block is
         # anonymous ("jr" instructions create new anonymous blocks, so if it's
```
simonlindholm, line 439:
```suggestion
        # anonymous (jump instructions create new anonymous blocks, so if it's
```

src/main.py
```diff
@@ -522,6 +540,18 @@ def parse_flags(flags: List[str]) -> Options:
     )
     filenames = args.filename + args.rodata_filenames
 
+    # Backwards compatability: MIPS targets can be descibed with --compiler
```
simonlindholm, line 543:
```suggestion
    # Backwards compatibility: MIPS targets can be described with --compiler
```

src/main.py
```diff
@@ -522,6 +540,18 @@ def parse_flags(flags: List[str]) -> Options:
...
+    if args.target_compiler == Target.CompilerEnum.IDO:
+        target = Target.parse("mips-ido-c")
+    elif args.target_compiler == Target.CompilerEnum.GCC:
+        target = Target.parse("mips-gcc-c")
+    elif args.target_compiler is None:
+        target = args.target
+    else:
+        parser.error(
+            "--compiler is partially supported for backwards compatability, use --target instead"
```
simonlindholm, line 552:
```suggestion
            "--compiler is partially supported for backwards compatibility, use --target instead"
```

src/flow_graph.py
```diff
@@ -641,19 +686,19 @@ def find_block_by_label(label: str) -> Optional[Block]:
...
                         jtbl_names.append(arg.argument.symbol_name)
             if len(jtbl_names) != 1:
                 raise DecompFailure(
                     f"Unable to determine jump table for {jump.mnemonic} instruction {jump.meta.loc_str()}.\n\n"
                     "There must be a read of a variable in the same block as\n"
-                    'the instruction, which has a name starting with "jtbl"/"jpt_".'
+                    'the instruction, which has a name starting with "jtbl"/"jpt_"/"lbl_".'
                 )
 
             jtbl_name = jtbl_names[0]
```
simonlindholm, line 704: the `is_constant_branch = jump.mnemonic in ["b", "j"]` check further down in this file would be good to arch-ify

src/parse_instruction.py
```diff
@@ -334,10 +344,21 @@ def expect(n: str) -> str:
...
+            elif maybe_reg in arch.all_regs:
                 value = maybe_reg
             else:
                 value = AsmGlobalSymbol(word)
+        elif tok == '"' and '"' in arg_elems[1:]:
+            # Quoted global symbol
+            # TODO: properly match pairs of quotes, avoid splitting on internal spaces
+            expect('"')
+            symbol = ""
+            while arg_elems[0] != '"':
```
simonlindholm, line 358:
```suggestion
            while arg_elems and arg_elems[0] != '"':
```

src/parse_instruction.py
```diff
@@ -334,10 +344,21 @@ def expect(n: str) -> str:
...
             word = parse_word(arg_elems)
             maybe_reg = Register(word)
-            if maybe_reg in arch.all_regs:
+            if word in arch.aliased_regs:
+                value = arch.aliased_regs[word]
+            elif maybe_reg in arch.all_regs:
                 value = maybe_reg
             else:
                 value = AsmGlobalSymbol(word)
+        elif tok == '"' and '"' in arg_elems[1:]:
```
simonlindholm, line 353:
```suggestion
        elif tok == '"':
```

src/parse_instruction.py
```diff
@@ -334,10 +341,16 @@ def expect(n: str) -> str:
...
-            if maybe_reg in arch.all_regs:
+            if word in arch.aliased_regs:
+                value = arch.aliased_regs[word]
+            elif maybe_reg in arch.all_regs:
                 value = maybe_reg
             else:
                 value = AsmGlobalSymbol(word)
+        elif tok == '"' and arg_elems[-1] == '"':
+            # Quoted global symbol
+            # TODO: properly match pairs of quotes, avoid splitting on internal spaces
```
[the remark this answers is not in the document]
simonlindholm: the comment refers to internal spaces as being a problem, but they're not

src/parse_instruction.py
```diff
@@ -350,6 +371,24 @@ def expect(n: str) -> str:
...
+                assert value
+                macro = Macro("sda21", value)
+                if not arg_elems:
+                    return macro
+                expect("(")
+                rhs = parse_arg_elems(arg_elems, arch)
+                assert isinstance(rhs, Register)
+                assert rhs in [Register("r2"), Register("r13")]
+                expect(")")
+                return AsmAddressMode(macro, rhs)
```
simonlindholm, line 390:
```suggestion
                value = Macro("sda21", value)
```
the existing code should be able to turn the value into an AsmAddressMode if needed I think

(+ needs an else clause or moving the `if tok == "-" and arg_elems[0] == "_":` check to the outer elif chain)

src/parse_instruction.py
```diff
@@ -350,6 +371,24 @@ def expect(n: str) -> str:
...
                 op = expect("&+-*")
 
+            # Parse `sym-_SDA_BASE_(r13)` as a Macro, equivalently to `sym@sda21(r13)`
+            if tok == "-" and arg_elems[0] == "_":
+                reloc_name = parse_word(arg_elems)
+                if reloc_name not in ("_SDA_BASE_", "_SDA2_BASE_"):
+                    raise DecompFailure(
+                        f"Unexpected symbol {reloc_name} in subtraction expression"
+                    )
+                assert value
```
simonlindholm, line 381: already asserted above

src/parse_instruction.py
```diff
@@ -365,6 +404,22 @@ def expect(n: str) -> str:
...
+                rhs = parse_arg_elems(arg_elems, arch)
+                assert isinstance(rhs, Register)
+                assert rhs in [Register("r2"), Register("r13")]
+                expect(")")
+                assert value
+                return AsmAddressMode(Macro(reloc_name, value), rhs)
+            else:
+                assert reloc_name in ("h", "ha", "l", "sda2", "sda21")
+                assert value
+                value = Macro(reloc_name, value)
```
simonlindholm, line 422:
```suggestion
            assert reloc_name in ("h", "ha", "l", "sda2", "sda21")
            assert value
            value = Macro(reloc_name, value)
```

src/types.py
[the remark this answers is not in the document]
simonlindholm: Yeah, I figured that much, but shouldn't this be an error in that case?

tests/add_test.py
```diff
@@ -183,30 +232,53 @@ def replace_last(s: bytes, a: bytes, b: bytes) -> bytes:
...
     test_dir = orig_file.parent
-    for asm_filename, flags in OUT_FILES_TO_IRIX_FLAGS.items():
-        asm_file_path = Path(str(test_dir / asm_filename) + ".s")
-        irix_compile_with_flag(
-            orig_file, asm_file_path, flags + COMMON_IRIX_FLAGS, env_vars
-        )
+    for asm_filename, compiler in compilers:
+        asm_file_path = test_dir / (asm_filename + ".s")
+        try:
+            if compiler.arch == "ido":
```
simonlindholm, line 271: looks a bit silly. :) arch -> name?

tests/elf_file.py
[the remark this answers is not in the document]
simonlindholm: Technically that comment doesn't exactly match what the code does (it allows overwriting a non-`...` symbol with another one), but I see what you mean!

tests/ppc_disasm.py
```diff
@@ -0,0 +1,568 @@
...
+}
+
+# These name substitutions are performed by doldisasm.py so that all symbols
+# and labels in the output asm only use alphanumerics and '$'
+SUBSTITUTIONS: Tuple[Tuple[str, str], ...] = (
+    ("<", "$$0"),
+    (">", "$$1"),
+    ("@", "$$2"),
+    ("\\", "$$3"),
+    (",", "$$4"),
```
simonlindholm, line 97: (huh, so backslashes and commas can actually occur? I guess we'll actually have to fix the parsing at some point then)

tests/ppc_disasm.py
```diff
@@ -0,0 +1,568 @@
...
+def is_load_store_reg_offset(insn: CsInsn, reg: Optional[CsRegister]) -> bool:
+    return insn.id in LOAD_STORE_MNEMONICS and (
+        reg == None or insn.operands[1].mem.base == reg
+    )
+
+
+# Converts the instruction to a string, fixing various issues with Capstone
+def instruction_to_text(insn: CsInsn, raw: int, section: ElfSection) -> Optional[str]:
+    # Look up the relocation for this address, which is either at offset+0 or offset+2
+    # (It should be possible to determine which offset to use from either the reolcation
```
simonlindholm, line 123:
```suggestion
    # (It should be possible to determine which offset to use from either the relocation
```

src/parse_instruction.py
zbanks: Heh, I had this and then factored it out, because it shouldn't be possible? (We know `arg_elems` contains a `"` from the outer check.)

If you had the same thought, it's probably worth leaving the check to improve readability.

tests/ppc_disasm.py
```diff
@@ -0,0 +1,568 @@
...
+        elif insn.operands[1].type == cs.ppc.PPC_OP_IMM:
+            if not label:
+                return None
+            return "%s%s %s, %s" % (
+                insn.mnemonic,
+                branchPred,
+                insn.reg_name(insn.operands[0].value.reg),
+                label,
+            )
+    if reloc is not None:
```
simonlindholm, line 179: the code here looks a bit backward to me: it makes inferences on reloc kind based on circumstantial evidence and assumes sda21/l/h/ha are the only kinds that can occur. Would be nicer to have a comprehensive switch over reloc.relocation_type and error out if it fails to match anything.

(totally fine if you don't want to spend time cleaning this up though seeing as it's (at least for now) test code, and if we turn it into non-test code we'll want to carefully review/rewrite it anyway)

src/parse_instruction.py
simonlindholm: I did have the same thought at first because I read this part of the code first. But yeah, I think removing the `'"' in arg_elems[1:]` check is good too, it's weird for a parser to look ahead like that

tests/ppc_disasm.py
zbanks: I added this for the `@`s. (We could try to support parsing them directly, but it's a bit tricky with `@` being used as a macro sigil)

I know `,`s occur in templates, but overall I figured it was better to match doldisasm.py here.

tests/ppc_disasm.py
zbanks: Yeah - the original doldisasm.py *had* to make these inferences, because it was working from the linked output. 
Our job here is much easier because we have the reloc types. 

I originally tried to get away with minimal changes to the code in this region, but at this point it's barely recognizable :smile: 
I agree having a switch over `relocation_type` would be much much better - but I think I'll punt on that for now.

src/translate.py
```diff
@@ -3231,12 +3459,31 @@ def reg_at(index: int) -> List[Register]:
         return arch.all_return_regs
```
simonlindholm, line 3459: the `jal` handling should include `bl` as well, to match the treatment in the main translation function

simonlindholm: also, in the
```
                for reg, val in arch.function_return(call):
                    ...
```
loop, we should assert that it loops over everything in `arch.all_return_regs`, to ensure output_regs_for_instr is always correct

src/translate.py
```diff
@@ -3206,6 +3432,8 @@ def output_regs_for_instr(
 ) -> List[Register]:
     def reg_at(index: int) -> List[Register]:
         reg = instr.args[index]
+        if isinstance(reg, AsmAddressMode):
+            return [reg.rhs]
```
simonlindholm, line 3436: This needs a comment. It's about instrs_store_update, right? Maybe we should let that use a separate helper for clarity.

Actually, wait a minute... I don't see instrs_store_update mentioned in this function at all! Similarly instrs_load_update, and maybe others are missing too?

simonlindholm (commented):
Most of the way through translate.py.

Rename MIPSFile to something else (AsmFile?)

src/demangle_codewarrior.py
```diff
@@ -0,0 +1,460 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
zbanks: I'll add a link to https://mirror.informatimago.com/next/developer.apple.com/tools/mpw-tools/compilers/docs/abi_spec.pdf - which has a description of a *similar* format in §3.4.1.2 but it doesn't match exactly. I'll also add comments mentioning some of the ways it differs. 

The arookas gist was the only source of information I could find (which is why I needed to use it to write this implementation & credit it; there wasn't any other documentation I could find to work from)

src/types.py
```diff
@@ -930,6 +943,117 @@ def ctype(ctype: CType, typemap: TypeMap, typepool: TypePool) -> "Type":
...
+        type = None
+        for term in sym_type.terms[::-1]:
+            if type is None and term.kind in CxxTerm.NONTERMINATING_KINDS:
+                type = Type.any()
+            if term.kind == CxxTerm.Kind.CONST:
+                pass
+            elif term.kind == CxxTerm.Kind.UNSIGNED:
+                unsigned = True
+            elif term.kind in (CxxTerm.Kind.POINTER, CxxTerm.Kind.REFERENCE):
+                type = Type.ptr(type).weaken_void_ptr()
```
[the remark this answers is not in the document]
zbanks: The idea is to replace `Type.ptr(Type.void())` with `Type.ptr(None)`. (`void *`s get mangled into `Pv` which is parsed into `[POINTER, VOID]`)

It might be a good idea to add a special case to `Type.unify()` and/or refactor how the ctypes parsing behaves -- but I think that's a refactor that'd be better to do separately?

src/translate.py
```diff
@@ -1304,6 +1363,7 @@ class GlobalSymbol(Expression):
     symbol_in_context: bool = False
     type_provided: bool = False
     initializer_in_typemap: bool = False
+    demangled_symbol: Optional[CxxSymbol] = None
```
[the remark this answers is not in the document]
zbanks: The `CxxSymbol` is used in `if_statements.py` to print the demangled declaration (symbol name + type).

I could add something like `demangled_str: Optional[str]` in addition to `demangled_type`?

src/arch_ppc.py
```diff
@@ -0,0 +1,863 @@
...
+
+    asm_patterns = [
+        FcmpoCrorPattern(),
+        TailCallPattern(),
+        DoubleNotPattern(),
+    ]
+
+    instrs_ignore: InstrSet = {
+        "nop",
+        # PPC: assume stmw/lmw are only used for saving/restoring saved regs
```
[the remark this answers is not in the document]
zbanks: I left "PPC" in some of the comments but reworded them to be more natural, ex:
```diff
-                # PPC: This is a much stricter filter than we use for MIPS,
+                # We use a much stricter filter for PPC than MIPS,
```

zbanks:
I'm still going through the comments (super helpful) - there are a few "heavier" changes that I might want to separate out into their own commits, so only things that I've :+1: are things that I have resolved. (Some things I commented on but did not change yet)

src/types.py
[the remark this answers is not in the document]
simonlindholm: Hm, my comment here was a bit unclear: I was more interested in the why than the what. I thought we only did this void* weakening when using void pointers in certain circumstances, not when creating types from them in general. E.g. when parsing C contexts I don't think we do this weakening?

I feel like I remember a discussion about unifying void with anything in unify() and deciding against it.

(Totally fine to leave this here for now on the basis of "it gives better results", naturally, it just struck me as weird)

src/demangle_codewarrior.py
```diff
@@ -69,9 +78,17 @@
 ]
 
 
+def read(src: TextIOBase, size: int) -> str:
```
simonlindholm, line 81: read -> read_exact to make it easier to distinguish from .read() at a glance

src/demangle_codewarrior.py
```diff
@@ -80,14 +97,14 @@ def peeking(src: TextIOBase) -> Iterator[None]:
 
 
 def peek(src: TextIOBase, n: int = 1) -> str:
-    # Read `n` bytes from `src` without advancing the offset
+    """Read `n` bytes from `src` without advancing the offset"""
```
simonlindholm, line 100: up to

src/arch_ppc.py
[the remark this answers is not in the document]
simonlindholm: :+1:! This review comment was actually about arch_ppc.py, where "PPC:" is even more silly, but I'm glad you did translate.py as well.

src/translate.py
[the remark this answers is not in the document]
simonlindholm: Ah, I see. Yeah, I think that would be good.

src/translate.py
```diff
@@ -3643,9 +3890,27 @@ def process_instr(instr: Instruction) -> None:
...
-            to_store = arch.instrs_store[mnemonic](args)
+            if mnemonic in arch.instrs_store:
+                to_store = arch.instrs_store[mnemonic](args)
+            else:
+                # PPC specific store-and-update instructions
+                # `stwu r3, 8(r4)` is equivalent to `$r3 = *($r4 + 8); $r4 += 8;`
+                to_store = arch.instrs_store_update[mnemonic](args)
+
+                # Update the register in the second argument
+                # TODO: is it OK to do this here, or does it need to happen at the end of the fn?
```
simonlindholm, line 3903: seems ok to me fwiw

src/translate.py
```diff
@@ -3693,6 +3958,13 @@ def process_instr(instr: Instruction) -> None:
             assert branch_condition is None
             branch_condition = arch.instrs_branches[mnemonic](args)
 
+        elif mnemonic in arch.instrs_decctr_branches:
+            assert branch_condition is None
+            branch_condition = arch.instrs_decctr_branches[mnemonic](args)
+
+            ctr = Register("ctr")
+            set_reg(ctr, BinaryOp.int(args.regs[ctr], "-", Literal(1)))
```
simonlindholm, line 3966: I believe the decrement happens before the comparison. Maybe we should add an asm pattern that splits this into two fictive instructions (decrement + conditional branch) to reduce the number of instruction categories?

src/translate.py
```diff
@@ -3704,18 +3976,31 @@ def process_instr(instr: Instruction) -> None:
                 branch_condition = cond_bit.negated()
 
         elif mnemonic in arch.instrs_jumps:
-            assert mnemonic == "jr"
-            if args.reg_ref(0) == arch.return_address_reg:
-                # Return from the function.
+            if mnemonic == "b":
+                if isinstance(args.raw_arg(0), AsmGlobalSymbol):
+                    # Unconditional jump
+                    pass
```
simonlindholm, line 3982: Remnant of tail call pattern? (mips puts `b` in instrs_ignore)

src/translate.py
```diff
@@ -518,39 +558,40 @@ def get_stack_info(
...
+                        internal_padding_added = True
+                    else:
+                        raise DecompFailure(
+                            f"Gap in callee-saved word stack region. "
+                            f"Saved: {callee_saved_offset_and_size}, "
+                            f"gap at: {offset} != {top}."
+                        )
+                top = offset + size
+                last_size = size
+            # Expand boundaries to multiples of 8 bytes
```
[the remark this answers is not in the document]
zbanks: Looks like the code this was commenting got removed (oops). I'll remove the comment now.

https://github.com/matt-kempster/mips_to_c/commit/db7e7a6c17df45b5ea81f8700f55612737654f10#diff-150d167b0fde23cd7aad13a7eda91371f864fc17b009e4a2a3c91daedfebc3b1R526-R528

src/translate.py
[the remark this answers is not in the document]
zbanks: I ended up only adding `demangled_str: Optional[str]`; I didn't need to keep `demangled_type` around after calculating it in `address_of_gsym`.

src/translate.py
zbanks: I added `instrs_store_update`/`instrs_load_update`, but I need to revisit this to audit that everything matches `process_instr` and to comment/refaactor as you suggested.

src/translate.py
zbanks: Yeah I think so, I was able to put `b` in instrs_ignore for PPC too now.

src/translate.py
```diff
@@ -3450,9 +3451,13 @@ def reg_at(index: int) -> List[Register]:
...
+    if mnemonic in arch.instrs_store_update:
+        return reg_at(1)
+    if mnemonic in arch.instrs_load_update:
+        return reg_at(0) + reg_at(1)
+    if mnemonic in arch.instrs_fn_call:
+        if instr.args and isinstance(instr.args[0], AsmGlobalSymbol):
+            fn_target = instr.args[0]
             if global_info.is_function_known_void(fn_target.symbol_name):
                 return []
     if mnemonic in arch.instrs_fn_call:
```
simonlindholm, line 3463:
```suggestion
```

src/translate.py
```diff
@@ -4117,8 +4117,9 @@ def process_instr(instr: Instruction) -> None:
             # void, then don't set any of these -- it might cause us to
             # believe the function we're decompiling is non-void.
             if not is_known_void:
-                for reg, val in arch.function_return(call):
-                    assert reg in arch.all_return_regs
+                return_reg_vals = arch.function_return(call)
+                assert {r for r, v in return_reg_vals} == set(arch.all_return_regs)
```
simonlindholm, line 4121: needs a comment saying this is to match output_regs_for_instr

src/translate.py
```diff
@@ -3841,23 +4140,100 @@ def process_instr(instr: Instruction) -> None:
...
+                            silent=True,
+                            type=Type.s32(),
+                        ),
+                        op="<",
+                        right=Literal(0),
+                        type=Type.s32(),
+                    ),
+                )
+                # TODO: Implement overflow checking
+                set_reg(Register("cr0_so"), ErrorExpr(f"overflow of {target}"))
```
simonlindholm, line 4203: I'd use a fn_op and remove the TODO. I imagine this never comes up in practice anyway (?).

src/translate.py
```diff
@@ -3841,23 +4140,100 @@ def process_instr(instr: Instruction) -> None:
...
+                # until the resulting bit is .use()'d.
+                set_reg(
+                    Register("cr0_gt"),
+                    BinaryOp(
+                        left=Cast(
+                            expr=target_reg,
+                            reinterpret=True,
+                            silent=True,
+                            type=Type.s32(),
+                        ),
```
simonlindholm, line 4182: Create a variable for this Cast to make the BinaryOp creation fit on a single line. Can reuse the same var for cr0_lt too

src/translate.py
```diff
@@ -3841,23 +4140,100 @@ def process_instr(instr: Instruction) -> None:
...
+            set_reg(update_reg, add_imm(args.regs[update_reg], offset, stack_info))
+
         else:
             expr = ErrorExpr(f"unknown instruction: {instr}")
+            if mnemonic.endswith("."):
+                # Unimplemented PPC instructions that modify CR0
+                set_reg(Register("cr0_eq"), expr)
+                set_reg(Register("cr0_gt"), expr)
+                set_reg(Register("cr0_lt"), expr)
+                set_reg(Register("cr0_so"), expr)
```
simonlindholm, line 4236: (this logic should go in output_regs_for_instr too, ideally)

src/translate.py
```diff
@@ -3643,9 +3895,26 @@ def process_instr(instr: Instruction) -> None:
...
+                # PPC specific store-and-update instructions
+                # `stwu r3, 8(r4)` is equivalent to `$r3 = *($r4 + 8); $r4 += 8;`
+                to_store = arch.instrs_store_update[mnemonic](args)
+
+                # Update the register in the second argument
+                update = args.memory_ref(1)
+                if not isinstance(update, AddressMode):
+                    raise DecompFailure(
+                        f"Unhandled store-and-update arg in {instr}: {update!r}"
+                    )
```
simonlindholm, line 3912: This error occurs for all storex's, right? Would be nice to unify code for that with loadx. (We can fix this later.)

src/translate.py
```diff
@@ -3841,23 +4140,100 @@ def process_instr(instr: Instruction) -> None:
...
+                        type=Type.s32(),
+                    ),
+                )
+                # TODO: Implement overflow checking
+                set_reg(Register("cr0_so"), ErrorExpr(f"overflow of {target}"))
+
+            elif (len(mn_parts) >= 2 and mn_parts[1] == "d") or mnemonic == "ldc1":
                 set_reg(target.other_f64_reg(), SecondF64Half())
 
+        elif mnemonic in arch.instrs_load_update:
```
simonlindholm, line 4208: With the load/store_update instructions, it could potentially be nice if we were somehow able to share EvalOnce for the addition created by the handler in arch.instrs_load_update[], and for the update. Particularly for x versions, since the imm ones likely won't end up needing a temp. Just theorizing though, I don't know if this would help in practice. I also guess ux instructions are probably quite rare in practice. (Edit: just checked and there's a grand total of 0 of them in the instruction tracking issue. :) So let's maybe not worry too much here.)

We could potentially get rid of this whole category of instructions by asm-pattern-replacing load-update's with update + load or load + update (with scratch registers if needed). (There's a the risk that addi + load with offset 0 won't give as good results as a single load with correct offset; I don't know how much pattern detection we do in that regard.)

Let's assert that rA != rB. ("If rA = 0 or rA = rD, the instruction form is invalid.")

simonlindholm (commented):
Through translate.py!

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+            ]
+            or PpcArch.is_constant_branch_instruction(instr)
+        )
+
+    @staticmethod
+    def is_branch_likely_instruction(instr: Instruction) -> bool:
+        return False
+
+    @staticmethod
+    def get_branch_target(instr: Instruction) -> JumpTarget:
```
simonlindholm, line 294: This method is identical in mips, can make it a method of Instruction

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+                "cr0_eq",
+                "cr0_so",
+                "ctr",
+            ]
+        ]
+    )
+    saved_regs = [
+        Register(r)
+        for r in [
+            # TODO: Some of the bits in CR are required to be saved (but usually the whole reg is?)
```
simonlindholm, line 204: But we don't care about those parts of CR, so it's not listed?

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+        "cror 2, N, 2",
+    )
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        fcmpo = m.body[0]
+        assert isinstance(fcmpo, Instruction)
+        if m.literals["N"] == 0:
+            return Replacement([m.derived_instr("fcmpo.lte", fcmpo.args)], len(m.body))
+        elif m.literals["N"] == 1:
+            return Replacement([m.derived_instr("fcmpo.gte", fcmpo.args)], len(m.body))
```
simonlindholm, line 111: Add .fictive suffixes -- I was confused by those instructions when I saw them below, and knowing that they were made up would have helped. A doc comment for this class would be good, because I don't understand what this pattern is for.

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+                [
+                    Instruction.derived("bl", instr.args, instr),
+                    Instruction.derived("blr", [], instr),
+                ],
+                1,
+            )
+        return None
+
+
+class DoubleNotPattern(SimpleAsmPattern):
```
simonlindholm, line 135: doc comment

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+    "and",
+    "or",
+    "nor",
+    "xor",
+    "andi",
+    "xori",
+    "sllv",
+    "srlv",
+    "srav",
+}
```
simonlindholm, line 96: These seem wrong: missing a lot of instructions and `srlv` is MIPS. Do PPC assembly also have shorthands like MIPS where `ins $a, $b` expands to `ins $a, $a, $b` or can we rip this support out altogether?

simonlindholm (commented):
Done reviewing!

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+        "stwx",
+    }
+
+    @staticmethod
+    def normalize_instruction(instr: Instruction) -> Instruction:
+        # Remove +/- suffix, which indicates branch-un/likely an can be ignored
+        if instr.mnemonic.startswith("b") and (
+            instr.mnemonic.endswith("+") or instr.mnemonic.endswith("-")
+        ):
+            return PpcArch.normalize_instruction(
```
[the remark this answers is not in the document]
zbanks: How would you feel about `@classmethod`?

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+
+    @staticmethod
+    def is_jumptable_instruction(instr: Instruction) -> bool:
+        return instr.mnemonic == "bctr"
+
+    @staticmethod
+    def missing_return() -> List[Instruction]:
+        return [Instruction("blr", [], InstructionMeta.missing())]
+
+    INSTRS_R0_AS_ZERO: ClassVar[Set[str]] = {
```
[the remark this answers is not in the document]
zbanks: I originally generated it by looking for `(rA|0)` in the processor manual; but that missed some instructions like `lfs`. I need to go through this more carefully still.

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+        # Remove +/- suffix, which indicates branch-un/likely an can be ignored
+        if instr.mnemonic.startswith("b") and (
+            instr.mnemonic.endswith("+") or instr.mnemonic.endswith("-")
+        ):
+            return PpcArch.normalize_instruction(
+                Instruction(instr.mnemonic[:-1], instr.args, instr.meta)
+            )
+
+        args = instr.args
+        if len(args) >= 2 and instr.mnemonic in PpcArch.INSTRS_R0_AS_ZERO:
```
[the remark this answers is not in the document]
zbanks: There are a few unimplemented instructions where it's not the second arg, like `dcbz_l` (Data Cache Block Set to Zero Locked) - but none that I saw looked like they'd be added anytime soon.

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+            ),
+            (
+                Register("r3"),
+                Cast(expr, reinterpret=True, silent=True, type=Type.intptr()),
+            ),
+            (
+                Register("r4"),
+                as_u32(Cast(expr, reinterpret=True, silent=False, type=Type.u64())),
+            ),
+            (Register("f2"), SecondF64Half()),
```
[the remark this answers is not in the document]
zbanks: Oh good point - `$f2` is used when returning `f128`s, which we're pretty far from supporting. I'll just remove this.

src/arch_ppc.py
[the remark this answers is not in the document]
simonlindholm: also fine

src/arch_ppc.py
simonlindholm: (still missing a doc comment)

src/translate.py
simonlindholm: Oh no, I didn't mean to remove the whole block, I just meant the two identical if checks
```
    if mnemonic in arch.instrs_fn_call:
        if instr.args and isinstance(instr.args[0], AsmGlobalSymbol):
            ...
        ...
    if mnemonic in arch.instrs_fn_call:
        ...
```
could be merged together.

zbanks: oh, hah, I thought it was because of the `return []` at the end of the function (but turns out you still need the check because of that second check)

src/arch_ppc.py
```diff
@@ -0,0 +1,862 @@
...
+        "mtctr": (Register("ctr"), lambda a: a.reg(0)),
+    }
+
+    instrs_ppc_compare: PpcCmpInstrMap = {
+        "cmpw": lambda a, op: BinaryOp.sintptr_cmp(a.reg(0), op, a.reg(1)),
+        "cmpwi": lambda a, op: BinaryOp.sintptr_cmp(a.reg(0), op, a.imm(1)),
+        "cmplw": lambda a, op: BinaryOp.uintptr_cmp(a.reg(0), op, a.reg(1)),
+        "cmplwi": lambda a, op: BinaryOp.uintptr_cmp(a.reg(0), op, a.imm(1)),
+        # TODO: There is a difference in how these two instructions handle NaN
+        # TODO: Assert that the first arg is cr0
```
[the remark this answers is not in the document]
zbanks: Added the check in. 

This changed 10 of the project tests (out of 1210): on some, it changed the error message, and others now raise `DecompFailure` (because they incorrectly handled instructions before). 

Error message change:
```diff
-src.error.DecompFailure: Instruction argument $r6 must be a literal
+src.error.DecompFailure: Instruction cmpwi $cr6, $r6, 0x0 not supported (first arg must be CR0)
```

src/translate.py
zbanks: I did this refactor, but it did end up making the output a bit worse...

And despite the ordering bug here, most of the C output was "correct" before (I think it could be from a bug from the way that phi assignments happen before `branch_condition`s are evaluated?)

src/translate.py
zbanks: I added the assertion, but didn't make any of the suggested improvements.

src/demangle_codewarrior.py
```diff
@@ -0,0 +1,460 @@
...
+                    if lookahead[1] not in "QFC0123456789":
+                        continue
+                base_name = chars[:-1]
+
+        src.read(len(base_name) + strip_underscores)
+
+        if base_name in CxxSymbol.STATIC_FUNCTIONS:
+            # This is a special case. A function like `__sinit_AnimalBase_cpp` is the static
+            # constructor (ctor) for "AnimalBase.cpp".
+            # "Demangle" this into `void AnimalBase_cpp::__sinit(void)`
```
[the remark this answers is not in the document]
zbanks: I tried this refactor but it didn't help much? After the `base_name.startswith(...)`, it's kind of awkward to extract the `AnimalBase_cpp` part? (Whereas splitting there is exactly what the loop is for.)

I think if I document the loop, it'll be better?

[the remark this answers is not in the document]
simonlindholm: The comments you added help a lot, thanks! My idea with extracting the AnimalBase_cpp part was just `base_name[8:]`, which I think should work? But the new comments makes the current logic understandable too.

src/translate.py
simonlindholm: I see... The worse output is unfortunate, albeit in line with how regular `for (i = 0; i < N; i++)` output also tends to look (`do { ... temp = phi + 1; phi = temp; } while (temp < lim);`). It would be nice to improve that if we can figure out a good way. If it's a big worsening we *could* consider keeping the incorrect version, e.g. by changing the asm pattern to `move $temp, $ctr`, `addi $ctr, $ctr, -1`, `<branch> $temp`, though it'd be a bit uncomfortable. I wonder if there's cases where that gives buggy output in practice, e.g. non-loop occurrences of $ctr.

And yeah, phi assignments are treated as happening all at once at the end of the block, past branch_condition's, which isn't possible to represent in C. That's definitely something that would be nice to fix, it's tripped people up in the past, but it's not obvious how. Would need temps for phis or some more involved refactoring of the whole phi/temp system. Or some wallpapering hacks for the most common cases.

src/demangle_codewarrior.py
```diff
@@ -0,0 +1,460 @@
...
+                    strip_underscores = 1
+                    break
+                elif not (c == "_" and peek(src) == "_"):
+                    continue
+                strip_underscores = 2
+                lookahead = peek(src, 2)
+                if len(lookahead) == 2:
+                    if lookahead[1] not in "QFC0123456789":
+                        continue
+                base_name = chars[:-1]
```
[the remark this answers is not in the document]
simonlindholm: This still makes little sense to me, even after the new comments: "Peek at the char after the "__": if it could be the start of the mangled type info, do not advance the `base_name`." but why? if the mangled type info starts there, isn't that exactly where you'd want to allow a base_name break point? I'm also confused as to what `QFC0123456789` represents exactly. Examples would help here I suspect.

(nit: the two ifs can be folded into one with an `and`)

Also, all this special treatment of underscores makes we wonder if that's related to C++'s rules on symbol names:

> [C++03] 17.4.3.1.2 Global names [lib.global.names]
> 
> Certain sets of names and function signatures are always reserved to the implementation:
> 
> * Each name that contains a double underscore (__) or begins with an underscore followed by an uppercase letter (2.11) is reserved to the implementation for any use.
> * Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.

Haven't associated this with name mangling before, but it makes a certain amount of sense. Not something to rely on if we can avoid it though, with those rules mostly being treated as loose guidelines by programmers.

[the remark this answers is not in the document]
zbanks: Heh, I didn't realize double underscore is disallowed!

Unfortunately, I think there are a few examples in SMS of symbols that violate this rule, like `__start__Q27JAInter9StreamLibFv` (`JAInter::StreamLib::__start (void)`) 

I've refactored it to remove the double-negative (`if not ...: continue`), re-wrote the comment, and added some tests that are more explicitly for this loop. I also explicitly called out that this is a heuristic, and highlighted some shortcomings. 

Let me know if there's something else I could add to make this more clear?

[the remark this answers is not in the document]
simonlindholm: Fantastic! Hearing that it's a heuristic explains a lot.

src/translate.py
zbanks: Yeah - I like that if we decide we want to go back to the incorrect version, it's possible to do so by modifying `BranchCtrPatter` (we don't need to bring back `instrs_branch_decctr`).

src/types.py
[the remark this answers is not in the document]
zbanks: I'm refactoring this so that `type` is never `None`: it can start as `Type.any()`, and if the last item in `sym_type.terms` is not a `NONTERMINATING_KIND` (i.e. it *is* a terminating kind, like a `CHAR`, `SHORT`, etc.), it'll be replaced.

src/translate.py
```diff
@@ -3231,12 +3459,31 @@ def reg_at(index: int) -> List[Register]:
...
+    if mnemonic.rstrip(".") in arch.instrs_destination_first:
+        reg = reg_at(0)
+        if mnemonic.endswith("."):
+            return [
+                Register("cr0_lt"),
+                Register("cr0_gt"),
+                Register("cr0_eq"),
+                Register("cr0_so"),
+            ] + reg
+        return reg
```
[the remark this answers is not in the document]
zbanks: I fixed the case for these mnemonics, but I agree it'd be nice to revisit this and figure out how to remove the duplication between `output_regs_for_instr` and `process_instr`.

src/translate.py
```diff
@@ -3746,6 +4035,15 @@ def process_instr(instr: Instruction) -> None:
...
             likely_regs: Dict[Register, bool] = {}
             for reg, data in regs.contents.items():
+                # We use a much stricter filter for PPC than MIPS,
+                # but it seems more accurate because the same registers are used
+                # for arguments & return values. The ABI can also mix & match the
+                # rN & fN registers, which makes the "require" heuristic less powerful.
+                if arch.arch == Target.ArchEnum.PPC and (
+                    data.meta.function_return or data.meta.inherited
+                ):
+                    continue
```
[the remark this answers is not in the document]
zbanks: Yeah, you're right. Changing this improves the output. 

https://gist.github.com/zbanks/0e165e64fbee562809d7bbd1dbf44a2b

src/arch_ppc.py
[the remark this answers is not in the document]
zbanks: Updated the list more carefully, by reading each instruction's page in the processor manual. 

I added all instructions, even ones that I don't think will be implemented, so that we don't have to worry about keeping this list up to date in the future. (I also had to add support for `$r0` as the first arg, even though I don't think we'll ever implement any of those instructions.)

src/translate.py
```diff
@@ -4169,8 +4190,8 @@ def process_instr(instr: Instruction) -> None:
             # if source = target, maybe we could overwrite that variable instead
             # of creating a new one?
             set_reg(target, val)
-            mn_parts = mnemonic.split(".")
-            if mnemonic.endswith("."):
+            mn_parts = arch_mnemonic.split(".")
+            if arch_mnemonic.startswith("ppc:") and arch_mnemonic.endswith("."):
```
simonlindholm, line 4205: (a bit weird that we're doing `.rstrip(".")` for all archs, but only giving the instruction special behavior for ppc. I guess making that strictly correct just bloats the code unnecessarily much though)

src/translate.py
zbanks: I don't know about a helper, because it'd still need to handle both the `Register` and ` AsmAddressMode` cases.
I did add comments + an `assert` to the branch.

README.md
```diff
@@ -1,5 +1,5 @@
 # `mips_to_c`
-Given some MIPS assembly, this program will attempt to convert it to C.
+Given some MIPS or PPC assembly, this program will attempt to convert it to C.
```
zbanks, line 2: @simonlindholm - I know you wanted me to update the README, do you think the changes I made are good enough, or are there areas I should expand on?

zbanks:
Okay, I think I've responded to all the comments and marked them with :+1:!

As mentioned in Discord, for comments that we're punting to later, I marked them with :rocket: and added a link to them in the PR description.

src/translate.py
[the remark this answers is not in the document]
simonlindholm: Nice! Judging by that diff we really need to add "if a2 is set, a1 also has to be" logic. Simpler for PPC than for MIPS!

src/translate.py
zbanks: Yeah agreed - I was kind of walking a line here between brevity & correctness, in particular because this check is repeated a few times throughout translate.py. (I felt it was worth adding `arch_mnemonic.startswith("ppc:") ` as long as it could still fit on one line)

src/translate.py
```diff
@@ -3468,13 +3475,21 @@ def reg_at(index: int) -> List[Register]:
...
             return [
                 Register("cr0_lt"),
                 Register("cr0_gt"),
                 Register("cr0_eq"),
                 Register("cr0_so"),
             ] + reg
+        elif (
+            len(mn_parts) >= 2
+            and mn_parts[0].startswith("mips:")
+            and mn_parts[1] == "d"
```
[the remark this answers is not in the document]
zbanks: It needs to match both `add.d` and `cvt.d.s`. 
I could do `.contains(".d")` or `.contains(".d.") or .endswith(".d")`? Or leave it as-is?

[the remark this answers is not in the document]
simonlindholm: Oh, right. Let's leave it then.

README.md
simonlindholm: I think it's good enough for now.

src/translate.py
```diff
@@ -465,6 +488,11 @@ def get_stack_info(
             # Moving the stack pointer.
```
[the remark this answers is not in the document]
simonlindholm: > change the check for jal above to cover ppc as well: if inst.mnemonic in arch.instrs_fn_call:

I think you missed this comment

simonlindholm (approved):
Let's go!

src/translate.py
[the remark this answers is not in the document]
zbanks: Ah, I also made the change in a different part of the file: https://github.com/matt-kempster/mips_to_c/pull/204/commits/9ce787e7fb5a152d5cfa3721c4843d4fd6dfb7a6#diff-150d167b0fde23cd7aad13a7eda91371f864fc17b009e4a2a3c91daedfebc3b1L3453-R3460

I'll fix it here too.

PR #205, opened 2022-02-03
Implement rlwimi, addic, and subfic for PPC

zbanks:
I've made a PR for these instructions because I want to make sure that the `rlwimi` implementation is easy to follow, and I want to get a second opinion on my strategy for implementing the instructions that set the carry bit.

`rlwimi $rA, $rS, SH, MB, ME` is a bit of a weird instruction because it reads from `$rA` and then writes the result back there. (This doesn't cause any problems though). This instruction is commonly emitted for expressions like `x &= ~0x10` or `x |= 0x10`, so I tried to detect these to emit nicer C.

`addic` is pretty straightforward: I was initially afraid of implementing it because I was worried about implementing the *carry* part, which updates the XER register. However, it's mostly used for `addic.`, which updates the condition registers. (Whereas there is no `addi.` instruction).

Plain `addic` (not `addic.`) is significantly rarer, and seems to mostly be used as part of idioms such as the `BoolCastPattern`.

`subfic` is used in expressions like `N - x`, where `N` is an immediate constant. (Expressions like `x - N` are covered by `addi`). Again, in most cases, the carry bit isn't used later.

Other carry-related arithmetic, like `addc`/`adde` & `subfc`/`subfe` usually show up paired, and seem to be related to 64-bit integer arithmetic. Implementing 64-bit types in PPC will be a bit trickier than MIPS: I think a value can be split across an arbitrary register pair. It also has diminishing returns, because 64-bit integers are relatively rare.

[Here's an exemplary diff from melee](https://gist.github.com/zbanks/1b3ad232d896ad6c3753dd5bb8533046) with both `rlwimi` bit twiddling, and `addic.` NULL-pointer checks. 
(It *is* a bit weird it's checking if a symbol's address is NULL, but this does seem to match my understanding of the asm, and it looks like it may be from a non-null assertion macro.)


tests/end_to_end/comparison/mwcc-o4p-out.c
```diff
@@ -1,16 +1,14 @@
 extern u32 global;
 
-s32 test(s32 arg0, s32 arg1, s32 arg2) {
-    MIPS2C_ERROR(unknown instruction: addic $r0, $r5, -0x1);
+f32 test(s32 arg0, s32 arg1, s32 arg2, f32 arg8) {
```
simonlindholm, line 3: f32? (why does this argument even exist?)

src/translate.py
```diff
@@ -3350,6 +3359,10 @@ def handle_rlwinm(
     upper_bits: Optional[Expression]
     if left_mask == 0:
         upper_bits = None
+    elif isinstance(source, Literal) and (source.value << left_shift) & left_mask == (
+        source.value << left_shift
+    ):
+        upper_bits = Literal(source.value << left_shift)
```
simonlindholm, line 3365: if `isinstance(source, Literal)` we should be able to compute the full result as `((source << shift) & mask) | ((source >> (32 - shift)) & mask)` and immediately return, rather than doing the upper_bits/lower_bits thing

simonlindholm (commented):
We need tests for rlwimi.

Partially supporting addic/subfic sounds good to me.

tests/end_to_end/comparison/mwcc-o4p-out.c
zbanks: This is a somewhat-common existing issue, related to return value heuristics. `$f1` can be the argument or the return value (or both). I could fix this in `determine_return_register` - but maybe that'd be better in a separate PR?

It looks like these tests are affected:
```
> rg -c 'return.*arg8' tests/end_to_end/
tests/end_to_end/modulo-by-power-of-two/mwcc-o4p-out.c:1
tests/end_to_end/rlwimi/mwcc-o4p-out.c:1
tests/end_to_end/division-by-power-of-two/mwcc-o4p-out.c:1
tests/end_to_end/struct/mwcc-o4p-out.c:1
tests/end_to_end/comparison/mwcc-o4p-out.c:1
tests/end_to_end/unreachable-return2/mwcc-o4p-out.c:1
tests/end_to_end/signed-conversion/mwcc-o4p-out.c:1
tests/end_to_end/load-types/mwcc-o4p-out.c:1
tests/end_to_end/loop/mwcc-o4p-out.c:3
tests/end_to_end/break/mwcc-o4p-out.c:4
tests/end_to_end/mult-by-constant/mwcc-o4p-out.c:1
```

tests/end_to_end/rlwimi/orig.c
```diff
@@ -0,0 +1,24 @@
+//int x = 123;
+//enum { A, B, C, D, E, F, G, H };
+
```
simonlindholm, line 3:
```suggestion
```

simonlindholm (approved):

PR #206, opened 2022-02-09
Mark args as inherited to improve return detection

zbanks:
[Original Comment](https://github.com/matt-kempster/mips_to_c/pull/205#discussion_r802457004)

> f32? (why does this argument even exist?)

I think it makes sense to mark *both* known args & possible args as `inherited=True`? But that's the part I'm least sure about here.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -3730,8 +3730,8 @@ def propagate_register_meta(nodes: List[Node], reg: Register) -> None:
 def determine_return_register(
     return_blocks: List[BlockInfo], fn_decl_provided: bool, arch: Arch
 ) -> Optional[Register]:
-    """Determine which of v0 and f0 is the most likely to contain the return
-    value, or if the function is likely void."""
+    """Determine which of the arch's base_return_regs (i.e. v0, f0) is the most
+    likely to contain the return value, or if the function is likely void."""
 
     def priority(block_info: BlockInfo, reg: Register) -> int:
         meta = block_info.final_register_states.get_meta(reg)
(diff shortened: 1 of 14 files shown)


simonlindholm:
Interesting suggestion for a fix. I'm not sure whether it's valid or breaks some invariant (it does make some semantic sense while also not really matching my idea of how `inherited` should behave), can think more about it later. Regardless, it feels like there's a lot of overlap here with PassedInArg(copied=False); is it possible to use that instead?

zbanks:
Actually, I tried this and hit an issue when the final value of the register contained a phi. This seemed to happen when there was a function call in a branch.

Here's the diff I was trying:
```diff
     def priority(block_info: BlockInfo, reg: Register) -> int:
         meta = block_info.final_register_states.get_meta(reg)
+        expr = block_info.final_register_states.get_raw(reg)
         if not meta:
             return 3
         if meta.uninteresting:
             return 1
+        if isinstance(expr, PassedInArg) and not expr.copied:
+            return 1
         if meta.function_return:
             return 0
         return 2
```

I've added a manual test for this.

simonlindholm:
Ah, right! That's certainly an argument for using a meta bit. (Maybe we can instead replace the existing `copied` property by a meta bit? I think we have the same problem for function calls as for returns. Although, hrm, we compute the meta bits too late, so we can't really use them there... :( Early use-def analysis might solve this.) I would spontaneously lean towards using `uninteresting` or creating a new bit instead of using `inherited`, but I haven't yet looked to see if `inherited` does in fact work.

zbanks:
>  I would spontaneously lean towards using `uninteresting` or creating a new bit instead of using `inherited`, but I haven't yet looked to see if `inherited` does in fact work.

Just checking: did you mean "but I haven't yet looked to see if *uninteresting* does in fact work" here? 

In either case, I've updated the PR to use `uninteresting` instead of `inherited`.

I did have to change where `uninteresting` is populated, since the existing code assumed it was always `False` before `propagate_register_meta`. Idk if you think that this now makes this property too confusing, because it isn't exclusively set by that function?

simonlindholm (approved):

simonlindholm:
> Just checking: did you mean "but I haven't yet looked to see if uninteresting does in fact work" here?

I did mean `inherited`, but your reading probably makes more sense. :)

PR #208, opened 2022-02-16
Refactor `Instruction` to include arch information

zbanks:
Add arch-specific information to each `Instruction` that describes how it should be interpreted.

This replaces the existing `ArchAsm.is_*_instruction(...)` functions by including this information directly inside the `Instruction` object.
`AsmInstruction` is introduced to represent a "raw" instruction, with just a mnemonic + args.

This PR doesn't change any behavior (that I'm aware of); it's a refactor to help with more complex pattern matching that I'm working on. Project & end-to-end tests still pass: tested MIPS with MM, PM; tested PPC with Melee, SMS, and SMB1.

The first commit (`036e579`) is equivalent to [the changes I posted in Discord](https://github.com/matt-kempster/mips_to_c/compare/master...zbanks:8c9394e), to help make re-reviewing a bit easier.

### DIFF
--- a/src/arch_mips.py
+++ b/src/arch_mips.py
@@ -5,13 +5,16 @@ from typing import (
     Optional,
     Set,
     Tuple,
+    Union,
 )
 
 from .error import DecompFailure
 from .options import Target
 from .parse_instruction import (
+    Argument,
     AsmAddressMode,
     AsmGlobalSymbol,
+    AsmInstruction,
     AsmLiteral,
     Instruction,
     InstructionMeta,
@@ -23,6 +26,7 @@ from .asm_pattern import (
     AsmMatcher,
     AsmPattern,
     Replacement,
+    ReplacementPart,
     SimpleAsmPattern,
     make_pattern,
 )
@@ -198,9 +202,7 @@ class ModP2Pattern1(SimpleAsmPattern):
         val = (m.literals["N"] & 0xFFFF) + 1
         if val & (val - 1):
             return None  # not a power of two
-        mod = m.derived_instr(
-            "mod.fictive", [m.regs["o"], m.regs["i"], AsmLiteral(val)]
-        )
+        mod = AsmInstruction("mod.fictive", [m.regs["o"], m.regs["i"], AsmLiteral(val)])
         return Replacement([mod], len(m.body) - 1)
 
 
@@ -224,8 +226,9 @@ class ModP2Pattern2(SimpleAsmPattern):
             val += ((m.literals["LO"] + 0x8000) & 0xFFFF) - 0x8000
         if not val or val & (val - 1):
             return None  # not a power of two
-        mod = m.derived_instr(
-            "mod.fictive", [m.regs["o"], m.regs["i"], AsmLiteral(val)]
+        mod = AsmInstruction(
+            "mod.fictive",
+            [m.regs["o"], m.regs["i"], AsmLiteral(val)],
         )
         return Replacement([mod], len(m.body) - 1)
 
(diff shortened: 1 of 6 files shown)


src/arch_mips.py
```diff
@@ -5,24 +5,29 @@
...
     JumpTarget,
     Register,
+    get_jump_target,
 )
 from .asm_pattern import (
     AsmMatch,
     AsmMatcher,
     AsmPattern,
     Replacement,
+    ReplacementPart,
```
simonlindholm, line 30: unused

src/flow_graph.py
```diff
@@ -312,10 +311,10 @@ def process_with_delay_slots(item: Union[Instruction, Label]) -> None:
             block_builder.set_label(item)
             return
 
-        if not arch.is_delay_slot_instruction(item):
+        if not item.has_delay_slot:
             block_builder.add_instruction(item)
             # Split blocks at jumps, at the next instruction.
-            if arch.is_jump_instruction(item):
+            if item.is_jump():
                 block_builder.new_block()
```
simonlindholm, line 318: pre-existing: I was a bit confused at this check: turns out it's always false and I guess added for compatibility with hypothetical future arches with delay slots and delay-slot-free branches. Turn it into an assert, I guess. And maybe rename the function process_mips since it has some mips-specific stuff in it.

Btw, the
```
# (Best-effort check for whether the instruction can be
# executed twice in a row.)
```
heuristic below is perhaps something we can improve with inputs/outputs for Instruction's.

src/parse_instruction.py
```diff
@@ -283,6 +262,14 @@ def constant_fold(arg: Argument) -> Argument:
     return arg
 
 
+def get_jump_target(label: Argument) -> JumpTarget:
+    if isinstance(label, AsmGlobalSymbol):
+        return JumpTarget(label.symbol_name)
+    if not isinstance(label, JumpTarget):
+        raise DecompFailure(f"Couldn't parse instruction: invalid branch target")
```
simonlindholm, line 269: (a bit inconsistent to use an exception here but asserts elsewhere in Arch.parse(). Doesn't matter much either way though since we suppress exception messages in parse_instruction)

src/arch_mips.py
```diff
@@ -527,138 +531,155 @@ class MipsArch(Arch):
...
+        elif mnemonic == "jal":
+            # Function call to label
+            assert isinstance(args[0], AsmGlobalSymbol)
+            function_target = args[0]
+            has_delay_slot = True
+        elif mnemonic == "jalr":
+            # Function call to pointer
+            assert isinstance(args[0], Register)
+            function_target = args[0]
+            has_delay_slot = True
```
simonlindholm, line 632: `jalr` supports a two-argument form: `jalr $a, $b` jumps to `$b` and saves pc+8 to `$a` (`$a` is always `$ra` in practice and elided by most disassemblers.) translate.py supports both forms. We should move that rewrite into normalize_instruction, probably normalizing towards the one-argument form.

simonlindholm (approved):

src/flow_graph.py
zbanks: Oh, that makes a lot of sense. I was also a bit confused about this when implementing PPC, but kind of worked around it. 
I'll fix that, rename the fn, and add a `TODO` about the heuristic.

I'm also going to remove `ArchAsm.uses_delay_slots` and use `arch.arch == MIPS` instead.

src/parse_instruction.py
zbanks: That's fair - shifting this around means the exception is now suppressed.

src/arch_mips.py
zbanks: Moved to `normalize_instruction`.
Technically, the previous code had a specific error for > 2 args, but I removed it because we mostly don't perform those kinds of assertions in translate.py?

PR #209, opened 2022-02-17
Add input/output argument decoding

zbanks:
Instead of merging this now, I could instead wait until I've written more code that uses these new fields? 

I plan on moving ` regs_clobbered_until_dominator` and `reg_always_set` into flow_graph or a new file and pre-calculating a data structure that tracks reg usage. 

Places I particularly want feedback:

- Should the `Instruction` inputs/outputs/clobbers be `Argument`, or more specific like `Union[Register, AsmAddressMode]`? (I am also considering that I may want to add the *size* alongside the `AsmAddressMode` information, and add another object to represent PPC indexed stores/loads like `stwx $rS, $rA, $rB`)

- Is it too weird to edit the inputs/outputs/clobbers in `narrow_func_call_outputs`? Is that function too weird in general?


simonlindholm (commented):
Somewhat cursory review.

`Argument` simultaneously includes too much (literals, binops, ...) and too little (stu instructions that update AsmAddressMode registers, store widths), so I definitely think we should use another type than that. (I agree that store widths make a lot of sense to record.) We could consider using a Union[Register, StackStore(loc, size)], since general "store to memory at register" seem tricky to do anything meaningful with. Unless the plan is to use that for "prevent" stuff?

The code in arch_*.py looks more boilerplate-y and annoying to review and easy to get wrong than what we had before, and there's no asserts for it yet so it's hard to test. Maybe we should wait on refactoring instruction maps and/or those asserts before landing this? The new translate.py looks really clean, on the other hand.

I agree `narrow_func_call_outputs` is weird, seems easy to accidentally call at the wrong time and operate on incorrect values before that. But I haven't thought about this too deeply.

src/asm_pattern.py
```diff
@@ -193,15 +193,30 @@ def derived_meta(self) -> InstructionMeta:
...
+                part = arch.parse(part.mnemonic, part.args, self.derived_meta())
+            if isinstance(part, Instruction):
+                # Update list of written registers
+                repl_writes.extend(part.outputs)
+                repl_writes.extend(part.clobbers)
+                final_instr = part
+            self.output.append(part)
+
+        # Calculate which regs are *not* written by the repl asm, but were in the input asm
+        # Denote the replacement asm as "clobbering" these regs by marking the final instr
```
simonlindholm, line 211: I should add, nice touch! I really like the extra error prevention here.

src/translate.py
```diff
@@ -4284,23 +4187,18 @@ def process_instr(instr: Instruction) -> None:
...
-                            val,
-                            emit_exactly_once=False,
-                            trivial=False,
-                            prefix=reg.register_name,
-                        )
-                    regs.set_with_meta(reg, val, RegMeta(function_return=True))
+            return_reg_vals = arch.function_return(call)
+            for reg, val in return_reg_vals:
+                if reg not in instr.outputs:
+                    continue
```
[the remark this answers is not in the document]
zbanks: Oh that's a great idea! (Also changed the output type from a list of tuples to a dict, now that we don't need the order from it)

src/asm_pattern.py
zbanks: I meant to post it before, but this change was responsible for exactly 1 diff in MM/PM, [in PM's `collision_main_lateral`](https://gist.github.com/zbanks/fdb4fe2217a9f1416e630f8648c694c0)

zbanks:
> `Argument` simultaneously includes too much (literals, binops, ...) and too little (stu instructions that update AsmAddressMode registers, store widths), so I definitely think we should use another type than that. (I agree that store widths make a lot of sense to record.) We could consider using a Union[Register, StackStore(loc, size)], since general "store to memory at register" seem tricky to do anything meaningful with. Unless the plan is to use that for "prevent" stuff?

Yeah - I think I'll switch to `Union[Register, StackStore(loc, size)]`. I don't have immediate plans on replacing the "prevent" stuff, and I think it'd be easy to extend the union in the future if we had an concrete use for tracking memory accesses.

> The code in arch_*.py looks more boilerplate-y and annoying to review and easy to get wrong than what we had before, and there's no asserts for it yet so it's hard to test. Maybe we should wait on refactoring instruction maps and/or those asserts before landing this? The new translate.py looks really clean, on the other hand.

Yeah, this is totally fair. I also think I initially underestimated the reviewing effort of this section -- I do see now it is a bit of a nightmare in that regard.

I would prefer to get this in before refactoring instruction maps, but I would be down to add asserts. (I have more progress on the usage analysis stuff & pattern matching, so I'd like to PR some of that first?) Do you think that's sufficient? Are there other things that would help add confidence or make reviewing easier? (I can post code coverage results after running on all projects tomorrow, I think that'll help a bit?)

(And maybe you can hold off on giving it a deep review until there's a plan here!)

> I agree `narrow_func_call_outputs` is weird, seems easy to accidentally call at the wrong time and operate on incorrect values before that. But I haven't thought about this too deeply.

Another idea I had was to collect all of the `is_function_known_void` symbols, and pass that set into `build_flow_graph` somehow? (maybe in `AsmData`?) This seemed more awkward, but it would be harder to use incorrectly?

zbanks:
I made `Access = Union[Register, StackAccess(offset, size), MemoryAccess(base, offset, size)]`. 

I don't have an immediate use for the tracking the details `MemoryAccess`, so I could replace it with a class that has no fields instead? I do need to have *some* marker for these: for example, any memory write *could* overwrite a stack reg.

simonlindholm:
For the sake of also covering ppc "extended" reads/writes it might be better to have an opaque MemoryAccess?

Problematic point about StackAccess: `uses_framepointer`, which takes until translate.py to get set. I wonder if that's an argument for generalized MemoryAccess, or for moving the framepointer analysis earlier, or for removing framepointer support... It was only used by a single file in sm64 (compiled with IDO -framepointer) and also comes up in functions that use alloca (which are broken in other ways and also aren't very common), so removing support would probably not have much ill effect, though I'm a bit reluctant.

> any memory write could overwrite a stack reg

a bit of a pessimistic assumption since most of the time we never generate pointers to the stack. I wonder if tracking escaped stack vars would be useful (later on). Pointer math does complicate it. Also, this is somewhat similar to how function calls can have arbitrary effects. Maybe it should be a boolean flag on the instruction instead. Not sure.

simonlindholm:
I think asserts should be sufficient yes, if we can have them for both inputs and outputs and have them pass when running project tests. (I don't necessarily *mind* reviewing the code in question -- I don't think it would be that hellish -- but I think my review would be less accurate than such asserts, and also meaningless if we refactor the instruction maps later on.)

zbanks:
> For the sake of also covering ppc "extended" reads/writes it might be better to have an opaque MemoryAccess?

The `MemoryAccess` I made does cover extended reads/writes (`offset` has type `Argument`, so the two regs are put in `base` & `offset`).

> Problematic point about StackAccess: `uses_framepointer`, which takes until translate.py to get set. I wonder if that's an argument for generalized MemoryAccess, or for moving the framepointer analysis earlier, or for removing framepointer support... It was only used by a single file in sm64 (compiled with IDO -framepointer) and also comes up in functions that use alloca (which are broken in other ways and also aren't very common), so removing support would probably not have much ill effect, though I'm a bit reluctant.

I think I'd be OK with any of of those options, but I'd maybe lean towards generalizing MemoryAccess? That doesn't seem so bad. It might have other surprising benefits down the line?

> > any memory write could overwrite a stack reg
> 
> a bit of a pessimistic assumption since most of the time we never generate pointers to the stack. I wonder if tracking escaped stack vars would be useful (later on). Pointer math does complicate it. Also, this is somewhat similar to how function calls can have arbitrary effects. Maybe it should be a boolean flag on the instruction instead. Not sure.

Yeah - I was leaning towards having pessimistic assumptions at first, and then experimenting later. 

> I think asserts should be sufficient yes, if we can have them for both inputs and outputs and have them pass when running project tests.

I've added a lot more assertions, but I'll do one more pass to make sure everything is checked.

src/translate.py
```diff
@@ -1957,8 +1957,11 @@ class RegInfo:
     stack_info: StackInfo = field(repr=False)
     contents: Dict[Register, RegData] = field(default_factory=dict)
     read_inherited: Set[Register] = field(default_factory=set)
+    active_instr: Optional[Instruction] = None
```
zbanks, line 1960: I don't *love* this... another option is to keep track of just the list of `inputs`/`outputs`? (That way no one can use this field to do `regs.active_instr.mnemonic` or other oddities?)

src/translate.py
```diff
@@ -2016,6 +2013,16 @@ def get_meta(self, key: Register) -> Optional[RegMeta]:
         data = self.contents.get(key)
         return data.meta if data is not None else None
 
+    @contextmanager
+    def current_instr(self, instr: Instruction) -> Iterator[None]:
+        self.active_instr = instr
+        try:
+            yield
+        except Exception as e:
+            raise InstrProcessingFailure(instr) from e
```
simonlindholm, line 2022:
```suggestion
            with current_instr(instr):
                yield
```
to avoid code duplication?

src/translate.py
simonlindholm: underscore-prefix it, maybe?

simonlindholm (commented):
Small bit of review. Getting late, to be continued tomorrow probably

src/arch_ppc.py
```diff
@@ -145,10 +148,13 @@ class BoolCastPattern(SimpleAsmPattern):
...
-        return Replacement(
-            [AsmInstruction("boolcast.fictive", [Register("r0"), m.regs["x"]])],
-            len(m.body),
-        )
+        body: List[ReplacementPart] = [
+            AsmInstruction("boolcast.fictive", [Register("r0"), m.regs["x"]])
+        ]
+        if m.regs["a"] != Register("r0"):
+            # Preserve `neg $a, $x` in case $a is accessed later, unless $a is $r0
+            body.append(m.body[0])
```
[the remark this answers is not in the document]
zbanks: Good point, this pattern doesn't work if $a is $r0.
Also, I think ordering of the two instructions depends on if $x is $a, or if $x is $r0.

Investigating this lead to fixing an unrelated bug in `fold_mul_chains`, which assumed all `UnaryOp`s were `-`. 

This shows up in melee's `mnsoundtest.s`, `func_8024AD58`. 
```asm
func_8024AD58:
...
// r0 holds the original value of $r4, "arg1"
/* 8024ADDC 002479BC  7C 80 00 D0 */	neg r4, r0
/* 8024ADE0 002479C0  30 04 FF FF */	addic r0, r4, -1
/* 8024ADE4 002479C4  7C 00 21 10 */	subfe r0, r0, r4
/* 8024ADE8 002479C8  54 1E 06 3E */	clrlwi r30, r0, 0x18
/* 8024ADEC 002479CC  4B FE 44 AD */	bl func_8022F298
/* 8024ADF0 002479D0  93 C1 00 24 */	stw r30, 0x24(r1)
```
And the fixed diff:
```diff
-    temp_r30_2 = arg1 * -1;
-    temp_f1 = func_8022F298(sp18, -!!arg1);
-    sp24 = (s32) temp_r30_2;
+    temp_r0 = !!arg1;
+    temp_f1 = func_8022F298(sp18);
+    sp24 = (s32) temp_r0;

```

src/arch_mips.py
```diff
@@ -606,35 +608,79 @@ def normalize_instruction(cls, instr: AsmInstruction) -> AsmInstruction:
...
             # Function call to pointer
-            assert isinstance(args[0], Register)
-            function_target = args[0]
+            assert (
+                len(args) == 2
+                and args[0] == Register("ra")
+                and isinstance(args[1], Register)
+            )
+            inputs = list(cls.argument_regs)
+            inputs.append(Register("ra"))
```
simonlindholm, line 674: we don't read from ra, do we?

src/arch_mips.py
```diff
@@ -667,14 +727,138 @@ def parse(
...
+                # Load instructions
+                assert len(args) == 2
+                inputs = [make_memory_access(args[1])]
+                if isinstance(args[1], AsmAddressMode):
+                    inputs.append(args[1].rhs)
+                if mnemonic in ("lwr", "lwl"):
+                    inputs.append(args[0])
+                elif mnemonic == "ldc1":
+                    outputs.append(args[0].other_f64_reg())
+            elif mnemonic == "la" and isinstance(args[1], AsmAddressMode):
```
simonlindholm, line 804: I think la can only take a symbol or (symbol + offset) as argument

src/arch_ppc.py
```diff
@@ -27,6 +29,7 @@
     AsmMatcher,
     AsmPattern,
     Replacement,
+    ReplacementPart,
```
simonlindholm, line 32: nit: unused

src/arch_ppc.py
```diff
@@ -409,13 +415,51 @@ def normalize_instruction(cls, instr: AsmInstruction) -> AsmInstruction:
...
+
+        def make_memory_access(arg: Argument) -> Access:
+            assert size is not None
+            if isinstance(arg, AsmAddressMode):
+                return MemoryAccess(
+                    base_reg=arg.rhs,
+                    offset=arg.lhs,
+                    size=size,
+                )
+            elif isinstance(arg, AsmGlobalSymbol):
```
simonlindholm, line 450: also need to support BinOp (symbol + offset). Should probably mirror mips's version of make_memory_access and run this if unconditionally

src/arch_ppc.py
```diff
@@ -430,45 +474,200 @@ def parse(
...
+                        len(args) == 3
+                        and isinstance(args[1], Register)
+                        and isinstance(args[2], Register)
+                    )
+                    inputs = [args[1], args[2], MemoryAccess(args[1], args[2], size)]
+                else:
+                    assert len(args) == 2 and isinstance(args[1], AsmAddressMode)
+                    inputs = [args[1].rhs, make_memory_access(args[1])]
+            elif mnemonic == "li" and isinstance(args[1], AsmAddressMode):
+                # `li $rD, sym@sda21(r2)` is equivalent to `addi $rD, $r2, sym@sda21`
```
simonlindholm, line 624: `li` can't read from memory, surely? looks from the implementation that we'd crash on that, too

simonlindholm (approved):
Went over arch_{ppc,mips}.py as well. It's all very boilerplatey and ad-hoc, and could definitely be cleaned up by strategic helper functions and such, but let's go with this until we refactor it all.

src/arch_mips.py
zbanks: translate.py doesn't track ra, but I thought it might be more consistent to still list it as an input here? Similarly, it'd be marked as an input in `lw    $ra, 0x14($sp)` or an output in `lw $ra, 0x14($sp)` -- despite also being ignored there too.

src/arch_mips.py
zbanks: There's a specific case for this in the `ida/handwritten.s` test. It has `la $v0, 4($v0)`:

```asm
glabel test
la $v0, symbol
lw $v0, ($v0)
la $v1, symbol + 8
lw $v1, 4($v1)

# same as addition
la $v0, 4($v0)
...
```

src/arch_ppc.py
zbanks: Ah, this was a quirk of the disassembler.
`li` is an alias for `addi`, so I thought this was a weird-but-valid form. However, `li $rD, sym@sda21(r2)` should have been disassembled as an *la* instruction, not *li*. 

[Reference, page 23](https://www.nxp.com/docs/en/application-note/AN2491.pdf)

I fixed the disassembler to emit `addi $rD, $r2, sym@sda21` instead of `li $rD, sym@sda21(r2)`, regenerated the test, and removed the above block of code. The test output doesn't change.

Although I could have changed it to `la`, and added support for `la` to PPC, but the decomp projects seem to all prefer the `addi` form instead of the `la` form

src/parse_instruction.py
```diff
@@ -143,6 +156,10 @@ class Instruction:
     args: List[Argument]
     meta: InstructionMeta
 
+    inputs: List[Access]
+    outputs: List[Access]
+    clobbers: List[Access]
```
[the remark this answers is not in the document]
zbanks: I added a comment to try to clarify overlapping, but you're right it's potentially confusing.

I'm worried that keeping outputs/clobbers disjoint will be a lot of work without much benefit? e.g. in ` narrow_func_call_outputs`, it would have to move the accesses from outputs to clobbers?

src/arch_mips.py
simonlindholm: I generally agree, but jalr writes to ra, it doesn't read it. And that write is then immediately clobbered by the function call.

zbanks: My bad! I was thinking about `jr $ra`! I'll remove it

PR #211, opened 2022-02-20
Handle stacks >=32kB from IDO

zbanks:
IDO generates a `lui`/`ori` pair to put the size in `$at` before subtracting it from `$sp`.

MWCC does not allow stacks to be >32kB, so all valid stacks should be able to be allocated in a single instruction. It emits the following error message:
```
### mwcceppc.exe Compiler:
#    File: tests\end_to_end\big-stack\orig.c
# ------------------------------------------
#      19: }
#   Error: ^
#   local data >32k
```

The IDO test had to be manually edited after it was generated; the disassembler marked the stack calculations with `%hi(D_123478)`/`%lo(D_123478)`/etc. relocs, which we don't expect mips2c to be able to convert back into literals.

The test outputs are still not perfect; mips2c still has trouble with the >0x8000 offsets, but it still seems like a worthwhile improvement.

This PR fixes [`mainProc.s`](https://gist.github.com/zbanks/9ef1cfb82da6fc26eb8bb9e79593a524) that was posted in Discord.


src/translate.py
```diff
@@ -488,6 +489,15 @@ def get_stack_info(
...
             assert isinstance(inst.args[2], AsmLiteral)
             info.allocated_stack_size = abs(inst.args[2].signed_value())
+        elif (
+            arch_mnemonic == "mips:subu"
+            and inst.args[0] == arch.stack_pointer_reg
+            and inst.args[1] == arch.stack_pointer_reg
+            and inst.args[2] == Register("at")
+            and at_value is not None
+        ):
+            # Moving the stack pointer more than 0xFFFF on MIPS
```
simonlindholm, line 499:
```suggestion
            # Moving the stack pointer more than 0x7FFF on MIPS
```

simonlindholm (approved):
I think you forgot to commit the test output.

The check in handle_addi_real that makes sure not to update regs["sp"] during addiu sp, sp, X is now insufficient. Probably doesn't actually matter given that we basically never read regs["sp"]. We could consider moving that logic to set_reg instead though. (At the risk of changing output of functions that use alloca, potentially. We should add a testcase for alloca!) No need to deal with this now if you don't feel like it, though worth a TODO comment if not.

zbanks:
> I think you forgot to commit the test output.

Man, I tried to be clever and add a commit containing *just* the test and it's original output, so that it was clear what the effect of the change was. Of course I then didn't notice the `push` without `-f` failed :facepalm:.

PR #212, opened 2022-03-10
Compute register usage as part of FlowGraph

zbanks:
This lays the groundwork for implementing "IR patterns", which need to (soundly) match reordered instructions.

If it's useful, my branch for doing IR pattern matching is [here](https://github.com/zbanks/mips_to_c/tree/ir-patterns). I'm not sure if we should just do one big PR off of that branch instead?

The changes here aren't intended to have any output diffs, but they do change:
- New `Warning: in ..., regs were read before being written to:` warnings.
    - I was unsure if these were worth adding at all: they usually aren't very useful to a decomper, but they are important to mips_to_c developers. 
    - Some were from unsound asm pattern replacements, like dropping the `"*"` instruction in `GccSqrtPattern`. (It's usually a `nop`, but sometimes it should be included)
- It's possible to get the state of argument/return registers incorrect by providing incorrect types in the context file. (e.g. marking a function `int foo(int);` as `void foo(void);`)
    - I had to change the `raise DecompError("Undeclared read...")` into a weaker `return ErrorExpr("Read from unset register...")` so that an incorrect context hint won't break an entire function.
- I added special error handling for `mtc0`. It now treats it as a no-dest instruction, instead of assuming that it writes to its first arg.
    - This changed the output of a few functions like `osUnmapTLBAll`

### DIFF
--- a/src/flow_graph.py
+++ b/src/flow_graph.py
@@ -1094,6 +1094,9 @@ class RefSet:
     def __iter__(self) -> Iterator[Reference]:
         return iter(self.refs)
 
+    def __len__(self) -> int:
+        return len(self.refs)
+
 
 @dataclass
 class AccessRefs:
@@ -1123,6 +1126,12 @@ class AccessRefs:
             self.refs[reg] = RefSet([ref])
         else:
             self.refs[reg].add(ref)
+
+    def extend(self, reg: Access, refs: RefSet) -> None:
+        if reg not in self:
+            self.refs[reg] = refs.copy()
+        else:
+            self.refs[reg].update(refs)
 
     def remove(self, reg: Access) -> None:
         self.refs.pop(reg)
--- a/src/flow_graph.py
+++ b/src/flow_graph.py
@@ -7,20 +7,22 @@ from typing import (
     Callable,
     Counter,
     Dict,
+    ItemsView,
     Iterator,
     List,
     Optional,
     Set,
     Tuple,
     Union,
+    ValuesView,
 )
 
 from .error import DecompFailure
 from .options import Formatter, Target
 from .parse_file import AsmData, Function, Label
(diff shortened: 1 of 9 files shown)


src/flow_graph.py
```diff
@@ -1096,15 +1253,157 @@ def reset_block_info(self) -> None:
...
+            for dep in deps:
+                if isinstance(dep, InstrRef):
+                    flow_graph.instr_references[dep].add(reg, ref)
+
+    if missing_regs:
+        print("/*")
+        print(f"Warning: in {function.name}, regs were read before being written to:")
+        for reg, ref in missing_regs:
+            print(f"   {reg} at {ref}: {ref.instruction()}")
+        print(f"*/")
```
zbanks, line 1391: As mentioned in the main body, I'm not sure if this is worth warning about or not. 
(I'm also not sure if there's a better wording here?)

src/flow_graph.py
```diff
@@ -1096,15 +1253,157 @@ def reset_block_info(self) -> None:
...
+                    sources.add(InstrRef(n, i))
+                    break
+                if reg in instr.clobbers:
+                    return RefSet.invalid()
+        else:
+            stack.extend(n.parents)
+
+    if not sources.is_valid():
+        return imdom_srcs
+    return sources
```
zbanks, line 1288: `phi_reg_sources` is based on the old `reg_always_set`, but returns the `RefSet` of the instructions that set the given reg (instead of just a boolean value).

In general, I'm also unsure if it's OK to use the naming convention `reg: Access`? I feel like "register" is a more concrete concept than "access" even if it's technically too specific? All these functions work just as well with `MemoryAccess`s as `Register`s -- I'd be happy to change/redocument anything if you have suggestions for making this clearer.

PR #214, opened 2022-03-17
IR pattern matching & register flow analysis

zbanks:
This PR uses the instruction input/output/clobber access sets to compute a dependency/usage graph before the translate stage.
This is then used to do IR pattern matching, to match the PPC float conversion patterns even when the instructions are reordered, reused, or split across multiple basic blocks.

It's really tempting to bring in a graph utility library like [networkx](https://networkx.org/), but I don't think it's worth the dependency yet. 

The biggest caveat is that the IR pattern-matching algorithm is somewhat naive and can be slow: it's technically `O(n^m)` with `n` being the input asm length, and `m` being the pattern length. Practically, it about ~doubles the runtime in the worst case for a few functions in the PPC test projects I was looking at. I think this is a fine tradeoff for now?

Aside from the PPC patterns, the changes here aren't intended to have any significant output diffs, but they do change:
- New `Warning: in ..., regs were read before being written to:` warnings.
    - I was unsure if these were worth adding at all: they usually aren't very useful to a decomper, but they are important to mips_to_c developers. 
    - Some were from unsound asm pattern replacements, like dropping the `"*"` instruction in `GccSqrtPattern`. (It's usually a `nop`, but sometimes it should be included)
- It's possible to get the state of argument/return registers incorrect by providing incorrect types in the context file. (e.g. marking a function `int foo(int);` as `void foo(void);`)
    - I had to change the `raise DecompError("Undeclared read...")` into a weaker `return ErrorExpr("Read from unset register...")` so that an incorrect context hint won't break an entire function.
- I added special error handling for `mtc0` in MIPS. It now treats it as a no-dest instruction, instead of assuming that it writes to its first arg.
    - This changed the output of a few functions like `osUnmapTLBAll`


src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
...
+                    sources.add(InstrRef(n, i))
+                    break
+                if reg in instr.clobbers:
+                    return RefSet.invalid()
+        else:
+            stack.extend(n.parents)
+
+    if not sources.is_valid():
+        return imdom_srcs
+    return sources
```
zbanks, line 1309: `phi_reg_sources` is based on the old `reg_always_set`, but returns the `RefSet` of the instructions that set the given reg (instead of just a boolean value).

In general, I'm also unsure if it's OK to use the naming convention `reg: Access`? I feel like "register" is a more concrete concept than "access" even if it's technically too specific? All these functions work just as well with `MemoryAccess`s as `Register`s -- I'd be happy to change/redocument anything if you have suggestions for making this clearer.

src/arch_ppc.py
```diff
@@ -894,6 +983,11 @@ def make_memory_access(arg: Argument) -> Access:
         # `sp100 = (bitwise f64) (s32) x; y = sp104;` instead of `y = (s32) x;`.
         # We should try to detect these idioms, along with int-to-float
         "fctiwz": lambda a: handle_convert(a.reg(1), Type.s32(), Type.floatish()),
+        "cvt.i.d.ficitve": lambda a: handle_convert(a.reg(1), Type.s32(), Type.f64()),
+        "cvt.d.i.fictive": lambda a: handle_convert(a.reg(1), Type.f64(), Type.s32()),
+        "cvt.d.u.fictive": lambda a: handle_convert(a.reg(1), Type.f64(), Type.u32()),
+        "cvt.s.i.fictive": lambda a: handle_convert(a.reg(1), Type.f32(), Type.s32()),
+        "cvt.s.u.fictive": lambda a: handle_convert(a.reg(1), Type.f64(), Type.u32()),
```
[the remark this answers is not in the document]
zbanks: I added `Type.sintish()` & `Type.uintish()` & changed the types here.

src/arch_ppc.py
```diff
@@ -173,6 +177,71 @@ def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
...
+
+
+class CheckConstantMixin:
+    def check(self, m: IrMatch) -> bool:
+        # TODO: Also validate that `K($k)` is the expected constant in rodata
+        return m.symbolic_registers["k"] in m.arch.constant_regs
+
+
+class SintToDoubleIrPattern(IrPattern, CheckConstantMixin):
+    replacement = "cvt.d.i.fictive $f, $i, K($k)"
```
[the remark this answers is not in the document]
zbanks: This is a constraint on the pattern `parts`: because there's an instruction that reads from `K($k)` in the pattern, `K($k)` must be listed as one of the `inputs` to the replacement. 

When doing pattern replacement, it's possible to elide `outputs` by moving them into `clobbers`, this is what's done to the `$a`, `$b`, `N($r1)`, etc. terms. 
There isn't an equivalent concept for `inputs`. Allowing you to elide `inputs` would affect the soundness of the replacement, *and* require additional code to support.

src/flow_graph.py
```diff
@@ -1013,10 +1032,169 @@ def terminate_infinite_loops(nodes: List[Node]) -> None:
...
+    def replace_instruction(self, instr: Instruction) -> None:
+        self.node.block.instructions[self.index] = instr
+
+    def __repr__(self) -> str:
+        line = self.instruction().meta.lineno
+        base = f"{self.node.block.index}.{self.index}"
+        return f"{base} (line {line})" if line else base
+
+
+Reference = Union[InstrRef, str]
```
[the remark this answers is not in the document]
simonlindholm: fleeting thought: I wonder if this should be folded into InstrRef. `str` kinda represents a reference to a pseudo-instruction at the very start that sets up constant registers. We could make that instruction explicit even, though I suspect it's not a good idea

src/ir_pattern.py
```diff
@@ -0,0 +1,443 @@
...
+                    flow_graph.instr_references[r].remove_ref(ref)
+
+        # Parse the asm & set the clobbers
+        new_instr = arch.parse(new_asm.mnemonic, new_asm.args, instr.meta.derived())
+        new_instr.clobbers.extend(
+            acc for acc in instr.outputs if acc not in new_instr.outputs
+        )
+        new_instr.clobbers.extend(
+            acc for acc in instr.clobbers if acc not in new_instr.clobbers
+        )
```
simonlindholm, line 335: we're clobbering a lot of registers here. What if they are read from? That should make us not do the replacement, right?

src/ir_pattern.py
```diff
@@ -0,0 +1,443 @@
...
+        assert isinstance(pattern.flow_graph.nodes[1], TerminalNode)
+        pattern_node = pattern.flow_graph.nodes[0]
+
+        # Perform a brute force-ish graph search to find candidate sets of instructions
+        # that match the pattern with the correct dependencies & arguments. This will
+        # have poor performance with large patterns that use common mnemonics.
+        partial_matches = [TryIrMatch(arch=arch)]
+        for i, pat_instr in enumerate(pattern_node.block.instructions):
+            if pat_instr.mnemonic == "nop":
+                continue
```
simonlindholm, line 358: huh? why do nops even show up in patterns?

src/ir_pattern.py
```diff
@@ -0,0 +1,443 @@
...
+                        continue
+                    if not state.match_accessrefs(
+                        pat_inputs, flow_graph.instr_inputs[cand_ref]
+                    ):
+                        continue
+                    next_partial_matches.append(state)
+
+            partial_matches = next_partial_matches
+            if not partial_matches:
+                break
```
simonlindholm, line 381: fwiw I don't think this optimization helps much

src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
             node.block.block_info = None
 
 
+def phi_reg_sources(node: Node, reg: Access, imdom_srcs: RefSet) -> RefSet:
```
simonlindholm, line 1277: So, we're presuming that `reg` references a static location here, e.g. a register, a stack slot, or a global variable. (But we shouldn't track global variables as phis and imo we shouldn't do alias analysis at this level so let's forget about the last point.) But Access is more general than that -- it supports "reg + whatever" -- and using that here would be incorrect. I feel like we should drop that generalized representation and change to a Union[Register, StackRef(int), MemoryAccess]. (sp + non constant is a MemoryAccess, similar to how computing sp + non constant with an add instruction and then referencing the resulting memory would work. We can relatively safely assume no aliasing I think)

simonlindholm (commented):
Some continued review (some of it probably too low level to actually be valuable feedback right now...)

src/arch_ppc.py
```diff
@@ -890,10 +993,25 @@ def make_memory_access(arg: Argument) -> Access:
         "fneg": lambda a: UnaryOp(op="-", expr=a.reg(1), type=Type.floatish()),
         "fmr": lambda a: a.reg(1),
         "frsp": lambda a: handle_convert(a.reg(1), Type.f32(), Type.f64()),
-        # TODO: This yields some awkward-looking C code, often in the form:
-        # `sp100 = (bitwise f64) (s32) x; y = sp104;` instead of `y = (s32) x;`.
-        # We should try to detect these idioms, along with int-to-float
-        "fctiwz": lambda a: handle_convert(a.reg(1), Type.s32(), Type.floatish()),
+        "fctiwz": lambda a: handle_convert(a.reg(1), Type.sintish(), Type.floatish()),
```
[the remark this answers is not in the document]
zbanks: As far as I can tell playing around on decomp.me with different compiler settings, it seems like `fctiwz` is used for either floats or doubles. ([Example scratch](https://decomp.me/scratch/Anc5r))

[This compiler-writer PDF](https://www.warthman.com/images/IBM_PPC_Compiler_Writer%27s_Guide-cwg.pdf), page 100, uses `fctiwz` for both float/double, though for double it uses a `lwa` instead of `lwz` that I'm unable to reproduce.

src/flow_graph.py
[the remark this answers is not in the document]
zbanks: I did go back-and-forth on folding `str` into `InstrRef` or not (at which point it would just be called `Reference`). I decided to represent these "constant registers" with a bare `str` for two reasons: 
- Python has poor support for sum types. It's awkward to make every field in the dataclass `Optional[...]` and frequently assert that the fields you expect are not `None`.
- There are times when I want to make it explicit that I have an *instruction* reference, and I expect  that the `.instruction()` method should return a valid `Instruction`. (The constant registers have nothing to return there)

I also decided to use a bare `str` instead of a complex dataclass/enum because the values are pretty static and aren't analyzed further. (And as said before, it's very cumbersome to define/use sum types.)

src/ir_pattern.py
zbanks: There is a check near line 426 for when the instruction is part of the pattern but will not be replaced (because its outputs are referenced by other non-pattern instructions): if the instruction would clobber an input, then the entire replacement is cancelled.
```python
                elif any(r in pattern_inputs for r in instr.clobbers + instr.outputs):
                    # If this instruction can't be replaced, but it clobbers a needed
                    # input register, then we can't perform the rewrite
```

However, in the more common case, I think there is an issue here: if the instruction clobbers an input register *and* is being replaced, the replaced `nop` should not be marked as clobbering the register. I think the current code is incorrect, but still behaves correctly because the register flow graph is not recomputed from scratch after the rewrites.

src/ir_pattern.py
zbanks: I use fake `nop` instructions in `IrPattern.compile` to act as placeholder references for inputs & outputs. 
I could make this more clear by using different mnemonic name(s), but I would need to add them to `MipsArch`/`PpcArch`?

src/flow_graph.py
```diff
@@ -1013,10 +1032,169 @@ def terminate_infinite_loops(nodes: List[Node]) -> None:
...
 class FlowGraph:
     nodes: List[Node]
 
+    # The set of all phi Accesses needed for each Node
+    node_phis: Dict[Node, AccessRefs] = field(default_factory=dict)
+    # The set of source References for each input Access to each Instruction
+    instr_inputs: Dict[InstrRef, AccessRefs] = field(default_factory=dict)
+    # The set of all downstream Instructions whose inputs depend on outputs from a given
+    # Instruction. This is the same information as `instr_inputs`, but reversed.
+    instr_references: Dict[InstrRef, AccessRefs] = field(default_factory=dict)
```
simonlindholm, line 1196: I imagine you considered making these properties of Node and Instruction respectively, but it didn't work out?

src/flow_graph.py
simonlindholm: (Also, re my comments about -framepointer from before that turn fp into an alias for sp, we should just do that analysis in an earlier pass than translate to the extent that we care about supporting it)

src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
             node.block.block_info = None
 
 
+def phi_reg_sources(node: Node, reg: Access, imdom_srcs: RefSet) -> RefSet:
+    """
+    Return the RefSet of all of the places `reg` is assigned to, if it is a valid phi.
+    Otherwise, return RefSet.invalid().
+    """
+    if node.immediate_dominator is None:
```
simonlindholm, line 1282: Should be possible to assert that this is not None, given the caller. (If it *is* None I think the right answer might be to include references to arguments etc. rather than return invalid? but we don't have to think about that)

simonlindholm: Completely unrelated, but I just realized we're assuming everywhere that the entry node has no parents/dominators, but I don't think we ever make sure that that's the case? (unless I missed some check in my cursory review) We might want to add an assert for that. (Or more ambitiously, inject an empty initial block if the first block has parents.)

src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
...
+        return RefSet.invalid()
+
+    seen = {node.immediate_dominator}
+    stack = node.parents[:]
+    sources = RefSet()
+
+    while stack:
+        n = stack.pop()
+        if n == node.immediate_dominator and not imdom_srcs.is_valid():
+            return RefSet.invalid()
```
simonlindholm, line 1292: if we reach immediate_dominator, don't we want to merge imdom_srcs into sources? (rather than having the `sources.is_valid()` check at the end of this function, which misses cases)

src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
...
+        else:
+            stack.extend(n.parents)
+
+    if not sources.is_valid():
+        return imdom_srcs
+    return sources
+
+
+def regs_clobbered_until_dominator(node: Node) -> Set[Access]:
+    if node.immediate_dominator is None:
```
simonlindholm, line 1313: assert not None

src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
...
+            for dep in deps:
+                if isinstance(dep, InstrRef):
+                    flow_graph.instr_references[dep].add(reg, ref)
+
+    if missing_regs:
+        print("/*")
+        print(f"Warning: in {function.name}, regs were read before being written to:")
+        for reg, ref in missing_regs:
+            print(f"   {reg} at {ref}: {ref.instruction()}")
+        print(f"*/")
```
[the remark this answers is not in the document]
simonlindholm: I loosely feel like since instruction inputs are kinda advisory anyway, given function calls, we can leave handling these kinds of errors to the translation phase. But it's not a strong opinion, warning here is fine too.

src/ir_pattern.py
```diff
@@ -0,0 +1,443 @@
...
+            acc for acc in instr.clobbers if acc not in new_instr.clobbers
+        )
+
+        # Replace the instruction in the block
+        ref.node.block.instructions[ref.index] = new_instr
+
+    for pattern_class in pattern_classes:
+        pattern = pattern_class.compile(arch)
+
+        # For now, patterns can't have branches: they should only have 2 Nodes,
```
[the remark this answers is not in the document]
zbanks: I agree it'd be a lot of extra work/complexity. I'm not particularly looking to implement this.

A lot of existing MIPS asm patterns involve branches, though. The patterns work pretty well but there are some instances with GCC that fail to match. I think it's fine to leave as-is, but those could be motivating examples in the future.

src/flow_graph.py
zbanks: I moved `node_phis` into `Node`, but I don't know about moving the other fields into `Instruction`.
(Right now `InstrRef` references `Reference` which references `Node`, so at a minimum, there's that dependency chain to flatten)

src/flow_graph.py
zbanks: Yeah you're right -- the existing code assumes the entry node has no parents. We should have to set up phis for the entry node if it has parents. 

I don't think the entry node can ever have an explicit dominator though? Its caller always acts as an implicit parent/immediate dominator?

src/flow_graph.py
```diff
@@ -1359,50 +1348,60 @@ def process_node(node: Node, reg_srcs: AccessRefs) -> None:
...
 
-            # Remove any clobbered accesses, or any MemoryAccesses that depend on
+            # Remove any clobbered locations, or any MemoryLocations that depend on
             # clobbered Registers. Ex: If `$v0` is clobbered, remove `$v0` and `4($v0)`
             for clob in ir.clobbers + ir.outputs:
-                for reg in reg_srcs:
-                    if access_depends_on(reg, clob) or access_may_overlap(reg, clob):
-                        reg_srcs.remove(reg)
+                for loc in loc_srcs:
+                    if location_depends_on(loc, clob) or locations_alias(loc, clob):
```
simonlindholm, line 1364: do we still need the depends_on behavior? it should now only cover invalidating sp+offset when sp is updated, but I think we can assume that we won't have sp accesses with different values of sp within a function. (If we encounter arches/compilers that do we should still be able to resolve offsets statically in  a separate pass.)

zbanks: Good catch! 
(I agree that we can probably assume that `$sp` is consistent within a function, but it also wouldn't be hard to add a check to cover it anyways)

src/ir_pattern.py
```diff
@@ -0,0 +1,465 @@
...
+            for i, ref in enumerate(refs_to_replace):
+                # Remove ref from all instr_uses
+                instr = ref.instruction()
+                for _, rs in flow_graph.instr_inputs[ref].items():
+                    for r in rs:
+                        if isinstance(r, InstrRef):
+                            flow_graph.instr_uses[r].remove_ref(ref)
+
+                # The last instruction in the source is rewritten with the replacement instruction
+                # (refs_to_replace is in reverse order), and the others are replaced with a nop.
```
simonlindholm, line 411: Can it happen that replacing the last instruction results in a read from a totally different input? Say we have the pattern
```
mov $t, $i
mov $o, $t
->
mov $o, $i
```
and match that against the asm
```
mov $t, $i
li $i, 42
mov $o, $t
```
resulting in
```
nop
li $i, 42
mov $o, $i
```
Like in the comment above, new fictive temp registers or using the in/out graph in translate.py would help...

src/ir_pattern.py
```diff
@@ -0,0 +1,443 @@
...
+        assert (
+            len(pattern.flow_graph.nodes) == 2
+        ), "branching patterns not yet supported"
+        assert isinstance(pattern.flow_graph.nodes[0], BaseNode)
+        assert isinstance(pattern.flow_graph.nodes[1], TerminalNode)
+        pattern_node = pattern.flow_graph.nodes[0]
+
+        # Perform a brute force-ish graph search to find candidate sets of instructions
+        # that match the pattern with the correct dependencies & arguments. This will
+        # have poor performance with large patterns that use common mnemonics.
```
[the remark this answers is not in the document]
simonlindholm: Mmm, true, that does seem annoying. :( It would be a nice way to recover perf though, I am kinda bothered by the current exponential approach.

src/flow_graph.py
```diff
@@ -1013,10 +1032,169 @@ def terminate_infinite_loops(nodes: List[Node]) -> None:
...
+        return f"{base} (line {line})" if line else base
+
+
+Reference = Union[InstrRef, str]
+
+
+@dataclass
+class RefSet:
+    """
+    An ordered set of References, with an empty set representing an invalid/missing state.
```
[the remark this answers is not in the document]
zbanks: As of the latest refactor, empty sets are no longer used to represent "invalid" -- let me know if this is still ambiguous.

src/flow_graph.py
```diff
@@ -1013,10 +1032,169 @@ def terminate_infinite_loops(nodes: List[Node]) -> None:
...
+
+    def __iter__(self) -> Iterator[Reference]:
+        return iter(self.refs)
+
+    def __len__(self) -> int:
+        return len(self.refs)
+
+
+@dataclass
+class AccessRefs:
```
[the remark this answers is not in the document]
zbanks: Renamed this to `LocationRefSetDict` to make it more of a generic data structure, as it's used in two distinct ways. 
(I tried to add documentation describing its motivation to where it's actually used.)

src/ir_pattern.py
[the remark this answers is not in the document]
zbanks: It's now backwards in the latest approach. 
There's still some "mild" "exponential" behavior in `permute_and_match_refset` to handle the situation I described above, but now it's slightly bounded by the *pattern* and not the input assembly. (For the existing patterns, it's only `O(n^2)` at worst)

src/ir_pattern.py
```diff
@@ -0,0 +1,465 @@
...
+        asm_data = AsmData()
+        flow_graph = build_flowgraph(func, asm_data, arch, fragment=True)
+        return cls(
+            flow_graph=flow_graph,
+            replacement_instr=replacement_instr,
+        )
+
+
+@dataclass
+class IrMatch:
```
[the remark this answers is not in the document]
zbanks: I could experiment with trying to share some code between the two once the rest of the PR looks good? I think it'd get too confusing if I try to refactor it now.

src/ir_pattern.py
```diff
@@ -0,0 +1,446 @@
...
+            return pat == cand
+        return self._match_var(self.ref_map, pat, cand)
+
+    def permute_and_match_refset(self, pat: RefSet, cand: RefSet) -> List["TryIrMatch"]:
+        """
+        Return a list of all possible TryIrMatch states, where every ref in `pat` is
+        matched to a unique ref in `cand`. This may modify the original TryIrMatch object.
+        Although this function is technically exponential in the size of `pat`, it
+        usually only contains a single ref.
+        """
```
simonlindholm, line 254: You said something about this only being needed to support overlapping stack accesses, right? I think clarifying that in the comment would be good. (I also don't fully get why it's needed: naively I'm thinking that in `write4 n(sp)`, `write4 (n+4)(sp)`, `read8 n(sp)` it would read n(sp) from the tested starting instruction, deduce n, evaluate n and n+4, and then find the instructions that write to n and n+4 respectively.)

simonlindholm (commented):
Still slowly working my way through this. Slow going because pattern matching logic is dense

src/ir_pattern.py
```diff
@@ -0,0 +1,446 @@
...
+    @classmethod
+    def compile(cls, arch: ArchFlowGraph) -> "IrPattern":
+        missing_meta = InstructionMeta.missing()
+        regf = RegFormatter()
+        replacement_instr = parse_instruction(cls.replacement, missing_meta, arch, regf)
+
+        name = f"__pattern_{cls.__name__}"
+        func = Function(name=name)
+        # Add a fictive nop instruction for each input to the replacement_instr
+        # This acts as a placeholder Reference to represent where the input was set
```
simonlindholm, line 74: Can you assert that these are all unique, here or elsewhere. It's implicitly assumed later on

src/ir_pattern.py
```diff
@@ -0,0 +1,446 @@
...
+                    f"{original_reg.register_name}_fictive_{replace_index}"
+                )
+                state.rename_reg(input_reg, temp_reg)
+                move_instr = arch.parse(
+                    "move.fictive", [temp_reg, original_reg], InstructionMeta.missing()
+                )
+                input_uses = pattern.flow_graph.instr_uses[pat_ref].get(input_reg)
+                assert len(input_uses) >= 1
+                for use in input_uses:
+                    state.map_ref(use).add_instruction_before(move_instr)
```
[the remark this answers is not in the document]
simonlindholm: Needs to update `instr_uses`/`instr_inputs` as well

src/ir_pattern.py
```diff
@@ -0,0 +1,446 @@
...
+
+            for i, ref in enumerate(refs_to_replace):
+                # Remove ref from all instr_uses
+                instr = ref.instruction
+                for loc in instr.inputs:
+                    for r in flow_graph.instr_inputs[ref].get(loc):
+                        if isinstance(r, InstrRef):
+                            flow_graph.instr_uses[r].remove_ref(ref)
```
simonlindholm, line 423: and clear out instr_inputs[ref] to maintain graph correctness, and add input edges for the final instruction pointing to the in.fictive's

simonlindholm: can also assert that instr_uses[ref] is empty, except for i == 0. For i == 0 I believe instr_uses is kept? (but I'm not 100% sure. I don't remember seeing asserts for outputs being the same)

simonlindholm: instr_uses seems super easy to screw up when dynamically updating the graph. I wonder if we can add some error checking to it. Some ideas:
- recompute it from scratch and check that didn't change. Can be done until we switch translate.py over to the graph based info and make that the canonical source of truth. (heavy-weight option, annoying, slow)
- check that instr_inputs/instr_uses match up
- check that instr_inputs/instr_uses don't point to removed nodes/nops
- check that instr_inputs is non-empty if we have input registers (if that's even true)

simonlindholm: (I wish github allowed multiple review comment threads to hang off the same source line...)

src/arch_ppc.py
```diff
@@ -661,17 +764,31 @@ def make_memory_access(arg: Argument) -> Access:
...
+            reads_memory=reads_memory,
+            writes_memory=writes_memory,
             jump_target=jump_target,
             function_target=function_target,
             is_conditional=is_conditional,
             is_return=is_return,
         )
 
+    def simplify_ir(self, flow_graph: FlowGraph) -> None:
+        simplify_ir_patterns(self, flow_graph, self.ir_patterns)
```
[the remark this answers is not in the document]
zbanks: I can't move this to translate.py, if I did then it couldn't be called from flow_graph.py in `build_flowgraph`.
I don't want to call it from `translate_to_ast`, because that function can be called multiple times for the same flow graph.

I could leave it in ir_patterns.py, and call `simplify_ir_patterns` from main.py?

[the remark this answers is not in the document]
simonlindholm: I was thinking keeping the declaration in flow_graph.py, but defining it in Arch in translate.py. Moving to main is also an option

src/ir_pattern.py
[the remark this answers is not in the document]
zbanks: I don't think this works. With the following pattern:
```python
    # inputs: $a, $b; output: $x
    replacement = "foo $x, $a, $b"
    parts = [
        "addis $p, $a, $b",
        "xoris $q, $a, $b",
        "subf $x, $p, $q",
    ]
```

This could match instructions in either `addis; xoris; subf` or `xoris; addis; subf` order in the original asm. And this move needs to come before whichever is first in the control flow.

I don't think `input_uses[0]` is guaranteed to come "first" -- at least, it's not an intended guarantee of `RefSet`.

I can try to make this more clear with a comment

src/parse_instruction.py
```diff
@@ -116,18 +117,83 @@ def arbitrary() -> "MemoryAccess":
...
+    if isinstance(base, Register):
+        return False
+    elif isinstance(base, MemoryAccess):
+        if isinstance(dep, MemoryAccess):
+            return False
+        if base.base_reg == dep:
+            return True
+        if isinstance(base.offset, AsmAddressMode) and base.offset.rhs == dep:
+            return True
+        if isinstance(base.offset, Macro) and base.offset.argument == dep:
```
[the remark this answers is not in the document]
zbanks: No, I removed this check.

src/ir_pattern.py
```diff
@@ -0,0 +1,446 @@
...
+                # pattern, like random nops.
+                cand = state.try_map_ref(pat_ref)
+                if not isinstance(cand, InstrRef):
+                    continue
+                if not state.match_instr(pat_ref.instruction, cand.instruction):
+                    continue
+                states = state.permute_and_match_inputrefs(
+                    pat_inputs,
+                    flow_graph.instr_inputs[cand],
+                )
```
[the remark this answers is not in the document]
zbanks: I think my plan for fixing this is to create special RefSets to represent phi registers with "the value of register X at the start of node Y". 

This makes it so that, for matching purposes, pattern inputs are matched just like internal registers.

src/flow_graph.py
```diff
@@ -1096,15 +1274,104 @@ def reset_block_info(self) -> None:
...
+            for loc in locs_clobbered_until_dominator(child):
+                child_loc_srcs[loc] = RefSet.phi(loc, child.block)
+
+            process_node(child, child_loc_srcs)
+
+    # Set all registers to come from an unspecified source at the start of the entry node
+    entry_node = flow_graph.entry_node()
+    entry_reg_srcs = LocationRefSetDict()
+    for r in arch.all_regs:
+        entry_reg_srcs.refs[r] = RefSet.phi(r, entry_node.block)
```
zbanks, line 1343: I mentioned this on Discord, but I was wondering if there was a better term for `PhiRef` that made it clear it also covered inputs to functions? 

It'd actually be OK to use a unique singleton here -- but since I'm using `PhiRef` to indicate the "the initial value of `loc` at the start of `node`" it doesn't seem too wild to also use it here?

src/ir_pattern.py
zbanks: I refactored all the `instr_uses`/`instr_inputs` manipulation in a way that I think makes it easier to verify that it's correct, and I added the 2nd option you listed.

Now, whenever `instr_uses` is modified, `instr_uses` is also updated shortly afterwards. 

The assertion at the end seems to work: if I comment out some of the instr_uses/instr_inputs update statements, it trips the assert.

src/ir_pattern.py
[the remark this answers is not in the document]
simonlindholm: No, I think it should work: the new fictive register is only read by the final instruction. A comment is probably a good idea in any case.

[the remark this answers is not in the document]
zbanks: You're right; I refactored it to only insert one move instruction per temp.

src/arch_ppc.py
```diff
@@ -667,6 +745,17 @@ def make_memory_access(arg: Argument) -> Access:
...
+
+    ir_patterns: List[typing.Type[IrPattern]] = [
```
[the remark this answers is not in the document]
zbanks: I introduced `CompiledIrPattern` (I didn't mind the long name since it's local).
It's not fully symmetric with `asm_pattern.py`, but it's closer.

zbanks:
I'm also happy to refactor `IrMatch`/`TryIrMatch` with `TryMatchState` from asm_pattern.py to reduce code duplication -- but I wanted to wait until the rest of the changes were mostly stable.

src/flow_graph.py
```diff
@@ -1059,9 +1059,17 @@ def terminate_infinite_loops(nodes: List[Node]) -> None:
...
 
 
-# Reference acts as a pointer to either a specific Instruction, or a special non-instruction
-# source represented by a str, such as a function argument or constant register
-Reference = Union[InstrRef, str]
+@dataclass(frozen=True)
+class PhiRef:
+    """Represents the initial value of a Location at the start of a Block."""
+
+    location: Location
```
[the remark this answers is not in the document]
zbanks: `PhiRef`/`BlockEntryRef` no longer exist; so there's no longer an asymmetry.

src/flow_graph.py
```diff
@@ -1013,10 +1060,150 @@ def terminate_infinite_loops(nodes: List[Node]) -> None:
...
+
+
 @dataclass(frozen=True)
 class FlowGraph:
     nodes: List[Node]
 
+    # The set of source References for each input Location to each Instruction
+    instr_inputs: Dict[InstrRef, LocationRefSetDict] = field(default_factory=dict)
+    # The set of all downstream Instructions whose inputs depend on outputs from a given
+    # Instruction. This is the same information as `instr_inputs`, but reversed.
```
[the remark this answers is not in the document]
zbanks: These now include all forms of `Reference` (and I've rewritten these comments)

src/flow_graph.py
```diff
@@ -1301,56 +1382,87 @@ def locs_clobbered_until_dominator(node: Node) -> Set[Location]:
...
+                    # Otherwise, this is likely undefined behavior in the original asm
+                    if ir.function_target is not None and inp in arch.argument_regs:
+                        ir.inputs.remove(inp)
+                    else:
+                        missing_regs.append((inp, ref))
+                for src in sources:
+                    flow_graph.add_dependency(ref, inp, src)
+
+            # Remove any clobbered locations, or any MemoryLocations that depend on
+            # clobbered Registers. Ex: If `$v0` is clobbered, remove `$v0` and `4($v0)`
```
simonlindholm, line 1419: stale comment

src/flow_graph.py
```diff
@@ -1301,56 +1382,87 @@ def locs_clobbered_until_dominator(node: Node) -> Set[Location]:
...
+        # final register flow_graph. This will eventually reach every node.
         for child in node.immediately_dominates:
-            # Create phi refsets for the loc_srcs map for this child
             child_loc_srcs = loc_srcs.copy()
+
             for loc in locs_clobbered_until_dominator(child):
-                child_loc_srcs[loc] = RefSet.block_entry(loc, child.block)
+                phi_reg_srcs = phi_loc_sources(child, loc, loc_srcs.get(loc))
+                if phi_reg_srcs is not None:
+                    # If phi_reg_srcs is empty, then the loc is inconsistently set, so it should
```
simonlindholm, line 1438: empty vs None inconsistency in this comment

src/flow_graph.py
```diff
@@ -1096,15 +1274,163 @@ def reset_block_info(self) -> None:
...
+            for inp in ir.inputs:
+                sources = RefSet()
+                for reg, srcs in reg_srcs.items():
+                    if access_must_overlap(reg, inp):
+                        sources.update(srcs)
+                if isinstance(inp, Register) and not sources.is_valid():
+                    # Registers must be written to before being read.
+                    # If the instruction is a function call and we don't have a source
+                    # for the argument, we can prune the argument from the input list.
+                    # Otherwise, this is likely undefined behavior in the original asm
```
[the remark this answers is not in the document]
simonlindholm: so I *think* I'd be inclined to leave `sources` as an empty set in this case, postponing the error until translate time, and removing print_warnings. Empty `sources` will be treated by other graph analysis code as "materialized the input from thin air" which is fine, and we can test for invalidity by checking whether it's empty as needed.

(Edit: oh, it's just for --debug. That's fine, then. Still skeptical of `ir.inputs.remove(inp)` though.)

src/flow_graph.py
```diff
@@ -1301,56 +1382,87 @@ def locs_clobbered_until_dominator(node: Node) -> Set[Location]:
...
-    entry_node = flow_graph.entry_node()
+        if isinstance(node, TerminalNode):
+            assert not node.block.instruction_refs
+            for inp in arch.all_return_regs:
+                sources = RefSet()
+                for loc, srcs in loc_srcs.items():
+                    if locations_alias(loc, inp):
+                        sources.update(srcs)
+                for src in sources:
+                    flow_graph.add_dependency("epilogue", inp, src)
```
simonlindholm, line 1456: same as above, would be nice to use a common helper

src/flow_graph.py
```diff
@@ -1301,56 +1382,87 @@ def locs_clobbered_until_dominator(node: Node) -> Set[Location]:
...
+
+    # Set all the registers that are valid to access at the start of a function
     entry_reg_srcs = LocationRefSetDict()
     for r in arch.all_regs:
-        entry_reg_srcs.refs[r] = RefSet.block_entry(r, entry_node.block)
+        entry_reg_srcs.refs[r] = RefSet.special("prologue")
 
-    # Recursively traverse every node, starting with the entry node to populate instr_inputs
+    # Recursively traverse every node, starting with the entry node
+    # This populates in node.block.phis and instr_inputs
```
simonlindholm, line 1464: stale

src/ir_pattern.py
```diff
@@ -0,0 +1,452 @@
...
+                    "in.fictive",
+                    [],
+                    meta=missing_meta,
+                    inputs=[],
+                    clobbers=[],
+                    outputs=[inp],
+                    reads_memory=False,
+                    writes_memory=False,
+                )
+            )
```
simonlindholm, line 87: Do we really need in.fictive instructions, or can we use the inputs in place of/addition to `all_registers` when creating the pattern use/def graph?

My feeling is that it should be fine to replace all in.fictive instructions by a single one, since all dependency maps are keyed by location, and then replace the single remaining instruction by nothing since we know use/def-analyze even the entry state.

src/parse_instruction.py
```diff
@@ -200,15 +204,24 @@ def __str__(self) -> str:
...
 class InstructionMeta:
+    # True if the original asm line was marked with a goto pattern
     emit_goto: bool
+    # Asm source filename & line number
     filename: str
     lineno: int
+    # True if the Instruction is not directly from the source asm
     synthetic: bool
+    # True if the Instruction was part of a matched IR pattern, but not elided
+    in_pattern: bool
```
simonlindholm, line 215: I'm a bit hesistant to put this in InstructionMeta, which is right now kinda best-effort for debugging purposes. Would it work to put the bit in Instruction itself?

src/translate.py
```diff
@@ -4161,6 +4177,8 @@ def process_instr(instr: Instruction) -> None:
                     data.meta.inherited or data.meta.function_return
```
simonlindholm, line 4177: extend comment above to mention in_pattern in addition to function_return (so note that there'll be trouble if we rely on this mechanism for mips)

src/ir_pattern.py
```diff
@@ -184,6 +183,11 @@ def map_location(self, key: Location) -> Location:
             return loc
         static_assert_unreachable(key)
 
+    def map_instruction(self, key: Instruction, arch: ArchFlowGraph) -> Instruction:
+        return arch.parse(
+            key.mnemonic, [self.map_arg(a) for a in key.args], key.meta.derived()
```
simonlindholm, line 188: we should use the final original instruction to derive meta, not the pattern

src/ir_pattern.py
```diff
@@ -0,0 +1,452 @@
...
+            for state in try_matches:
+                # By pattern construction, pat_ref should be in the state's ref_map
+                # (It would be missing for "disjoint" or irrelevant instructions in the
+                # pattern, like random nops: these aren't allowed)
+                cand = state.try_map_ref(pat_ref)
+                if not isinstance(cand, InstrRef):
+                    continue
+                if not state.match_instr(pat_ref.instruction, cand.instruction):
+                    continue
+                states = state.permute_and_match_inputrefs(
```
simonlindholm, line 379: With the new meaning of instr_inputs, this is wrong: in case of phis we may end up matching the pattern input to one of the phi inputs at random. We need to somehow distinguish between phis and overlapping stack locations, rejecting phis while permuting on overlapping locations. Tracking each stack byte/word separately would help for this.

simonlindholm: Another, hackier, option: mark each instruction with whether it has phis or not. For ir_pattern.py that's enough information, and for translate.py (once we start using the use/def graph) it will also be, because it doesn't use StackLocation so it doesn't have to consider overlaps as possible multi-input sources.

src/flow_graph.py
```diff
@@ -1281,6 +1290,78 @@ def reset_block_info(self) -> None:
...
+    def add_dependency(self, dst: Reference, loc: Location, src: Reference) -> None:
+        """Update instr_inputs/instr_uses that `dst` depends on `loc` from `src`"""
+        self.instr_inputs[dst].add(loc, src)
+        self.instr_uses[src].add(loc, dst)
+
+    def remove_dependencies(self, ref: Reference) -> None:
+        """Remove all inputs for `ref` from instr_inputs/instr_uses"""
+        for loc, uses in self.instr_inputs[ref].items():
+            for use in uses:
+                self.instr_uses[use].remove_ref(ref)
```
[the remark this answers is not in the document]
zbanks: I'll refactor this to `self.instr_uses[use].get(loc).remove(ref)` and delete the `remove_ref` fn entirely

src/ir_pattern.py
zbanks: The dependency maps are keyed by location, but the `TryIrMatch.ref_map` match state is not. 

I think of `ref_map` as a many-to-one relationship between pattern instructions & matched instructions, which is also why the direction of the map matters.

src/translate.py
zbanks: Adding a comment - but I'm deliberately using the `in_pattern` heuristic for MIPS as well (unless I shouldn't?)

simonlindholm: It's good for MIPS too, it's just not as effective, because for MIPS we accept arguments set in other basic blocks and for those the meta bit won't have propagated.

src/ir_pattern.py
simonlindholm: I don't quite see how that matters here -- surely the in.fictive instructions are not in ref_map, since inputs to the pattern are allowed to be phis? (I should probably take another look at the code for improved understanding of how the in.fictive's are used)

zbanks: (Putting aside [this issue](https://github.com/matt-kempster/mips_to_c/pull/214#discussion_r855645733) where RefSets are used ambiguously both for phis & overlapping assignments)

The in.ficitve instructions are in ref_map. For an example like the following:

```asm
# replacement: example.fictive $z, $a, $b
fadds $x, $a, $b
fadds $y, $a, $b
fadds $z, $x, $y
```

During matching, we need to check that both uses of `$a` come from the same Reference, and same for `$b`. But, `$a` can come from a different Reference than `$b`. 
The `in.fictive $a` (which has no inputs and outputs only `$a`) acts as placeholder in `ref_map` so that we can check that all uses of `$a` come from the same place. 

You're right, however, that the matching needs to be stricter when phis are involved. There probably needs to be a bigger change made to handle inputs differently than internal registers (so that only inputs can match phis).

zbanks:
Still need to tackle the changes to `flow_graph.py`/`ir_pattern.py`, but I wanted to check that my initial refactor of `StackLocation` is fine.
Namely, I've simplified some of the logic in `translate.py` that makes some of the checks a bit weaker -- but I haven't found any output diffs in the e2e tests or MM, OOT, or PM.

src/ir_pattern.py
zbanks: Did a bigger refactor that eliminates `locations_alias` by assuming that Locations alias iff they are exactly equal. 

This [introduced some small diffs](https://gist.github.com/zbanks/f9c80e4272b3ed61922671ca51cbb150) but they're all  improvements. 

The new strategy makes `ref_map` many-to-many in a way that supports mapping a single reference to a RefSet representing a phi.

src/flow_graph.py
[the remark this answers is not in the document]
zbanks: I can eliminate the `ir.inputs.remove(inp)` -- in practice it seems to hurt slightly more than it helps. [Here's a diff for removing it on pikmin/melee](https://gist.github.com/zbanks/0fd2143f164926a26c4be0aae601a555).
(Anyways, translate.py does something similar to this when it populates `likely_regs`)

I want to keep the logic for this around though, and I'll try to explain it in a comment: I think this analysis is useful to check that instruction parsing is implemented correctly.

src/flow_graph.py
```diff
@@ -1095,16 +1277,216 @@ def reset_block_info(self) -> None:
         for node in self.nodes:
             node.block.block_info = None
 
+    def add_dependency(self, *, dst: Reference, loc: Location, src: Reference) -> None:
+        """Update instr_inputs/instr_uses that `dst` depends on `loc` from `src`"""
```
simonlindholm, line 1281: I just realized I've been misreading the direction of this in all the callers, reading "dst: a, src: b" as "there is a dependency arrow from b to a" (shows how easily my eyes glance over when reading graph manipulation code that I noticed this only now, hah). Can we find better names than src/dst?

simonlindholm:
> Still need to tackle the changes to flow_graph.py/ir_pattern.py, but I wanted to check that my initial refactor of StackLocation is fine.

Looks reasonable to me.

simonlindholm:
As another option, we could leave StackLocation as is (with a size), but add a member function for "split this into bytes/words" (which would use a separate type) and use the split-up version when computing dependencies. I could see having the original size available when doing analysis on stack variables in translate in the future. (For example it would enable rolling back the translate changes.) But we can leave that for later once it actually turns out to be needed.

src/ir_pattern.py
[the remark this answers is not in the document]
zbanks: For posterity: I experimented with reducing duplication on this branch: https://github.com/zbanks/mips_to_c/compare/zbanks:ir-patterns...zbanks:unify-ir-asm-match

though I wasn't very happy with this refactor because it involved a lot of weird OOP tricks.

zbanks (commented):
Did a bit of a self-review with fresher eyes

src/flow_graph.py
```diff
@@ -1095,16 +1260,218 @@ def reset_block_info(self) -> None:
...
+        nodes = duplicate_premature_returns(nodes)
+
     compute_relations(nodes)
-    terminate_infinite_loops(nodes)
-    return FlowGraph(nodes)
+    if not fragment:
+        terminate_infinite_loops(nodes)
+
+    flow_graph = nodes_to_flowgraph(
+        nodes, function, arch, print_warnings=print_warnings
```
simonlindholm, line 1463:
```suggestion
        nodes, function, arch, print_warnings=print_warnings or fragment
```

src/flow_graph.py
```diff
@@ -1095,16 +1260,218 @@ def reset_block_info(self) -> None:
...
+    )
+    if not fragment:
+        arch.simplify_ir(flow_graph)
+
+    return flow_graph
```
simonlindholm, line 1468: (pondering whether it would be wise to create a separate build_fragment_flowgraph function to get rid of the "if not fragment"s, not sure)

src/ir_pattern.py
```diff
@@ -0,0 +1,423 @@
...
+                if not state.match_inputrefs(
+                    pat_inputs,
+                    flow_graph.instr_inputs[cand],
+                ):
+                    continue
+                next_try_matches.append(state)
+            try_matches = next_try_matches
+
+        for n, state in enumerate(try_matches):
+            # Perform any additional pattern-specific validation or computation
```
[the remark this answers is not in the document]
zbanks: Ah yeah, the idea was to compute literals. 
I could have it pass a `TryIrMatch` instead, but I think that'd really weaken the intent behind the `TryIrMatch`/`IrMatch` split.
I'll leave the "computation" comment out for now, since no patterns make use of it yet anyways.

src/arch_ppc.py
```diff
@@ -173,6 +178,87 @@ def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
...
+class CheckConstantMixin:
+    def check(self, m: IrMatch) -> bool:
+        # TODO: Also validate that `K($k)` is the expected constant in rodata
+        return m.symbolic_registers["k"] in (Register("r2"), Register("r13"))
+
+
+class SintToDoubleIrPattern(IrPattern, CheckConstantMixin):
+    # The replacement asm for these patterns reference the float constant `K($k)`
+    # as an input, even though the value is ignored. This is needed to mark `$k`
+    # as an input to the pattern for matching.
```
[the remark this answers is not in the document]
zbanks: Yeah - my reasoning was that it would probably be a mistake to omit an input here? It seemed like a reasonable assertion, but I'd be fine with the change you proposed if there were an explicit use case for it.

src/arch_ppc.py
```diff
@@ -545,11 +630,12 @@ def make_memory_access(arg: Argument) -> Access:
...
                 )
                 inputs = [args[0], args[1], args[2]]
-                outputs = [MemoryAccess(args[1], args[2], size)]
+                writes_memory = True
             else:
                 assert len(args) == 2 and isinstance(args[1], AsmAddressMode)
                 inputs = [args[0], args[1].rhs]
-                outputs = [make_memory_access(args[1])]
+                outputs = make_memory_access(args[1])
+                writes_memory = not outputs
```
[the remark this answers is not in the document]
zbanks: did what the other comment said and just yanked out the reads/writes_memory vars entirely.

src/ir_pattern.py
```diff
@@ -0,0 +1,423 @@
...
+    """
+    IrMatch represents the matched state of an IrPattern.
+    This object is considered read-only; none of its methods modify its state.
+    Its `map_*` methods take a pattern part and return the matched instruction part.
+    """
+
+    symbolic_registers: Dict[str, Register] = field(default_factory=dict)
+    symbolic_labels: Dict[str, str] = field(default_factory=dict)
+    symbolic_args: Dict[str, Argument] = field(default_factory=dict)
+    ref_map: Dict[Reference, RefSet] = field(default_factory=dict)
```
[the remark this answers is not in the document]
zbanks: "prologue" can be used as a key if you were to reference a register directly in the pattern (e.g. use `$r2` instead of `$k`)

src/ir_pattern.py
```diff
@@ -0,0 +1,423 @@
...
+        else:
+            var_map[key] = value
+        return True
+
+    def match_arg(self, pat: Argument, cand: Argument) -> bool:
+        if isinstance(pat, AsmLiteral):
+            return pat == cand
+        if isinstance(pat, Register):
+            # Single-letter registers are symbolic
+            if len(pat.register_name) > 1:
```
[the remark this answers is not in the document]
zbanks: I'll add it as a private helper for now

src/ir_pattern.py
```diff
@@ -0,0 +1,423 @@
...
+            return False
+        return all(self.match_arg(*args) for args in zip(pat.args, cand.args))
+
+    def match_refset(self, pat_set: RefSet, cand_set: RefSet) -> bool:
+        pat = pat_set.get_unique()
+        assert pat is not None
+
+        cand = cand_set.get_unique()
+        if isinstance(pat, str) and isinstance(cand, str):
+            return pat == cand
```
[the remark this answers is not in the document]
zbanks: I should have removed this when I expanded the type of `ref_map`

src/ir_pattern.py
```diff
@@ -0,0 +1,423 @@
...
+
+        # For now, patterns can only have 1 output register (which must be set by the
+        # final instruction in the pattern). This simplifies the replacement step because
+        # we can replace the final instruction and know that all the pattern inputs have
+        # been assigned by there, and the output has not yet been used.
+        assert len(pattern.replacement_instr.outputs) == 1
+        assert pattern.replacement_instr.outputs == tail_ref.instruction.outputs
+
+        # Start the matching by finding all possible matches for the last instruction
+        try_matches = []
```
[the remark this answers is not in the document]
zbanks: I made the change to the loop structure, that worked really well. But, it'd be a bit more work to split the matching/replacing out into a separate method, there are a lot of local variables (the partition of `pattern_refs`, `fictive_reg_index`, `arch`, `flow_graph`, etc)

src/ir_pattern.py
```diff
@@ -158,23 +166,21 @@ def map_arg(self, key: Argument) -> Argument:
...
-            # Uppercase symbols are symbolic; everything else is literal
-            if key.symbol_name.isupper():
+            if self._is_symbolic_sym(key):
                 return self.symbolic_args[key.symbol_name]
             return key
         if isinstance(key, AsmAddressMode):
             rhs = self.map_arg(key.rhs)
             assert isinstance(rhs, Register)
-            return AsmAddressMode(lhs=self.map_arg(key.lhs), rhs=rhs)
+            return AsmAddressMode(lhs=self.map_arg(key.lhs), rhs=self.map_reg(rhs))
```
simonlindholm, line 175: now it's mapped twice

src/ir_pattern.py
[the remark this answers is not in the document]
simonlindholm: Ah! I think we want prologue to always map to prologue, then?

src/flow_graph.py
```diff
@@ -656,10 +712,13 @@ def find_block_by_label(label: str) -> Optional[Block]:
...
-        new_node.successor = build_graph_from_block(
-            next_block, blocks, nodes, asm_data, arch
-        )
+        if block.index + 1 < len(blocks):
+            next_block = blocks[block.index + 1]
+            new_node.successor = build_graph_from_block(
+                next_block, blocks, nodes, asm_data, arch
+            )
+        else:
+            new_node.successor = terminal_node
```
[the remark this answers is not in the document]
simonlindholm: > (I don't know any cases where that safeguard doesn't work)

```
glabel test
nop
.foo:
nop
b .foo
nop
jal bar
nop
```
should fail to trigger it, same with `beqz $a0, .foo` instead of `b .foo`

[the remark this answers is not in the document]
zbanks: I added a check to produce this with `beqz $a0, .foo` (the `b .foo` version produces the same C as master)


```
$ cat bad.s 
glabel test
nop
.foo:
nop
beqz $a0, .foo
nop
jal bar
nop
$ ./mips_to_c.py bad.s
/*
Decompilation failure in function test:

Missing return instruction in block: # 1 (foo)
    nop
    beqz $a0, .foo
    nop

# 1 -> cond: 2, def: 2
*/
```

src/ir_pattern.py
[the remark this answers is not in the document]
zbanks: Not always - in the pattern, you could use a register like `$r1` or `$cr0` which is not set as part of the pattern, but *is* set by some instruction in the candidate function. 
This happens with `$r1` in the existing patterns: for the pattern, it's source will be the prologue; but in the candidate function it's source will be an `addi r1, r1, N` instruction. 

I had to modify `PrologueRef` & `EpilogueRef` to create unique instances for each Location. (I have been going back-and-forth on whether or not this is necessary, but it is needed in this case.)

I'm also using "PrologueRef" to mean "anything *before* the function," but I'm worried this might get confused with the concept of a *function prologue* where registers are saved -- is this too confusing? I could pick a better pair of terms

src/flow_graph.py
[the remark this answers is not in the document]
simonlindholm: Nice. What's unfortunate though is that this is a hard error while the safeguard repairs the function and tries anyway, and also that we now have two separate ways of handling this kind of error... Here's a suggestion:
- add the safeguard `jr $ra` unconditionally
- check if it has in edges; if so, emit a warning
- now that the last block is guaranteed to always be a return node, remove the block.index bounds checking again, as well as the DecompFailure you added

This doesn't work for fragments, but those can instead add a safeguard empty block and when encountering that in `build_graph_from_block` return TerminalNode.

(this is getting complex for such a minor point, sorry...)

src/flow_graph.py
```diff
@@ -1379,10 +1381,10 @@ def nodes_to_flowgraph(
     missing_regs = []
 
     def process_node(node: Node, loc_srcs: LocationRefSetDict) -> None:
-        def add_source_dependencies(dst: Reference, input_loc: Location) -> RefSet:
+        def add_source_uses(use: Reference, input_loc: Location) -> RefSet:
```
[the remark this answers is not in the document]
simonlindholm: fwiw I find add_uses_of_loc just as ambiguous, it sounds like "take a (source, loc) and add uses to it"

src/ir_pattern.py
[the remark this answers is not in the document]
zbanks: They're not fully interchangeable, but they are close. There is a PrologueRef for every register in `arch.all_regs`, but there isn't a PrologueRef for symbolic registers like `$a`. 

(I admit this *does* tie back to your other point about uncoupling the pattern `parts` from `replacement_instr`'s inputs, and if the "regs were read before being written to" is actually necessary -- but as-is they enable that current behavior)

src/flow_graph.py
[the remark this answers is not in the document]
zbanks: > This doesn't work for fragments, but those can instead add a safeguard empty block and when encountering that in `build_graph_from_block` return TerminalNode.

I riffed on this and reverted most of the changes, but introduced this smaller change to `build_nodes` that fixes all the logic.
```python
    if fragment:
        blocks.append(terminal_node.block)
```

src/ir_pattern.py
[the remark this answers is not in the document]
simonlindholm: Mmm. It does bother me that that very minor distinction is enough to create two separate ways of dealing with the same thing. There's one additional difference where with in.fictive's we make temporary registers with copies of the inputs, while with static registers we don't. (And it would be a bit tricky to do so: rename_reg doesn't work for non-symbolic regs, and heuristics in translate.py that depend on specific static regs ($sp in particular) would break. We could add an additional guard check that static regs used in the output instruction still have the same sources, I guess... in practice this won't come up though, ofc. Unless we find some pattern that only works with specific regs but the regs do change across the function.

One ambitious thing we could (later) do in this direction is add a pre-pass that removes stack pointer changes from the asm. Then all static regs would become *really* static and we could go back to a PrologueRef without location argument.

src/flow_graph.py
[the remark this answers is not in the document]
simonlindholm: That's a fair solution. You can back out the added DecompFailure too since it's now impossible to trigger. (I might experiment with my suggested improved safeguard later, would be nice to finally do that properly.)

simonlindholm (approved):

src/ir_pattern.py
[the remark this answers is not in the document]
zbanks: Yeah, when I was experimenting I tried creating a temp for the statics as well: making a temp for `$sp` ended up confusing translate.py and affected the output. (Though, I didn't investigate... it might have been an easy fix.)

PR #215, opened 2022-03-19
Parse args like `%hi(fp)` where `fp` is a symbol

zbanks:
Previously, since `fp` is the name of a MIPS register, this would be incorrectly parsed as `Macro("hi", Register("fp"))`.

This change disables the sigil-less register coercion when parsing `Macro` or `BinOp` expressions.

### DIFF
--- a/src/parse_instruction.py
+++ b/src/parse_instruction.py
@@ -278,8 +278,9 @@ def constant_fold(arg: Argument) -> Argument:
     return arg
 
 
-def detect_bare_reg(arg: Argument, arch: ArchAsmParsing) -> Argument:
-    """Check if `arg` is an AsmGlobalSymbol whose name matches a known Register"""
+def replace_bare_reg(arg: Argument, arch: ArchAsmParsing) -> Argument:
+    """If `arg` is an AsmGlobalSymbol whose name matches a known or aliased register,
+    convert it into a Register and return it. Otherwise, return the original `arg`."""
     if isinstance(arg, AsmGlobalSymbol):
         maybe_reg = Register(arg.symbol_name)
         if maybe_reg in arch.aliased_regs:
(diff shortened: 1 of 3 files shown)


src/parse_instruction.py
```diff
@@ -280,6 +278,17 @@ def constant_fold(arg: Argument) -> Argument:
     return arg
 
 
+def detect_bare_reg(arg: Argument, arch: ArchAsmParsing) -> Argument:
+    """Check if `arg` is an AsmGlobalSymbol whose name matches a known Register"""
```
simonlindholm, line 282: the name/doc comment makes this sound like it returns a bool, and also doesn't communicate that it handles aliased registers

src/parse_instruction.py
```diff
@@ -313,8 +322,6 @@ def expect(n: str) -> str:
             if "$" in reg:
                 # If there is a second $ in the word, it's a symbol
                 value = AsmGlobalSymbol(word)
-            elif reg in arch.aliased_regs:
-                value = arch.aliased_regs[reg]
             else:
                 value = Register(reg)
```
simonlindholm, line 326: maybe detect_bare_reg here, to make the logic less error prone? I agree it shouldn't be needed, though

simonlindholm (approved):

src/parse_instruction.py
zbanks: I'm not 100% sure what you mean here:

- This part of the code needs to be able to handle symbolic register names from asm patterns like `$y`, so `value = Register(reg)` cannot be replaced with `value = replace_bare_reg(AsmGlobalSymbol(reg), arch)`, because `replace_bare_reg` won't turn `y` into `Register("y")`. 

- I don't think `value = AsmGlobalSymbol(word)` here needs a `replace_bare_reg`: the replacement is done when `value` is *used* not when it is constructed. (see the `# Global symbol` section)

- I could handle `Register` input values in `replace_bare_reg`, and then change this code to do `value = replace_bare_reg(Register(reg), arch)` -- but I don't think that is more or less clean than what I've done on my latest commit?

simonlindholm: The last point is what I was thinking, under the misapprehension that `replace_bare_reg` already accepted Register inputs. The way you have it with the latest commit is good, though!

PR #216, opened 2022-03-26
Enable use of MIPS ABI register float aliases in input

EllipticEllipsis:
Converts register names used in the input to the standard internal representation, building on the `aliased_regs` functionality. Uses a dataclass with a dict to keep track of the register names the input has used, and can output the same names in the temps/phis.


src/parse_file.py
```diff
@@ -458,7 +460,11 @@ def process_label(label: str, *, glabel: bool) -> None:
                     lineno=lineno,
                     synthetic=False,
                 )
-                instr: Instruction = parse_instruction(line, meta, arch)
+                instr: Instruction
+                if mips_file.current_function is not None:
+                    instr = parse_instruction(line, meta, arch, mips_file.current_function.used_reg_names)
+                else:
+                    instr = parse_instruction(line, meta, arch, UsedRegNames())
```
simonlindholm, line 467:
```suggestion
                if mips_file.current_function is not None:
                    used_reg_names = mips_file.current_function.used_reg_names
                else:
                    used_reg_names = UsedRegNames()
                instr = parse_instruction(line, meta, arch, used_reg_names)
```

src/parse_instruction.py
```diff
@@ -237,6 +237,24 @@ class NaiveParsingArch(ArchAsmParsing):
...
+
+    def input_to_internal(self, reg: Register, arch: ArchAsmParsing) -> Register:
+        if reg in arch.aliased_regs:    
+            if reg not in self.used_names.values():
+                self.used_names[arch.aliased_regs[reg]] = reg
+            return arch.aliased_regs[reg]
+        else:
+            return reg
+
+    def internal_to_output(self, reg: Register) -> Register:
```
simonlindholm, line 252: I think we should keep Registers canonical, i.e. as far as possible avoid having e.g. Register("fv0"). So can you make this return `str` and input_to_internal` accept `str` as input?

src/parse_instruction.py
```diff
@@ -237,6 +237,24 @@ class NaiveParsingArch(ArchAsmParsing):
...
         return instr
 
+@dataclass
+class UsedRegNames:
+    used_names: Dict[Register, Register] = field(default_factory=dict)
+
+    def input_to_internal(self, reg: Register, arch: ArchAsmParsing) -> Register:
+        if reg in arch.aliased_regs:    
+            if reg not in self.used_names.values():
+                self.used_names[arch.aliased_regs[reg]] = reg
```
simonlindholm, line 247:
```suggestion
            self.used_names[arch.aliased_regs[reg]] = reg
```
(equivalent but O(1) instead of O(n))

src/parse_instruction.py
```diff
@@ -278,15 +296,13 @@ def constant_fold(arg: Argument) -> Argument:
...
+def replace_bare_reg(arg: Argument, arch: ArchAsmParsing, used_reg_names: UsedRegNames) -> Argument:
     """If `arg` is an AsmGlobalSymbol whose name matches a known or aliased register,
     convert it into a Register and return it. Otherwise, return the original `arg`."""
     if isinstance(arg, AsmGlobalSymbol):
         maybe_reg = Register(arg.symbol_name)
-        if maybe_reg in arch.aliased_regs:
-            return arch.aliased_regs[maybe_reg]
         if maybe_reg in arch.all_regs:
-            return maybe_reg
+            return used_reg_names.input_to_internal(maybe_reg, arch)
```
simonlindholm, line 305: aliased_regs aren't included in all_regs, so this needs to check `in all_regs or in aliased_regs`

src/translate.py
```diff
@@ -3749,7 +3750,7 @@ def assign_phis(used_phis: List[PhiExpr], stack_info: StackInfo) -> None:
         if not phi.replacement_expr and phi.propagates_to() == phi:
             counter = name_counter.get(phi.reg, 0) + 1
             name_counter[phi.reg] = counter
-            prefix = f"phi_{phi.reg.register_name}"
+            prefix = f"phi_{stack_info.function.used_reg_names.internal_to_output(phi.reg).register_name}"
```
simonlindholm, line 3753: extract this long expression into a temp

src/parse_instruction.py
```diff
@@ -237,6 +237,24 @@ class NaiveParsingArch(ArchAsmParsing):
...
+class UsedRegNames:
+    used_names: Dict[Register, Register] = field(default_factory=dict)
+
+    def input_to_internal(self, reg: Register, arch: ArchAsmParsing) -> Register:
+        if reg in arch.aliased_regs:    
+            if reg not in self.used_names.values():
+                self.used_names[arch.aliased_regs[reg]] = reg
+            return arch.aliased_regs[reg]
+        else:
+            return reg
```
zbanks, line 250: Maybe something like this instead? It's stricter: it disallows using different aliases to refer to the same internal register?

```suggestion
    def input_to_internal(self, reg: Register, arch: ArchAsmParsing) -> Register:
        internal_reg = arch.aliased_regs.get(reg, reg)
        existing_reg = self.used_names.get(internal_reg)
        if existing_reg is None:
            self.used_names[internal_reg] = reg
        elif existing_reg != reg:
            raise DecompFailure(
                f"Source uses multiple names for {internal_reg} ({existing_reg}, {reg})"
            )
        return internal_reg
```

src/parse_instruction.py
```diff
@@ -237,6 +237,24 @@ class NaiveParsingArch(ArchAsmParsing):
...
+
+    def internal_to_output(self, reg: Register) -> Register:
+        if reg in self.used_names:
+            return self.used_names[reg]
+        else:
+            return reg
```
zbanks, line 256:
```suggestion
    def internal_to_output(self, reg: Register) -> Register:
        return self.used_names.get(reg, reg)
```

src/parse_file.py
zbanks, line 467: very small nit: I think I'd prefer something like this, to minimize the code in the branch
```suggestion
                if mips_file.current_function is not None:
                    used_reg_names = mips_file.current_function.used_reg_names
                else:
                    used_reg_names = UsedRegNames()
                instr = parse_instruction(line, meta, arch, used_reg_names)
```

zbanks: Simon said the same thing, ha

src/parse_instruction.py
EllipticEllipsis: I'm surprised it works at all if that's the case.

src/parse_instruction.py
```diff
@@ -278,15 +296,13 @@ def constant_fold(arg: Argument) -> Argument:
...
     """If `arg` is an AsmGlobalSymbol whose name matches a known or aliased register,
     convert it into a Register and return it. Otherwise, return the original `arg`."""
     if isinstance(arg, AsmGlobalSymbol):
         maybe_reg = Register(arg.symbol_name)
-        if maybe_reg in arch.aliased_regs:
-            return arch.aliased_regs[maybe_reg]
-        if maybe_reg in arch.all_regs:
-            return maybe_reg
+        if maybe_reg in arch.all_regs or maybe_reg.register_name in arch.aliased_regs:
+            return used_reg_names.input_to_internal(maybe_reg.register_name, arch)
```
simonlindholm:
```suggestion
        reg_name = arg.symbol_name
        if Register(reg_name) in arch.all_regs or reg_name in arch.aliased_regs:
            return used_reg_names.input_to_internal(reg_name, arch)
```
to get rid of maybe_reg

src/parse_instruction.py
```diff
@@ -232,11 +232,32 @@ class NaiveParsingArch(ArchAsmParsing):
...
-    aliased_regs: Dict[Register, Register] = {}
+    aliased_regs: Dict[str, Register] = {}
 
     def normalize_instruction(self, instr: AsmInstruction) -> AsmInstruction:
         return instr
 
+@dataclass
+class RegFormatter:
+    """Converts register names used in input assembly to the internal register representation, 
+    saving the original names for use in the output"""
```
simonlindholm, line 243: A bit focussed on the input parsing part, which is a bit weird given the name of the class. Shrug

simonlindholm (commented):
Looks pretty good, just needs a test. (And maybe zbanks has opinion on bikeshed colors)

zbanks (approved):

PR #217, opened 2022-03-28
Fix reg_vars behaviour for individual register names

EllipticEllipsis:
These were not being converted properly, so changed the function call to translate the name correctly. Behaviour is currently to silently use the names from the assembly rather than whatever the user specified. Possibly a note should be generated at output to explain this if it arises, but it would require modifying the parsing functions a fair amount, I imagine.

### DIFF
--- a/src/parse_instruction.py
+++ b/src/parse_instruction.py
@@ -247,6 +247,10 @@ class RegFormatter:
 
     def parse(self, reg_name: str, arch: ArchAsmParsing) -> Register:
         internal_reg = arch.aliased_regs.get(reg_name, Register(reg_name))
+        return internal_reg
+
+    def parse_and_store(self, reg_name: str, arch: ArchAsmParsing) -> Register:
+        internal_reg = arch.aliased_regs.get(reg_name, Register(reg_name))
         existing_reg_name = self.used_names.get(internal_reg)
         if existing_reg_name is None:
             self.used_names[internal_reg] = reg_name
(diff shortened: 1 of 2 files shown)


src/parse_instruction.py
```diff
@@ -246,6 +246,10 @@ class RegFormatter:
     used_names: Dict[Register, str] = field(default_factory=dict)
 
     def parse(self, reg_name: str, arch: ArchAsmParsing) -> Register:
+        internal_reg = arch.aliased_regs.get(reg_name, Register(reg_name))
+        return internal_reg
```
simonlindholm, line 250:
```suggestion
        return arch.aliased_regs.get(reg_name, Register(reg_name))
```

src/translate.py
```diff
@@ -4891,12 +4891,16 @@ def make_arg(offset: int, type: Type) -> PassedInArg:
     elif options.reg_vars == ["all"]:
         reg_vars = arch.saved_regs + arch.simple_temp_regs + arch.argument_regs
     else:
-        reg_vars = list(map(Register, options.reg_vars))
+        reg_vars = list(
+            map(
+                lambda x: stack_info.function.reg_formatter.parse(x, arch),
+                options.reg_vars,
+            )
+        )
```
simonlindholm, line 4899:
```suggestion
        reg_vars = [
            stack_info.function.reg_formatter.parse(x, arch) for x in options.reg_vars
        ]
```

simonlindholm (approved):
Good catch!

PR #219, opened 2022-04-15
Support enum Literals & case labels

zbanks:
I had worked on this branch a few months ago, but at the time I didn't think it was particularly useful because most (N64) projects didn't make much use of enum types. 

However, it looks like the GCN projects have header files which *do* use enum types, and it seems to help a lot with graphics code. Here's an example diff:
```diff
-    GXSetTevAlphaOp(1, 0U, 0, 0, 1U, 0);
-    GXSetBlendMode(1, 4, 5, 0xF);
-    GXSetVtxAttrFmt(0, 9, 1, 3, 0U);
-    GXSetVtxAttrFmt(0, 0xB, 1, 5, 0U);
-    GXSetVtxAttrFmt(0, 0xD, 1, 2, 0xFU);
+    GXSetTevAlphaOp(GX_TEVSTAGE1, 0U, GX_TB_ZERO, GX_CS_SCALE_1, 1U, GX_TEVREG0);
+    GXSetBlendMode(GX_BM_BLEND, GX_BL_SRCALPHA, GX_BL_INVSRCALPHA, GX_LO_SET);
+    GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_POS, GX_POS_XYZ, GX_S16, 0U);
+    GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_CLR0, GX_POS_XYZ, GX_RGB565, 0U);
+    GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_TEX0, GX_POS_XYZ, GX_U16, 0xFU);
```


src/translate.py
```diff
@@ -1505,6 +1505,10 @@ def dependencies(self) -> List[Expression]:
         return []
 
     def format(self, fmt: Formatter, force_dec: bool = False) -> str:
+        enum_value = self.type.get_enum_value(self.value)
+        if enum_value is not None:
+            return enum_value
```
simonlindholm, line 1510: a test hitting this would be nice

src/translate.py
```diff
@@ -4693,6 +4697,9 @@ def for_type(type: Type) -> Optional[str]:
 
                 value = read_uint(size)
                 if value is not None:
+                    enum_value = type.get_enum_value(value)
+                    if enum_value is not None:
+                        return enum_value
```
simonlindholm, line 4702: this too

src/c_types.py
```diff
@@ -78,10 +78,16 @@ class Function:
     is_variadic: bool
 
 
+@dataclass(eq=False)
+class Enum:
+    tag: Optional[str]
+    values: Dict[int, str]
```
simonlindholm, line 84: "values" sounds like a mapping of str to int to me, would prefer "names"/"labels"/similar (also in naming like "get_enum_value")

src/c_types.py
```diff
@@ -368,6 +369,7 @@ def parse_enum(enum: ca.Enum, typemap: TypeMap) -> None:
             value = next_value
         next_value = value + 1
         typemap.enum_values[enumerator.name] = value
+        typemap_enum.values[value] = enumerator.name
```
simonlindholm, line 372: So if there are multiple names mapping to a single value we take the last one. Makes sense; maybe worth observing in a comment

simonlindholm (approved):
This is great

tests/end_to_end/enums/mwcc-o4p-out.c
```diff
@@ -0,0 +1,15 @@
...
+            }
+            return array->unk0;
+        }
+        if (x != 4) {
+            return ZERO;
+        }
+        return array->unk8;
+    }
+    return array->unk4;
+}
```
zbanks: This test fails to detect the switch statement, and the types are not fully unified across comparisons -- but both of those are separate from this PR.

tests/end_to_end/enums/irix-o2-out.c
```diff
@@ -0,0 +1,14 @@
...
+    case ZERO:
+        return *array;
+    case TWO:
+        return *array;
+    case THREE:
+        return *array;
+    default:
+        return ZERO;
+    }
+}
```
zbanks: This test shows an outstanding issue with `%hi`/%lo`, where the `+4` part of `%lo(array+4)` is discarded, so the output here is incorrect. 
(The asm was hand-crafted, but the original asm had `0xE4` instead of `%lo(array+4)`; afaict my edits aren't the cause of this issue)


This test does showcase all 3 places enum names can be used:
- As part of a data declaration (in `array[4] = ...` on line 1)
- As part of a switch label
- As a `Literal` (in `return ZERO;` on line 12)

PR #220, opened 2022-04-16
Improve asm file parsing; prune oversized structs

zbanks:
This PR is a combination of a few small fixes:

- Prune inferred fields in structs that exceed the (known) size of the struct itself.
    - [Diff for N64 projects](https://gist.github.com/zbanks/56826ed69047b6cd25fdbb9ca36b5c36) -- prevents issues like including a whole `Actor` or `GlobalContext` inside a smaller struct.
- Handle quoted symbols individually when parsing asm arguments, to parse terms like `"foo<bar>@baz"@ha`
- Still use `csv` to parse asm directives in `parse_file.py`, but consolidate the `.strip()` operations.
- Allow `0` to symbolize the zero register in AsmAddressMode, like in `lwz r4, 0xc0(0)`
- Handle `.if 0` and `.if 1` directives instead of skipping them

### DIFF
--- a/src/parse_instruction.py
+++ b/src/parse_instruction.py
@@ -406,6 +406,8 @@ def parse_arg_elems(
             else:
                 # Address mode.
                 rhs = replace_bare_reg(rhs, arch, reg_formatter)
+                if rhs == AsmLiteral(0):
+                    rhs = Register("zero")
                 assert isinstance(rhs, Register)
                 value = AsmAddressMode(value or AsmLiteral(0), rhs)
         elif tok == '"':
--- a/src/parse_instruction.py
+++ b/src/parse_instruction.py
@@ -267,20 +267,23 @@ valid_number = "-xX" + string.hexdigits
 
 
 def parse_word(elems: List[str], valid: str = valid_word) -> str:
-    S: str = ""
+    ret: str = ""
     while elems and elems[0] in valid:
-        S += elems.pop(0)
-    return S
+        ret += elems.pop(0)
+    return ret
 
 
-def parse_quoted(elems: List[str], quote_chars: str) -> str:
-    S: str = ""
-    while elems and elems[0] not in quote_chars:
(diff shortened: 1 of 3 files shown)


src/parse_instruction.py
```diff
@@ -274,6 +273,16 @@ def parse_word(elems: List[str], valid: str = valid_word) -> str:
     return S
 
 
+def parse_quoted(elems: List[str], quote_chars: str) -> str:
+    S: str = ""
```
simonlindholm, line 277: S -> ret

src/parse_instruction.py
```diff
@@ -274,6 +273,16 @@ def parse_word(elems: List[str], valid: str = valid_word) -> str:
...
+    while elems and elems[0] not in quote_chars:
+        # Handle backslash-escaped characters
```
simonlindholm, line 279:
```
# We only need to care about \\, \" and \' in this context.
```

src/parse_instruction.py
```diff
@@ -274,6 +273,16 @@ def parse_word(elems: List[str], valid: str = valid_word) -> str:
...
+        if elems[0] == "\\":
+            elems.pop(0)
+        S += elems.pop(0)
```
simonlindholm, line 282: return early if not elems, to leave (currently somewhat non-existent) unterminated string literal error handling to the caller

src/parse_instruction.py
```diff
@@ -274,6 +273,16 @@ def parse_word(elems: List[str], valid: str = valid_word) -> str:
     return S
 
 
+def parse_quoted(elems: List[str], quote_chars: str) -> str:
```
simonlindholm, line 276: quote_chars -> quote_char, and `in quote_chars` -> `== quote_char`

simonlindholm (approved):

src/parse_instruction.py
zbanks: I think I fixed what you were asking about? But if not, I can push another commit

PR #221, opened 2022-04-24
Expand % pattern to expressions

zbanks:
The previous pattern for detecting mod operations was limited to taking the mod of a `Literal` constant.

However, PPC has no mod instruction, so mod is implemented using the same series of instructions that GCC uses to mod by a constant in MIPS, `x % y` becomes `(x - ((x / y) * y))`.

This change produces no diff in any of the MIPS project corpuses, but makes several improvements to the PPC ones.
[Diff in the SMS project](https://gist.github.com/zbanks/5adcd36cabc76beebd382105446f4c71)

(Leaving this in draft until the big PR gets in; I just don't want to forget about this)

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -3039,20 +3039,20 @@ def fold_divmod(original_expr: BinaryOp) -> BinaryOp:
                 right=Literal(new_denom),
             )
 
-    # Detect `%`: (x - ((x / N) * N)) --> x % N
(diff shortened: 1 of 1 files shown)


simonlindholm (approved):
(test could be nice but I'm happy with this either way, feel free to merge this before the IR pattern PR if you want)

PR #223, opened 2022-04-26
Rename the project to m2c

zbanks:
External tasks:
- [x] Simon to add https://simonsoftware.se/other/m2c.html
- [ ] Matt to rename the repository
- [ ] (post-merge) Update decomp.me [wrapper](https://github.com/decompme/decomp.me/blob/main/backend/coreapp/m2c_wrapper.py)

Open questions:
- Do we want to make `-t`/`--target` a required option?
    - Maybe make `mips_to_c.py` default to `--target mips-ido-c`?
- Are there any other links we need to update?
- Any other breaking CLI changes we want to make?


simonlindholm (approved):
Remove the mips_to_c.py symlink imo, it breaks tab completion.

Mention the historical name somewhere.

I'm torn as to whether to require a --target or not. Maybe it is actually worth doing at the same time as renaming the entry point?

zbanks:
I removed `m2c.py` as well as all of the deprecated CLI arguments (like `--structs`, `--rodata`, and function without `--function`).
Added a note about the `mips_to_c` name in the README.

src/parse_file.py
```diff
@@ -276,9 +276,9 @@ def parse_incbin(
     return None
 
 
-def parse_file(f: typing.TextIO, arch: ArchAsm, options: Options) -> MIPSFile:
+def parse_file(f: typing.TextIO, arch: ArchAsm, options: Options) -> AsmFile:
     filename = Path(f.name).name
-    mips_file: MIPSFile = MIPSFile(filename)
+    mips_file: AsmFile = AsmFile(filename)
```
simonlindholm, line 281: if you feel like changing it there's a bunch more instances of "mips" in this file, e.g. this local var name, and also the doc comment at the top of asm_instruction.py.

zbanks: After this I checked with `rg -i mips src/ *.py` and the remaining uses look genuinely MIPS-specific.

matt-kempster:
Looks great!

I'll change the repository URL after this merges.

PR #224, opened 2022-04-27
Cast Literals in comparisons

zbanks:
This idea was mentioned by Simon earlier today in Discord.

The output looks pretty good: there are some NULLs & enums that show up. The only downside is that more literals now have `U` suffixes? (And this change causes a *lot* of diffs in the project corpuses)

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -999,12 +999,8 @@ class BinaryOp(Condition):
         # For comparisons to a Literal, cast the Literal to the type of the lhs
         # (This is not done with complex expressions to avoid propagating incorrect
         # type information: end-of-array pointers are particularly bad.)
-        if self.is_comparison() and isinstance(right_expr, Literal):
-            right_expr = as_type(right_expr, left_expr.type, True)
-
-            # Casts & suffixes for (in)equality comparisons are too verbose
-            if self.op in ("==", "!="):
-                right_expr = elide_literal_casts(right_expr)
+        if self.op in ("==", "!=") and isinstance(right_expr, Literal):
+            right_expr = elide_literal_casts(as_type(right_expr, left_expr.type, True))
 
         if (
             not self.is_floating()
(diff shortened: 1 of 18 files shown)


tests/end_to_end/multi-switch/irix-o2-noswitch-out.c
```diff
@@ -11,9 +11,9 @@ s32 test(s32 arg0) {
     phi_a0_3 = arg0;
     phi_a0_4 = arg0;
     phi_a0_5 = arg0;
-    if (arg0 >= 0x33) {
-        if (arg0 >= 0x6C) {
-            if (arg0 != 0xC8) {
+    if (arg0 >= 0x00000033) {
+        if (arg0 >= 0x0000006C) {
+            if (arg0 != 0x000000C8) {
```
zbanks, line 16: These tests have `--zfill-constants` enabled, which is why they have more zeros. (otherwise, this isn't normal)

zbanks:
With the latest changes, now most of the project diffs are just `0` --> `NULL`

src/translate.py
```diff
@@ -996,6 +996,16 @@ def format(self, fmt: Formatter) -> str:
                 type=self.type,
             ).format(fmt)
 
+        # For comparisons to a Literal, cast the Literal to the type of the lhs
+        # (This is not done with complex expressions to avoid propagating incorrect
+        # type information: end-of-array pointers are particularly bad.)
+        if self.is_comparison() and isinstance(right_expr, Literal):
```
simonlindholm, line 1002: We probably don't want enum names or NULL to show up for </>/<=/>= comparisons. Can we move the `self.op in ("==", "!=")` check to up here?

simonlindholm (approved):

src/translate.py
zbanks: ah, I thought you were concerned with </>/<=/>= & U; this sounds good then.

PR #225, opened 2022-04-30
Move instruction logic out of translate.py and into arch_*.py

zbanks:
Add `Instruction.eval_fn` and set it in `Arch.parse(...)` to move instruction evaluation logic out of translate.py. 

I also refactored some of the `instrs_*` members: now they're only referenced in a single function, so they no longer need to be as generic. 

The only output changes from this PR are in the tracebacks for certain errors.

### DIFF
--- a/src/arch_mips.py
+++ b/src/arch_mips.py
@@ -44,16 +44,14 @@ from .translate import (
     Cast,
     CmpInstrMap,
     CommentStmt,
+    ErrorExpr,
     ExprCondition,
     ExprStmt,
-    ErrorExpr,
     Expression,
     InstrArgs,
     InstrMap,
-    InstrSet,
     Literal,
     NodeState,
-    PairInstrMap,
     SecondF64Half,
     StmtInstrMap,
     StoreInstrMap,
@@ -686,6 +684,8 @@ class MipsArch(Arch):
         is_return = False
         eval_fn: Optional[Callable[[NodeState, InstrArgs], object]] = None
 
+        instr_str = str(AsmInstruction(mnemonic, args))
+
         memory_sizes = {
             "b": 1,
             "h": 2,
@@ -706,6 +706,11 @@ class MipsArch(Arch):
                     assert size <= 4
                     return [loc]
             return []
+
+        def unreachable_eval(s: NodeState, a: InstrArgs) -> None:
+            raise DecompFailure(
+                f"Instruction {instr_str} should be replaced before eval"
+            )
 
         if mnemonic == "jr" and args[0] == Register("ra"):
             # Return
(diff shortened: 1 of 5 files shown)


src/translate.py
```diff
@@ -4241,207 +4221,34 @@ def current_instr(self, instr: Instruction) -> Iterator[None]:
...
+def evaluate_instruction(instr: Instruction, state: NodeState) -> None:
+    # Check that instr's attributes are consistent
+    if instr.is_return:
+        assert isinstance(state.node, ReturnNode)
+    if instr.is_conditional:
+        assert state.branch_condition is None and state.switch_control is None
+
+    # Handle function_targets here, instead of inside eval_fn, because we all the
+    # information we need is already encoded in the Instruction
+    if instr.function_target is not None:
```
simonlindholm, line 4233: Function calls need a comment in arch_* if we are to leave this here. Weakly in favor of moving this logic into arch_* though, to have a single place where we handle instructions.

(Also pondering whether we can make `function_target` a boolean property, but then return type narrowing would need to poke at raw instruction arguments, so probably a bad idea)

src/arch_ppc.py
```diff
@@ -734,24 +811,90 @@ def make_memory_access(arg: Argument, size: int) -> List[Location]:
...
+            eval_fn = lambda s, a: s.set_reg(dest_reg, a.reg(0))
         elif mnemonic in cls.instrs_ppc_compare:
             assert len(args) == 3 and isinstance(args[1], Register)
             inputs = [r for r in args[1:] if isinstance(r, Register)]
             outputs = list(cr0_bits)
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                base_reg = a.reg_ref(0)
+                if base_reg != Register("cr0"):
+                    raise DecompFailure(
```
simonlindholm, line 866: pre-existing, but DecompFailure is a bit harsh, esp with the current error handling which a) is per block, and b) leaves stale registers, causing cascading failures

src/arch_mips.py
```diff
@@ -817,14 +849,24 @@ def make_memory_access(arg: Argument) -> List[Location]:
...
-        elif mnemonic in cls.instrs_source_first:
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                store = cls.instrs_store[mnemonic](a)
+                if store is not None:
+                    s.store_memory(
+                        source=store.source, dest=store.dest, reg=a.reg_ref(0)
+                    )
+
+        elif mnemonic in "mtc1":
```
simonlindholm, line 860:
```suggestion
        elif mnemonic == "mtc1":
```

simonlindholm (approved):

src/arch_ppc.py
zbanks: Fixing this, unsurprisingly, changed the test output: functions that previously crashed with `DecompFailure` now partially decompile. (They're very messy because the non-CR0 bits are not supported by the branch instructions either)

PR #226, opened 2022-04-30
Add support for MIPS add, sub, and bal instructions

Tmcg2:
As best as I can tell `bal` is nearly equivalent to `jal`,
 with the difference being that `bal` can only be used if the
 function being called is in the same file.

That is to say, `bal` indicates that the function being
 called is position-independent code.

Also includes some formatting updates from black

Signed-off-by: Taggerung <tyler.taggerung@gmail.com>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
     - id: black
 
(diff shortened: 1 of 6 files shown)


zbanks:
Thanks, looks good!

I'm interested in seeing where these instructions show up in asm (I found [`tkmk00_decode.s`](https://github.com/n64decomp/mk64/blob/master/asm/tkmk00_decode.s) and some others in that repo). 
If they're ever emitted by a compiler, and we understand *when* the compiler chooses them over jal/addu/subu, we may be able to incorporate additional information into the output? (But I'm fine revisiting this later.)

Tmcg2:
They also show up in https://github.com/n64decomp/mk64/blob/master/asm/mio0_decode.s, which is where I first ran into them.

PR #228, opened 2022-04-30
Allow context arg to accept multiple files

Tmcg2:
Given 3 context files, these changes make the following invocations valid:

`m2c.py some_asm.s --context ctx_1.c ctx_2.c ctx_3.c`
`m2c.py some_asm.s --context ctx_1.c --context ctx_2.c --context ctx_3.c`
`m2c.py some_asm.s --context ctx_1.c --context ctx_2.c ctx_3.c`

Note that the use of `extend` for `action` requires Python 3.8 or later

Signed-off-by: Taggerung <tyler.taggerung@gmail.com>

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -251,7 +251,8 @@ def parse_flags(flags: List[str]) -> Options:
         "--context",
         metavar="C_FILE",
         dest="c_contexts",
-        action="append",
+        action="extend",
+        nargs="+",
         type=Path,
         default=[],
         help="Read variable types/function signatures/structs from an existing C file. "


zbanks:
Thanks for the PR -- though I'd like to hear more about the intended use case.

In general, I believe it's more common to have multiple input asm files than it is to have multiple context files.
Generally there are a limited number of context files, because they are usually prepossessed to expand macros. (And the preprocessor can be used to concatenate multiple context files [like this](https://github.com/n64decomp/mk64/blob/master/ctx_includes.c).)
And on the other hand, there are a lot of projects where globbing with `*.s` does the right thing (`./m2c.py path/to/overlay/*.s`).

Separately, I'm hesitant to use the `action="extend"` for two reasons. First, this repository is compatible with Python 3.6 and I don't think this is worth upgrading for. Second, because we have positional arguments, this can introduce parsing ambiguities that I think can be confusing to users. It would break commands like `./m2c.py --context ctx.c asm.s`.

I'd like to hear more about your use case, though. Maybe there's an alternate solution? Also happy to chat on Discord (`zbanks#2143`).

Tmcg2:
Its probably just better for me to change how I do stuff on my end then try to force a change here. I had these changes floating around on my system for reasons that no longer apply to my actual workflow. I think in my head the original intent was make it so that I could use glob patterns for context files, but I don't do that anymore and I don't intend to in the future.

PR #229, opened 2022-05-02
Split out evaluator helpers from translate.py

zbanks:
This PR moves functions from `translate.py` to a new `evaluate.py` file verbatim, in the same order they originally appeared in the original file. 
(Exception: `CarryBit.add_to` & `CarryBit.sub_from` were rewritten as `carry_add_to` and `carry_sub_from` and placed at the end of the file)


simonlindholm:
Sorry to be slow to get to this (and to totally bit rot it with by phi refactoring...) I think I like the first part, it feels like a relatively clean conceptual split. The second I'm a bit more torn about; it's probably good but I need to digest it first... I'm a bit cautious about the split given my vim grep-based workflow, but I think I'll manage. (And I'm not planning similarly invasive changes as my phi work any time soon.)

PR #230, opened 2022-05-02
Support .set

simonlindholm:


src/asm_file.py
```diff
@@ -385,68 +396,68 @@ def process_label(label: str, *, glabel: bool) -> None:
...
                     curr_section = ".bss"
-                elif line.startswith(".text"):
+                elif directive == ".text":
                     curr_section = ".text"
+                elif directive == ".set":
+                    # Either a directive like ".set noreorder", or defining a variable
+                    _, _, args_str = line.partition(" ")
+                    args = split_arg_list(args_str)
+                    if len(args) == 2:
+                        defines[args[0]] = try_parse(lambda: parse_int(args[1]))
```
zbanks, line 435: Should we warn on `len(args) > 2`?

src/options.py
```diff
@@ -107,7 +107,7 @@ class GlobalDeclsEnum(ChoicesEnum):
     use_cache: bool
     dump_typemap: bool
     pdb_translate: bool
-    preproc_defines: Dict[str, int]
+    preproc_defines: Dict[str, Optional[int]]
```
zbanks, line 110: Maybe a comment here or somewhere that `None` represents unset macros, and `1` is used for macros set without a value? 
Naively I might expect that `-D FOO` would set `preproc_defines["FOO"] = None` instead of `1`.

zbanks (approved):

src/asm_file.py
```diff
@@ -358,24 +368,25 @@ def process_label(label: str, *, glabel: bool) -> None:
...
-                if line.startswith(".ifdef"):
+                level = 1 if defines[macro_name] is not None else 0
+                if directive == ".ifdef":
                     level = 1 - level
                 ifdef_level += level
                 ifdef_levels.append(level)
-            elif line.startswith(".if"):
-                macro_name = line.split()[1]
+            elif directive.startswith(".if"):
+                macro_name = directive.split()[1]
```
zbanks, line 389: I think this should stay `macro_name = line.split()[1]`? This breaks on some existing files.

simonlindholm: Oops! Good catch

PR #232, opened 2022-05-07
Add `--deterministic-vars` to make phi/temp diffs easier to read

zbanks:
The `--deterministic-vars` option names `temp` vars after the source line number, and `phi` vars after the node index. These are both more stable than using a per-register incrementing counter. 

This should help make diffs smaller when testing changes to `translate.py`: adding/removing variables will not cause unrelated variables to be renamed, keeping the diff brief. 

I didn't have any better ideas for a shorter argument name (if you have any better ideas lmk). I did not add the option to the website, since I believe this is mostly useful for local testing & development. It can be added to a `./run_test.py` run with `-- --deterministic-vars`.

Here's an example diff for before/after `7a467845` ("Avoid unwrapping same-value phis when possible"), both with & without this option: https://gist.github.com/zbanks/a21dabff7e0bf2c25908aaa879a77764 . In this case, using `--deterministic-vars` makes the output 10% shorter.

Maybe I should have this enabled by default in `run_tests.py` for all project tests?


simonlindholm (approved):
Nice idea with using line numbers for this, it makes it easier to trace things back to the asm as well (could in principle do the same for phis for consistency, using the line number of the first instruction of the block if one exists, but it's not as useful there). I agree with making this the default for project tests.

PR #233, opened 2022-05-11
EvalOnce-wrap everything in RegInfo

simonlindholm:
Use EvalOnce wrappers for:
- initial registers (arguments, sp, gp, etc.)
- SecondF64Half
- floating point comparisons
- literals

and statically type-check that RegInfo.contents contains only EvalOnceExpr, RegisterVar, and PhiExpr. (Of which the latter two will probably eventually be merged.) Add RegMeta.initial to replace PassedInArg.copied (not strictly necessary but the wrappers got in the way of checking for "copied" and it seemed a nice cleanup).

This is in preparation for the phi refactoring I'm working on.

OOT project diff is approximately empty, haven't checked other projects.

### DIFF
--- a/src/translate.py
+++ b/src/translate.py
@@ -300,7 +300,6 @@ class StackInfo:
         real_location = location & -4
         arg = PassedInArg(
             real_location,
-            copied=True,
             stack_info=self,
             type=self.unique_type_for("arg", real_location, Type.any_reg()),
         )
@@ -1230,7 +1229,6 @@ class RegisterVar(Expression):
 @dataclass(frozen=True, eq=True)
 class PassedInArg(Expression):
     value: int
-    copied: bool = field(compare=False)
     stack_info: StackInfo = field(compare=False, repr=False)
     type: Type = field(compare=False)
 
@@ -1942,8 +1940,22 @@ class RawSymbolRef:
 
 @dataclass
 class RegMeta:
+    """Metadata associated with a register value assignment.
+
+    This could largely be computed ahead of time based on instr_inputs/
+    instr_uses, except for the fact that function pointer argument passing
+    is determined dynamically based on type info.
+
+    Except for a few properties, metadata is propagated across blocks only
+    after translation is done (because parent block metadata is not generally
+    available when constructing reg metas)."""
+
     # True if this regdata is unchanged from the start of the block
     inherited: bool = False
+
+    # True if this regdata is unchanged from the start of the function
+    # Propagated early
+    initial: bool = False
 
     # True if this regdata is read by some later node
     is_read: bool = False
@@ -1956,6 +1968,7 @@ class RegMeta:
     uninteresting: bool = False
 
     # True if the regdata must be replaced by variable if it is ever read
+    # Propagated early
     force: bool = False
 
     # True if the regdata was assigned by an Instruction marked as in_pattern;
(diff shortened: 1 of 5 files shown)


src/translate.py
```diff
@@ -1942,9 +1940,23 @@ def __str__(self) -> str:
...
+    after translation is done (because parent block metadata is not generally
+    available when constructing reg metas)."""
+
     # True if this regdata is unchanged from the start of the block
     inherited: bool = False
 
+    # True if this regdata is unchanged from the start of the function
+    # Propagated early
+    initial: bool = False
+
```
zbanks, line 1959: This is a nice change!

src/translate.py
```diff
@@ -1942,9 +1940,23 @@ def __str__(self) -> str:
...
 class RegMeta:
+    """Metadata associated with a register value assignment.
+
+    This could largely be computed ahead of time based on instr_inputs/
+    instr_uses, except for the fact that function pointer argument passing
+    is determined dynamically based on type info.
+
+    Except for a few properties, metadata is propagated across blocks only
+    after translation is done (because parent block metadata is not generally
+    available when constructing reg metas)."""
```
zbanks, line 1951: Maybe mention `propagate_register_meta` here?
This is sort of related to something I brought up in the Discord recently, that "translation" is a bit ambiguous and it'd be nice to have terms for the sub-phases. (The propagation here happens after block translation, but during `translate_to_ast`...)

tests/end_to_end/arg-temp/test.s
```diff
@@ -0,0 +1,5 @@
+glabel test
+sw $zero, ($sp)
+move $v0, $a0
+jr $ra
+nop
```
zbanks, line 5: This is a really interesting test case -- did you add this because you have a plan to separate out `$a0` from `0($sp)` in a future PR?

src/translate.py
```diff
@@ -4856,13 +4885,17 @@ def make_arg(offset: int, type: Type) -> PassedInArg:
     for slot in abi.arg_slots:
         stack_info.add_known_param(slot.offset, slot.name, slot.type)
```
zbanks, line 4886: I think there's some interaction in `add_known_param` that was broken here: the diff for MM with context is very large and seems to use the original `Actor *` type for the first argument instead of the derived struct? 

[First 10k lines of diff](https://gist.github.com/zbanks/a86da755f7e7f076a9b6584f7b10da90), `ArrowFire_Init` is a good example.

The `add_known_param` function is a little weird, and has a way to override `slot.type` for the arg. (I guess the `type` of the `PassedInArg` wasn't used before?)

zbanks (changes requested):

src/translate.py
zbanks: If you need some context files, [here are the ones I use for OOT & MM](https://gist.github.com/zbanks/537c38bee61f8fec1c67b01782cb4738).

tests/end_to_end/arg-temp/test.s
simonlindholm: This test case tests that temps for arguments and prevent_later_uses on arguments works correctly: before this PR, the output wouldn't include temp_a0 and thus be incorrect. With the PR there's a temp for a0 and the store to 0(sp) marks it as forced.

src/translate.py
simonlindholm: Good catch, serves me right for not running mm tests. I made it return the type to use for the parameter. (I wonder if that machinery could change to use the newly introduced arg temps, instead of having special handling in if_statements.py.) New diff, for both mm and oot this time: https://gist.github.com/simonlindholm/95b45d1096a964a67bb89088d4e0581b Looks like a slightly improvement, but completely by accident: I don't know what actually triggers the changes.

src/translate.py
simonlindholm: Done, went with the term "block translation" for now.

PR #235, opened 2022-05-26
Look at stack struct for local variables

simonlindholm:
This adds missing sp vars that were either only read from (due to being fake and actually part of structs that we fail to detect), or are structs/arrays which never ended up in local_vars because we only generated subfield accesses. Despite adding fake vars I think it's an improvement to have them in the output for the sake of coherence?

(Hard to give project diffs because of how separated declarations are from usages.)

### DIFF
--- a/src/if_statements.py
+++ b/src/if_statements.py
@@ -1424,21 +1424,23 @@ def get_function_text(function_info: FunctionInfo, options: Options) -> str:
         # Format the body first, because this can result in additional type inferencce
         formatted_body = body.format(fmt)
 
-        local_vars = function_info.stack_info.local_vars
+        stack_struct = function_info.stack_info.stack_struct
+        assert stack_struct is not None
+        stack_struct_fields = stack_struct.fields
+
         # GCC's stack is ordered low-to-high (e.g. `int sp10; int sp14;`)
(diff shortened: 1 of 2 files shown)


src/if_statements.py
```diff
@@ -1424,21 +1424,23 @@ def get_function_text(function_info: FunctionInfo, options: Options) -> str:
...
-                comment = None
-                if local_var.value in function_info.stack_info.weak_stack_var_locations:
-                    comment = "compiler-managed"
-                function_lines.append(
-                    SimpleStatement(f"{type_decl};", comment=comment).format(fmt)
-                )
-                any_decl = True
+            stack_struct_fields = stack_struct_fields[::-1]
+        for field in stack_struct_fields:
+            type_decl = field.type.to_decl(field.name, fmt)
```
zbanks, line 1436: I think `LocalVar.toplevel_decl(...)` can be deleted now?

zbanks (approved):
It seems like the output changes are pretty small, so I'm good with adding this. (I don't feel super strongly about it though)

On Discord you mentioned: 
> maybe we should instead emit unkspXX for reads from unwritted/unaddressed stack locations
> to show that something weird is probably up

-- I think this would be neat, but I don't know if it's worth adding? (It might not be worth the code)

PR #236, opened 2022-05-28
Move "end" switch labels to inside the switch block

zbanks:
This moves `case ...:` labels that were previously immediately outside of the switch block to `case ...: break;` statements at the top of the switch.

This avoids syntax errors (and is easier to read). It fixes the issue that camthesaxman brought up on Discord about `func_80048084`.

This might not be the absolute cleanest way of implementing this: it's a bit odd to remove labels from `case_nodes[...]`. But it doesn't break `--gotos-only` where `add_labels_for_switch(...)` is used without `build_switch_statement(...)`.

[Diffs in PPC projects](https://gist.github.com/zbanks/6b662e2b99ec77686c8f76493a6e2e99) -- it's relatively small for 6 projects, and it all looks reasonable to me.


src/if_statements.py
```diff
@@ -1058,6 +1059,22 @@ def build_switch_statement(
     """
     switch_body = Body(print_node_comment=context.options.debug)
 
+    # If there are any case labels to jump to the `end` node immediately after the
+    # switch block, emit them as `case ...: break;` at the start of the switch block
+    # instead. This avoids having "dangling" `case ...:` labels outside of the block.
+    remaining_labels = []
+    for index, case_label in context.case_nodes[end]:
+        if index == switch_index:
+            comment = f"switch {switch_index}" if switch_index != 0 else None
```
simonlindholm, line 1068: this is the third time we repeat this logic, better make a helper out of it. (making SimpleStatement.comment into a list?)

simonlindholm (approved):
Seems reasonable. A test would be good.

PR #237, opened 2022-06-06
Split out evaluator helpers from translate.py

zbanks:
This PR just contains the first part of #229 -- I re-did the split to avoid running into any merging issues and preserve the `zbanks:split-translate-evaluate` branch (I didn't want to force-push over it).

---

This  moves functions from `translate.py` into a new `evaluate.py` file verbatim, in the same order they originally appeared.

As an exception: `CarryBit.add_to` & `CarryBit.sub_from` were rewritten as `carry_add_to` and `carry_sub_from` and placed at the end of the file.


simonlindholm (approved):
Could be a good opportunity to add a doc comment at the top of the new file explaining what it's for.

PR #239, opened 2022-06-21
Simplify brief flag checks such as !(var & 4)

ethteck:
This PR makes flag checks briefer and more in line with how projects tend to actually write this code:
`(flagVar & 0x100) == 0` becomes `!(flagVar & 0x100)`
`(flagVar & 0x100) != 0` becomes `(flagVar & 0x100)`


ethteck:
I implemented the paren handling I mentioned above, but the way I did it and the scope might not be ideal. lmk what you think

src/translate.py
```diff
@@ -1092,6 +1092,19 @@ def format(self, fmt: Formatter) -> str:
...
+            fmt.coding_style.brief_flag_checks
+            and isinstance(left_expr, BinaryOp)
+            and left_expr.op == "&"
+            and isinstance(right_expr, Literal)
+            and right_expr.value == 0
+        ):
+            if self.op == "==":
+                return f"!({strip_parens(lhs)})"
+            elif self.op == "!=":
+                return f"({strip_parens(lhs)})"
```
simonlindholm, line 1106: This simplification is only valid within if/while, not for e.g. `x = ((a & 1) != 0)`. I think you want to move this logic to `simplify_condition`. (Where you also don't have to fiddle with formatting stuff like parentheses, nicely enough)

simonlindholm (commented):
I'd be fine with doing this by default

PR #246, opened 2022-12-19
Package for use as a library

nanaian:
- Rename `src` to `m2c`, conforming to [Poetry's expectation](https://python-poetry.org/docs/pyproject/#packages)
- Add a `pyproject.toml` so that other projects (such as decomp.me) can add m2c as a git dependency


matt-kempster:
Looks good to me. Could we add some documentation to README.md as well?

simonlindholm:
thoughts:
- is there a way to make this work with `src/`, or is it needed to get `m2c.main` and similar import paths? (keeping src/ would help my auto-complete habits slightly)
- does mypy type-checking still work for decomp.me with this change, for calls into m2c?
- is 0.1.0 the right version to indicate that aren't explicitly versioning m2c?
- license: "GPL-3.0" is probably more right than "GPL-3.0-or-later"

simonlindholm:
- using m2c as a library is incredibly niche, we shouldn't put that towards the top of the README

nanaian:
> is there a way to make this work with src/, or is it needed to get m2c.main and similar import paths?

I'm pretty certain its required - if we make it `src` then you have to do `import src`.

> does mypy type-checking still work for decomp.me with this change, for calls into m2c?

AFAIK, yes, but I will double-check

> is 0.1.0 the right version to indicate that aren't explicitly versioning m2c?

Unsure, but a version is required

> license: "GPL-3.0" is probably more right than "GPL-3.0-or-later"

Will update to `GPL-3.0-only` if that's more correct

matt-kempster:
You can relegate the poetry documentation to the very bottom of the README (or near it). I agree that it doesn't deserve the real-estate up top.

nanaian:
OK, I've made the requested changes. To check that the pyproject.toml dependencies are actually valid, 899a43611ccf52bfb8b6992ce0944f175fab8fe3 updates CI to use the Poetry virtualenv for tests/mypy/black, but for whatever reason it doesn't appear to be running (perhaps changes to workflows need to be approved?).

nanaian:
Also- to check that the pyproject.toml dependencies are actually valid, a2429798129b8b478f6b5b4b676250e15eb8f97b updates CI to use the Poetry virtualenv for tests/mypy/black.

simonlindholm:
> Also- to check that the pyproject.toml dependencies are actually valid, https://github.com/matt-kempster/m2c/commit/a2429798129b8b478f6b5b4b676250e15eb8f97b updates CI to use the Poetry virtualenv for tests/mypy/black. I need approval to check the workflow runs correctly, though!

Nice! I clicked the "run workflow" button, but I think the yaml still gets taken from the master branch, not from the PR, for security reasons, so it may need the PR to be merged and iterated on in follow-up PRs in case it fails (which is fine). Or one could use some kind of "run GHA locally" tool to check it, I guess.

simonlindholm:
hm, it does look like the updated yaml ran though, and failed

README.md
```diff
@@ -376,3 +376,11 @@ Finally, `git add` your test to track it.
 ./run_tests.py --overwrite
 git add tests/end_to_end/my-new-test
 ```
+
+### Installatio nwith Poetry
```
simonlindholm, line 380:
```suggestion
### Installation with Poetry
```

pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+]
+license = "GPL-3.0-only"
+readme = "README.md"
+repository = "https://github.com/matt-kempster/m2c"
+packages = [{include = "m2c"}]
+
+[tool.poetry.dependencies]
+python = "^3.7"
+pycparser = "^2.21"
+graphviz = {version = "^0.20.1", optional = true}
```
simonlindholm, line 19: apparently `optional=true` dependencies are useless unless you have an "extras" section with an extra -> dep mapping that allows users to do `poetry install -E theextra`. So I guess either add one of those or make the dep non-optional.

simonlindholm:
>>  does mypy type-checking still work for decomp.me with this change, for calls into m2c?
>
> AFAIK, yes, but I will double-check

Apparently it did not, you needed a py.typed file for it to work (noticed this when looking at a silent splat/spimdisasm type-checking failure). Added in b52d92c2340f6f4ba1aafb464188bb698752fbb0.

PR #250, opened 2023-04-19
Updated manual.s

UEPuepUEP:
Added the msub instruction


simonlindholm:
Does PS2 support msub, or is it PSP only? Is msub an instruction that's actually emitted by compilers? If yes on both questions, this PR does seem useful as long as the test expectation is also updated, to serve as a reminder of what remains to be done for PS2 support. Though it would be even more useful to have a PR that actually implemented support, of course. :)

PR #251, opened 2023-04-19
Updated manual.s

UEPuepUEP:
Added the ext, ins, seb, and seh instructions


simonlindholm:
This is the wrong place for a test of these instructions to go, it would need to be somewhere PSP-specific. Should it be combined with #250?

PR #252, opened 2023-04-20
msub msubu instruction

UEPuepUEP:


m2c/arch_mips.py
```diff
@@ -1182,6 +1182,13 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
             outputs = [Register("hi"), Register("lo")]
             if args[0] != Register("zero"):
                 outputs.append(args[0])
+               
+            inputs = [args[1], args[2]]
+            if mnemonic in ("msub", "msubu"):
+                inputs.append(Register("lo"))
+            outputs = [Register("hi"), Register("lo")]
+            if args[0] != Register("zero"):
+                outputs.append(args[0])
```
simonlindholm, line 1191: Just extend the `mnemonic in ("madd", "maddu")` check above.

(overwriting the `inputs`/`outputs` that were calculated above would be wrong!)

simonlindholm (commented):
Can you add a test for this (e.g. extending ps2/manual.s), and run `./run_tests.py --overwrite` and commit the test expectation?

simonlindholm:
(I'm also still interested in whether msub/msubu are actually used in the real world.)

UEPuepUEP:
> Can you add a test for this (e.g. extending ps2/manual.s), and run `./run_tests.py --overwrite` and commit the test expectation?

So would I add msub + msubu to that file or open a new directory?

simonlindholm:
If that instruction is available for PS2 it's probably better to add it to that file, or if it's PSP only  a new one? It doesn't actually matter much though

PR #258, opened 2023-05-14
Fix `M2C_ERROR` when the message contains a comma

ribbanya:
Uses a variadic macro to ensure valid C syntax.

An alternative approach would be to emit either strings or comments within the macro, rather than unescaped C.

### DIFF
--- a/m2c/translate.py
+++ b/m2c/translate.py
@@ -818,7 +818,7 @@ class ErrorExpr(Condition):
 
     def format(self, fmt: Formatter) -> str:
         if self.desc is not None:
-            return f"M2C_ERROR({self.desc})"
+            return f"M2C_ERROR(/* {self.desc} */)"
         return "M2C_ERROR()"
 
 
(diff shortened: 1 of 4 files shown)


simonlindholm:
Downside of this is that IDO doesn't support `...`. Another alternative is to emit `M2C_ERROR((message))`; could also choose to change syntax only in valid-syntax mode and/or when the message contains a comma. Not sure what alternative I like best here...

ribbanya:
What do you think about wrapping the `M2C_ERROR` emissions (as in, the format strings) in a function, which wraps the message in a comment?

```py
def m2c_error(err):
    return f"M2C_ERROR(/* {error} */)"
```

simonlindholm:
M2C_ERROR emission is already localized into a function:
```py
        if self.desc is not None:
            return f"M2C_ERROR({self.desc})"
        return "M2C_ERROR()"
```
I think I'm fine going with the comment syntax

simonlindholm:
Just make sure to run `./run_tests.py --overwrite` to update test expectations, and I can merge tomorrow

ribbanya:
Okay, I don't have a local setup of any of the tested projects, so I'll have to do that first.

PR #260, opened 2023-07-04
Add K&R brace style option

r-burns:
doldecomp/melee puts function opening braces on their own lines (and no other braces). https://en.wikipedia.org/wiki/Indentation_style#K&R_style

This command-line option is mutually exclusive with the `--allman` option.

### DIFF
--- a/website.py
+++ b/website.py
@@ -40,6 +40,8 @@ if "source" in form:
         cmd.append("--no-casts")
     if "allman" in form:
         cmd.append("--allman")
+    if "knr" in form:
+        cmd.append("--knr")
     if "extraswitchindent" in form:
         cmd.append("--indent-switch-contents")
     if "leftptr" in form:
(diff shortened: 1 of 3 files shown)


website.py
```diff
@@ -40,6 +40,8 @@ def print_headers(content_type: str) -> None:
         cmd.append("--no-casts")
     if "allman" in form:
         cmd.append("--allman")
+    if "knr" in form:
```
simonlindholm, line 43: this either needs to be made an elif or the checkbox converted to a dropdown so you can't accidentally pass both options

simonlindholm (approved):

website.py
r-burns: Sounds good, I don't really know web stuff so I'll just switch this to an elif.

r-burns:
Thanks :)

PR #263, opened 2023-08-25
Fix ppc store update instructions

SeekyCt:
Currently, indexed store update instructions give an error along the lines of
```c
        // Error: Expected instruction argument to be of the form offset($register), but found $r3
        // At instruction: stwux $r9, $r3, $r5
```
due to the call of `a.memory_ref` in the eval_fn (which is only valid for the non-indexed versions).

This PR fixes this by splitting out the eval_fn into separate ones for the indexed and non-indexed versions like load update instructions do

One thing that's worth noting is that none of the current tests contain any of these instructions, and #187 doesn't actually list them; Super Paper Mario has some functions using the instructions if that helps (possibly means that only later CW versions actually emitted them)

### DIFF
--- a/m2c/arch_ppc.py
+++ b/m2c/arch_ppc.py
@@ -769,27 +769,41 @@ class PpcArch(Arch):
                 )
                 inputs = [args[0], args[1], args[2]]
                 outputs = [args[1]]
+
+                def eval_fn(s: NodeState, a: InstrArgs) -> None:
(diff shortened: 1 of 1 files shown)


simonlindholm:
Makes sense. Thanks! The note about Super Paper Mario is useful, sounds like we should use that for project tests if m2c works picks up more again in the future.

PR #266, opened 2023-09-01
fold_divmod mwcc mult_hi ordering fix

SeekyCt:
The final if statement of fold_divmod also assumes that right_expr will be the literal, which MWCC doesn't always follow. Currently a MULT_HI extracted from `MULT_HI(x, N) >> M` would be normalised to fix this, but if original_expr was the MULT_HI then it might not be in order. Fixing this removes a large chunk of the test failures for MWCC.

Afaik all remaining failures for finding MWCC divisions come down to three things, which I don't think I understand the maths well enough to fix:
- Denominators generated are exceeding `(1 << 29)`, which blocks the translation of `((x / N) + ((x / N) >> 31)) --> x / N`
- MWCC also uses inner subtraction in negative divisions
- The `round_div` fails for the MULT_HI constants used in -4 and -7

### DIFF
--- a/tests/end_to_end/gcc-division/mwcc-o4p-out.c
+++ b/tests/end_to_end/gcc-division/mwcc-o4p-out.c
@@ -9,19 +9,17 @@ void test(u32 a) {
 
 void test_s8(s8 c) {
     s8 sp8;
-    s8 temp_r4;
 
     sp8 = c;
     func_00400090((u32) ((s8) (u8) sp8 / 2));
-    func_00400090(MULT_HI(0x55555556, (s8) (u8) sp8) + ((s8) (u8) sp8 / 6442450941));
+    func_00400090((s8) (u8) sp8 / 3);
     func_00400090((s8) (u8) sp8 / 5);
     func_00400090((s8) (u8) sp8 / 7);
     func_00400090((s8) (u8) sp8 / 10);
     func_00400090((s8) (u8) sp8 / 100);
     func_00400090((s8) (u8) sp8 / 255);
     func_00400090((s8) (u8) sp8 % 2);
-    temp_r4 = (s8) (u8) sp8;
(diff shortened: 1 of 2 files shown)


tests/end_to_end/gcc-division/mwcc-o4p-out.c
```diff
@@ -387,14 +385,14 @@ void test_u32_mod(u32 u) {
...
     func_00400090((u32) (((u32) (sp8 - temp_r3_11) >> 1U) + temp_r3_11) >> 0x1EU);
-    func_00400090(MULTU_HI(2, sp8));
+    func_00400090(sp8 / 2147483648);
     temp_r3_12 = MULTU_HI(-3, sp8);
     func_00400090((u32) (((u32) (sp8 - temp_r3_12) >> 1U) + temp_r3_12) >> 0x1FU);
-    temp_r3_13 = MULTU_HI(3, sp8);
+    temp_r3_13 = sp8 / 1431655765;
     func_00400090((u32) (((u32) (sp8 - temp_r3_13) >> 1U) + temp_r3_13) >> 0x1FU);
-    func_00400090(sp8 / 4294967294);
+    func_00400090((u32) (sp8 / 2) >> 0x1FU);
```
simonlindholm, line 397: Hm, this is wrong, both before and after (it should be division by 0xFFFFFFFF)

tests/end_to_end/gcc-division/mwcc-o4p-out.c
```diff
@@ -305,16 +303,16 @@ void test_u32_div(u32 u) {
...
 }
 
 void test_u32_mod(u32 u) {
```
simonlindholm, line 318: I just noticed that this function tests the exact same as the div one, seems it was copy-pasted while forgetting to change / -> %.

simonlindholm: I pushed a fix for this. Wasn't able to update gcc-o2.s though since I've lost track of which compiler that corresponded to.

PR #270, opened 2024-03-21
mips eabi64 support

ethteck:
Very much a WIP, but this PR brings us from 

```c
void func_001099D8(void) {
    sceGsSetDefLoadImage(&D_002B8700, 0xE00, 8, 1);
    FlushCache(0);
    sceGsSyncPath(0, 0);
    sceGsExecLoadImage(&D_002B8700, &func_F20000);
    sceGsSetDefLoadImage(&D_002B8760, 0x1400, 8, 1);
    FlushCache(0);
    sceGsSyncPath(0, 0);
    sceGsExecLoadImage(&D_002B8760, &func_F20000 + 0x48000);
}
```
to the more accurate
```c
void func_001099D8(s64 arg0, s64 arg2, s64 arg4) {
    sceGsSetDefLoadImage(&D_002B8700, 0xE00, 8, 1, 0, 0, 0x200, 0xC0);
    FlushCache(0);
    sceGsSyncPath(0, 0);
    sceGsExecLoadImage(&D_002B8700, &func_F20000);
    sceGsSetDefLoadImage(&D_002B8760, 0x1400, 8, 1, 0, 0, 0x200, 0xC0);
    FlushCache(0);
    sceGsSyncPath(0, 0);
    sceGsExecLoadImage(&D_002B8760, &func_F20000 + 0x48000);
}
```

That being said, the signature of this function is broken now, and some actual code changes will probably be needed here


m2c/arch_mips.py
```diff
@@ -1848,3 +1851,284 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+            "s4",
+            "s5",
+            "s6",
+            "s7",
+            "f20",
+            "f22",
+            "f24",
+            "f26",
+            "f28",
+            "f30",
```
simonlindholm, line 1940: I imagine f21-f31 as well?

m2c/arch_mips.py
```diff
@@ -1848,3 +1851,284 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+    ]
+
+    all_regs = saved_regs + temp_regs + [stack_pointer_reg, Register("zero")]
+
+    aliased_gp_regs = {
+        "s8": Register("fp"),
+        "r0": Register("zero"),
+    }
+
+    o32abi_float_regs = {
```
simonlindholm, line 1955: this is not really o32 abi, is it? is there a reference for this list?

m2c/arch_mips.py
```diff
@@ -1848,3 +1851,284 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+            if slot.reg is None or slot.reg not in likely_regs:
+                continue
+
+            # Don't pass this register if lower numbered ones are undefined.
+            if slot == candidate_slots[0]:
+                # For varargs, a subset of regs may be used. Don't check
+                # earlier registers for the first member of that subset.
+                pass
+            else:
+                # Only r3-r10/f1-f13 can be used for arguments
```
simonlindholm, line 2111:
```suggestion
                # Only a1-a7/f13-f19 can be used for arguments
```

m2c/arch_mips.py
```diff
@@ -1848,3 +1851,284 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+            AbiArgSlot(28, Register("a7"), Type.any_reg()),
+        ]
+
+    def function_abi(
+        self,
+        fn_sig: FunctionSignature,
+        likely_regs: Dict[Register, bool],
+        *,
+        for_call: bool,
+    ) -> Abi:
```
simonlindholm, line 2053: Can you add a comment to this function and to ArchPpc.function_abi that they duplicate each other?

poetry.lock
```diff
@@ -84,63 +84,63 @@ files = [
```
simonlindholm, line 84: Any reason to update this file? It's a bit annoying to review changes to

simonlindholm (approved):
Sorry for extreme slowness in getting to this.

poetry.lock
ethteck: nope - just my compulsion to update things. will revert 😁

m2c/arch_mips.py
ethteck: you're right, it's n32, whoops. I renamed it accordingly

https://gist.github.com/EllipticEllipsis/27eef11205c7a59d8ea85632bc49224d#general-purpose-registers

m2c/arch_mips.py
ethteck: In the n32 abi, which I think we decided is closer to what eabi64 is, the odd ones aren't saved it seems. @AngheloAlf does this sound right to you?

simonlindholm: Empirically all of f20-f31 seem to be saved: https://decomp.me/scratch/f51s0

AngheloAlf: Indeed, odd registers are used and saved in eabi64.
We probably copy-pasted the n32 convention and forgot to update it.

ethteck: gotcha - thanks both of you for checking. updated

m2c/arch_mips.py
ethteck: We just pass the following flags to `as` for Kingdom Hearts: `-no-pad-sections -EL -march=5900 -mabi=eabi`. I am not totally sure how to answer your question unfortunately..

m2c/arch_mips.py
simonlindholm: Can you remove them from `simple_temp_regs` as well?

m2c/arch_mips.py
simonlindholm: It's a vague question, I was basically just hoping to get some more information about how these aliases work. I gather from experimenting with the flags you mention that `as` doesn't actually accept these register names as inputs, nor does objdump output them (`-Mreg-names=eabi` is not a thing). And the same seems to be true for integer registers.

So then, I take it all the names are just made up by us? And we made a decision to stay backwards compatible with o32 names, given the choice of naming $12 $t4 instead of $t0? Are eabi projects using some common prelude.inc that define these names for use with `as`, or is everyone sticking to numbered registers? Or are there projects assembling code using o32 register names? If so, should we support that, and make $t0 an alias for $a4?

Using the n32 naming of fp regs seems like a poor choice: it names $f21 $ft8 despite it being a saved register. I guess we have a few choices here:
- don't support abi names of fp regs
- use suboptimal names from o32 or n32 (but it's a bit sad to enshrine this in m2c, it makes it harder to change)
- come up with our set of names (and patching <https://github.com/decompals/binutils-mips-ps2-decompals>?)

ethteck: For now, I'm fine with option 1.. does it mean just getting rid of this dict?

simonlindholm: Yes, should be. Do you have any thoughts on the questions about integer registers?

simonlindholm:
I'll merge this and then push some follow-ups on top.

simonlindholm:
https://github.com/matt-kempster/m2c/commit/5bd755819b99d9deb632bc18d70c2a6eb2cc2f2f

PR #271, opened 2024-03-21
Add support for PS2's `paddub` and `sq`/`lq` instructions

AngheloAlf:
Compilers like MWCC 2.3 fairly commonly, and since they are kinda simple I thought it would be good to have them

- `paddub`: Behaves like `addu` but for quad registers. From what I have seen it is mostly used as a `move`
image
- `sq`/`lq`: Those behave like `sw` and `lw` but for quad registers. Most commonly used to save registers into the stack
image
image

### DIFF
--- a/m2c/arch_mips.py
+++ b/m2c/arch_mips.py
@@ -159,6 +159,7 @@ LENGTH_THREE: Set[str] = {
     "divu",
     "ddiv",
     "ddivu",
+    "paddub",
 }
 
 
@@ -881,6 +882,8 @@ class MipsArch(Arch):
                 return AsmInstruction("not", [args[0], args[1]])
             if instr.mnemonic == "addiu" and args[2] == AsmLiteral(0):
                 return AsmInstruction("move", args[:2])
+            if instr.mnemonic == "paddub" and args[2] == AsmLiteral(0):
+                return AsmInstruction("move", args[:2])
             if (
                 instr.mnemonic == "ori"
                 and args[1] == Register("zero")
@@ -1494,6 +1497,7 @@ class MipsArch(Arch):
         "addiu": lambda a: handle_addi(a),
         "add": lambda a: handle_add(a),
         "addu": lambda a: handle_add(a),
+        "paddub": lambda a: handle_add(a),
         "sub": lambda a: (
             fold_mul_chains(fold_divmod(BinaryOp.intptr(a.reg(1), "-", a.reg(2))))
         ),
--- a/m2c/arch_mips.py
+++ b/m2c/arch_mips.py
@@ -961,7 +961,7 @@ class MipsArch(Arch):
                 if loc is None:
                     return []
                 elif size == 16:
-                    return [loc, replace(loc, offset=loc.offset + 8)]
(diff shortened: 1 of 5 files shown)


m2c/arch_mips.py
```diff
@@ -957,6 +961,8 @@ def make_memory_access(arg: Argument) -> List[Location]:
                 loc = StackLocation.from_offset(arg.lhs)
                 if loc is None:
                     return []
+                elif size == 16:
+                    return [loc, replace(loc, offset=loc.offset + 8)]
                 elif size == 8:
                     return [loc, replace(loc, offset=loc.offset + 4)]
```
AngheloAlf, line 967: I want to bring attention here. I'm not sure if this is correct, I fiddled with it until m2c stopped complaining about gaps on the stack

m2c/types.py
```diff
@@ -927,6 +927,16 @@ def u64() -> Type:
...
+    @staticmethod
+    def s128() -> Type:
+        return Type(TypeData(kind=TypeData.K_INT, size_bits=128, sign=TypeData.SIGNED))
+
+    @staticmethod
+    def u128() -> Type:
+        return Type(
+            TypeData(kind=TypeData.K_INT, size_bits=128, sign=TypeData.UNSIGNED)
+        )
+
```
simonlindholm, line 939: no need for these (128-bit types aren't used for arithmetic so they don't make much sense)

m2c/types.py
```diff
@@ -942,6 +952,12 @@ def reg64(*, likely_float: bool) -> Type:
         likely = TypeData.K_FLOAT if likely_float else TypeData.K_INT
         return Type(TypeData(kind=kind, likely_kind=likely, size_bits=64))
 
+    @staticmethod
+    def reg128(*, likely_float: bool) -> Type:
```
simonlindholm, line 956: I'd remove the likely_float argument and use kind = K_INT, unless there are 128-bit floats

m2c/arch_mips.py
```diff
@@ -1494,6 +1501,7 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
         "addiu": lambda a: handle_addi(a),
         "add": lambda a: handle_add(a),
         "addu": lambda a: handle_add(a),
+        "paddub": lambda a: handle_add(a),
```
simonlindholm, line 1504: this is dubious (paddub isn't a 128-bit add), and also shouldn't be needed -- as long as we have the paddub -> move conversion that's enough

m2c/arch_mips.py
simonlindholm: My guess is that `[loc, replace(loc, offset=loc.offset + 4), replace(loc, offset=loc.offset + 8), replace(loc, offset=loc.offset + 12)]` is the right thing to return here

m2c/arch_mips.py
```diff
@@ -159,6 +159,7 @@
     "divu",
     "ddiv",
     "ddivu",
+    "paddub",
```
simonlindholm, line 162: similar to the handling of paddub below I think this could be removed too

PR #273, opened 2024-05-18
Add support for `seb`/`seh` instructions from mips32r2

Xeeynamo:
These instructions are found specifically on PlayStation Portable and they are backported to its processor from the mips32r2 specs.

`SEH` is a single instruction that replaces the combo `sll DST_REG, SRC_REG, 0x10` and `sra DST_REG, DST_REG, 0x10` to perform a `s16` cast. Same concept for `SEB` but when a cast to `s8` is required.

SEB instruction

SEH instruction

### DIFF
--- a/m2c/arch_mips.py
+++ b/m2c/arch_mips.py
@@ -62,6 +62,7 @@ from .translate import (
     as_intish,
     as_s64,
     as_sintish,
+    as_type,
     as_u32,
     as_u64,
     as_uintish,
@@ -86,8 +87,6 @@ from .evaluate import (
     handle_lwl,
     handle_lwr,
     handle_or,
-    handle_seb,
-    handle_seh,
     handle_sltiu,
     handle_sltu,
     handle_sra,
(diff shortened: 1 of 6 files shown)


m2c/evaluate.py
```diff
@@ -451,6 +451,16 @@ def handle_lwr(args: InstrArgs) -> Expression:
     return ErrorExpr("Unable to handle lwr; missing a corresponding lwl")
 
 
+def handle_seb(args: InstrArgs) -> Expression:
+    # Sign extend byte, it is a cast to s16
```
mkst, line 455: Is the comment wrong - cast to s8?

m2c/arch_mips.py
```diff
@@ -1662,6 +1664,9 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
         # Unaligned loads
         "lwl": lambda a: handle_lwl(a),
         "lwr": lambda a: handle_lwr(a),
+        # Sign extend
+        "seb": lambda a: handle_seb(a),
+        "seh": lambda a: handle_seh(a),
```
simonlindholm, line 1669: these are small enough that I'd suggest inlining them

simonlindholm (approved):

PR #274, opened 2024-05-19
pyproject.toml: Add `m2c` entrypoint when installed as package

abelbriggs1:
Currently, downstream users have to invoke some kind of Python script directly in order to run the CLI. This requires them to either write their own `m2c.py` wrapper script to invoke `m2c.main:main`, or manually download the repository (usually as a Git submodule).

This PR adds a freestanding command entrypoint, `m2c`, which can be run to invoke the `m2c` CLI simply by installing the Python package. This should streamline usage and dependency management for downstream users who don't use `m2c` as a library.


ethteck:
I think it'd be nicer to have the command be just `m2c` but this seems like a good idea

simonlindholm:
Huh, that was easy. Worth adding a note about it in the README?

Issue #272, opened 2024-05-13
How to properly define jumptables in PowerPC so that m2c can recognize them?


dbalatoni13:
There is no example for it and I'm not sure how to name them/where to place them so they get picked up by m2c.

dbalatoni13:
The problem was that m2c does not support the assembly format of dtk.

PR #275, opened 2024-05-28
Added compatibility for decomp-toolkit macros

dbalatoni13:
[encounter/decomp-toolkit](https://github.com/encounter/decomp-toolkit) outputs assembly files that are currently not fully compatible with m2c.
One problem is that labels have the format
```
.obj name, local
    .float 0
.endobj name
```
This makes m2c not detect any data, bss, rodata or jump table labels.
Another problem is that jump tables have the prefix "jumptable_".
Ultimately, strings use the type ".string".

### DIFF
--- a/m2c/asm_file.py
+++ b/m2c/asm_file.py
@@ -528,8 +528,14 @@ def parse_file(f: typing.TextIO, arch: ArchAsm, options: Options) -> AsmFile:
                         for w in args:
                             fval = try_parse(lambda: float(w))
                             asm_file.new_data_bytes(struct.pack(">d", fval))
-                    elif directive in (".asci", ".asciz", ".ascii", ".asciiz"):
-                        z = directive.endswith("z")
+                    elif directive in (
+                        ".asci",
+                        ".asciz",
+                        ".ascii",
+                        ".asciiz",
+                        ".string",
+                    ):
+                        z = directive.endswith("z") or directive == ".string"
                         asm_file.new_data_bytes(
                             parse_ascii_directive(line, z), is_string=True
                         )
@@ -548,6 +554,14 @@ def parse_file(f: typing.TextIO, arch: ArchAsm, options: Options) -> AsmFile:
                         data = parse_incbin(args, options, warnings)
                         if data is not None:
                             asm_file.new_data_bytes(data)
+                    elif directive == ".obj":  # decomp-toolkit label format
+                        parts = line.split()
+                        if len(parts) >= 3:
+                            try:
+                                kind = LabelKind[parts[2].upper()]
+                                process_label(parts[1].removesuffix(","), kind=kind)
+                            except KeyError:
+                                pass
 
         elif ifdef_level == 0:
             if directive == "jlabel":
--- a/m2c/asm_file.py
+++ b/m2c/asm_file.py
@@ -512,6 +512,13 @@ def parse_file(f: typing.TextIO, arch: ArchAsm, options: Options) -> AsmFile:
                         # ".rel name, label" expands to ".4byte name + (label - name)"
                         assert len(args) == 2
                         emit_word(args[1])
+                    elif directive == ".obj":
+                        # dtk disassembler label format
+                        assert len(args) == 2
+                        kind = (
+                            LabelKind.LOCAL if args[1] == "local" else LabelKind.GLOBAL
+                        )
+                        process_label(args[0], kind=kind)
                     elif directive in (".short", ".half", ".2byte"):
                         for w in args:
                             ival = try_parse(lambda: parse_int(w)) & 0xFFFF
@@ -554,14 +561,6 @@ def parse_file(f: typing.TextIO, arch: ArchAsm, options: Options) -> AsmFile:
                         data = parse_incbin(args, options, warnings)
                         if data is not None:
                             asm_file.new_data_bytes(data)
-                    elif directive == ".obj":  # decomp-toolkit label format
-                        parts = line.split()
-                        if len(parts) >= 3:
-                            try:
-                                kind = LabelKind[parts[2].upper()]
-                                process_label(parts[1].removesuffix(","), kind=kind)
-                            except KeyError:
-                                pass
 
         elif ifdef_level == 0:
             if directive == "jlabel":
--- a/m2c/flow_graph.py
+++ b/m2c/flow_graph.py
@@ -860,7 +860,7 @@ def build_graph_from_block(
                             and isinstance(arg.argument, AsmGlobalSymbol)
                             and any(
                                 arg.argument.symbol_name.startswith(prefix)
-                                for prefix in ("jtbl", "jpt_", "lbl_")
+                                for prefix in ("jtbl", "jpt_", "lbl_", "jumptable_")
                             )
                         ):
                             jtbl_names.add(arg.argument.symbol_name)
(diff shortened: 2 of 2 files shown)


m2c/asm_file.py
```diff
@@ -528,7 +530,13 @@ def process_label(label: str, *, kind: LabelKind) -> None:
                         for w in args:
                             fval = try_parse(lambda: float(w))
                             asm_file.new_data_bytes(struct.pack(">d", fval))
-                    elif directive in (".asci", ".asciz", ".ascii", ".asciiz"):
+                    elif directive in (
+                        ".asci",
+                        ".asciz",
+                        ".ascii",
+                        ".asciiz",
+                        "string",
```
simonlindholm, line 538: This won't work -- this section is guarded by `directive.startswith(".")`

m2c/asm_file.py
```diff
@@ -339,7 +339,9 @@ def re_comment_replacer(match: Match[str]) -> str:
     re_whitespace_or_string = re.compile(r'\s+|"(?:\\.|[^\\"])*"')
     re_local_glabel = re.compile("L(_.*_)?[0-9A-F]{8}")
     re_local_label = re.compile("loc_|locret_|def_|lbl_|LAB_|jump_")
-    re_label = re.compile(r'(?:([a-zA-Z0-9_.$]+)|"([a-zA-Z0-9_.$<>@,-]+)"):')
+    re_label = re.compile(
+        r'(?:(?:\.obj )?([a-zA-Z0-9_.$]+)|"([a-zA-Z0-9_.$<>@,-]+)")[:,]'
```
simonlindholm, line 343: Instead of complicating this regex, can you add a case near where jlabel/glabel/etc are handled?

simonlindholm:
Thanks, compatibility with more formats is always welcome. (Sorry for not responding in time to #272, I've been a bit strapped for time.)

m2c/asm_file.py
dbalatoni13: My bad, it should actually be ".string"

m2c/asm_file.py
dbalatoni13: I will give it a try.

dbalatoni13:
No problem!

dbalatoni13:
Does it look alright? I also made sure to set z to True for .string.

dbalatoni13:
Sorry, my previous edit deleted the jump table change, I'm not really experienced with git yet. I added it back now.

m2c/asm_file.py
```diff
@@ -548,6 +554,14 @@ def process_label(label: str, *, kind: LabelKind) -> None:
                         data = parse_incbin(args, options, warnings)
                         if data is not None:
                             asm_file.new_data_bytes(data)
+                    elif directive == ".obj":  # decomp-toolkit label format
+                        parts = line.split()
```
simonlindholm, line 558: can you make use of `args` instead, which is the arguments split by comma? we don't want to make unnecessary assumptions about whitespace here

m2c/asm_file.py
```diff
@@ -548,6 +554,14 @@ def process_label(label: str, *, kind: LabelKind) -> None:
...
+                        if len(parts) >= 3:
+                            try:
+                                kind = LabelKind[parts[2].upper()]
```
simonlindholm, line 561: LabelKind[...] exposes internal details in an ugly way and isn't searchable. Can you use some manual if-else switch instead?

m2c/flow_graph.py
```diff
@@ -862,7 +860,7 @@ def find_block_by_label(label: str) -> Optional[Block]:
                             and isinstance(arg.argument, AsmGlobalSymbol)
                             and any(
                                 arg.argument.symbol_name.startswith(prefix)
-                                for prefix in ("jtbl", "jpt_", "lbl_")
+                                for prefix in ("jtbl", "jpt_", "lbl_", "_jumptable")
```
simonlindholm, line 863: just double checking: `_jumptable` vs `jumptable_`?

dbalatoni13: You are right.

m2c/asm_file.py
dbalatoni13: Is it fine if I set it to global by default and change it to local if the string equals local?
I did some testing and my variables get marked as static not depending on the LabelKind at all, is that normal?

m2c/asm_file.py
dbalatoni13: sure

m2c/asm_file.py
simonlindholm: Sure, that's fine. A global label here means one that starts a function; for data it doesn't make a difference whether the label is global or not. And yeah I think m2c will happily mark things as static regardless of this, it's basing that decision on the context file instead.

dbalatoni13: I see.

m2c/asm_file.py
```diff
@@ -548,6 +554,12 @@ def process_label(label: str, *, kind: LabelKind) -> None:
                         data = parse_incbin(args, options, warnings)
                         if data is not None:
                             asm_file.new_data_bytes(data)
+                    elif directive == ".obj":  # decomp-toolkit label format
+                        if len(args) == 2:
```
simonlindholm, line 558: `assert len(args) == 2` instead to have some minimum of error handling. Or `raise DecompFailure("...")` in an else branch but that feels like effort.

m2c/asm_file.py
```diff
@@ -548,6 +554,12 @@ def process_label(label: str, *, kind: LabelKind) -> None:
...
+                            kind = LabelKind.GLOBAL
+                            if args[1] == "local":
+                                kind = LabelKind.LOCAL
```
simonlindholm, line 561:
```suggestion
                            kind = LabelKind.LOCAL if args[1] == "local" else LabelKind.GLOBAL
```
if you want

m2c/asm_file.py
```diff
@@ -548,6 +554,12 @@ def process_label(label: str, *, kind: LabelKind) -> None:
                         data = parse_incbin(args, options, warnings)
                         if data is not None:
                             asm_file.new_data_bytes(data)
+                    elif directive == ".obj":  # decomp-toolkit label format
```
simonlindholm, line 557: btw, maybe move this up next to the handling for .rel which is also related to decomp-toolkit?

simonlindholm (approved):

m2c/asm_file.py
dbalatoni13: sure, black adds brackets btw

m2c/asm_file.py
dbalatoni13: Okay, I will use assert

dbalatoni13:
I ran black and it was strange that it changed the formatting in other sections of this file and also multiple other files. I didn't commit those other changes, what do you think the issue probably is?

simonlindholm:
Test failure due to too old black version? Because I don't get reformatting in flow_graph.py locally.

simonlindholm:
23.12.1 is the version used in CI btw (mentioned in pyproject.toml), and 24.4.2 seems to be behave the same for me. You can probably just `python3 -m pip install --upgrade black`

simonlindholm:
Actually, nvm, 23.12.1 and 24.4.2 do not behave the same, I just had parallel install of black somehow. (depot_tools??? this is a new way of screwing up python installs that I heretofore was not aware of.) I'll push an update to master to make things simple.

simonlindholm:
so just rebase or merge with master and you should be good

dbalatoni13:
Oh, okay. I did a rebase.

PR #284, opened 2024-11-06
Add the posibility to type saved registers

dbalatoni13:
This is often useful when a void* is cast to a specific type and loaded into a saved register. It should be tested whether it causes issues in optimized code.

### DIFF
--- a/m2c/translate.py
+++ b/m2c/translate.py
@@ -476,8 +476,11 @@ class StackInfo:
     def maybe_get_register_var(self, reg: Register) -> Optional[Var]:
         return self.reg_vars.get(reg)
 
-    def add_register_var(self, reg: Register, name: str) -> None:
-        type = Type.floatish() if reg.is_float() else Type.intptr()
+    def add_register_var(self, reg: Register, name: str, given_type: Type) -> None:
+        if given_type is not None:
+            type = given_type
+        else:
+            type = Type.floatish() if reg.is_float() else Type.intptr()
         var = Var(self, prefix=f"var_{name}", type=type)
         self.reg_vars[reg] = var
         self.temp_vars.append(var)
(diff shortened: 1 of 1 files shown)


m2c/translate.py
```diff
@@ -4365,6 +4368,34 @@ def setup_planned_vars(
...
+        reg_struct_name, stack_info.global_info.typemap
+    )
+    if reg_struct is None:
+        return {}
+    types = {}
+    for reg_field in reg_struct.fields:
+        corresponding_regs = [
+            r
+            for r in regs
+            if reg_field.name.endswith(str(r).removeprefix("$")) and not r.is_float()
```
simonlindholm, line 4386:
```py
reg_name = stack_info.function.reg_formatter.format(reg)
name == f"var_{reg_name}"
```
I think. Why the `and not r.is_float()`?

Actually, hm, instead of hard-coding variable name, it might be cleaner if this function just parsed out a struct field name -> type mapping, and then `add_register_var` did the register -> struct field name lookup, since that is the one who knows the variable name.

m2c/translate.py
```diff
@@ -4365,6 +4368,34 @@ def setup_planned_vars(
             stack_info.planned_vars[key] = var
 
 
+def get_declared_reg_types(
+    stack_info: StackInfo, regs: List[Register]
+) -> [Dict[Register, Type]]:
```
simonlindholm, line 4373: this is invalid type syntax -- the [] need to be removed. Can you run mypy?

simonlindholm (commented):
Looks reasonable, although I think you might currently be ~the only user of --reg-vars so it's not going to see a lot of use. :) Extending it to `setup_planned_vars` or temp vars would make it a lot more useful, although, that looks a bit tricky, with the naming there being performed lazily...

m2c/translate.py
dbalatoni13: You are right. Sorry, I didn't get a notification so I completely forgot about this PR, I will get back to it once I have time.

dbalatoni13:
As I'm probably the only one using --reg-vars and doing it properly is tricky, I'd prefer working on my other ideas first!

PR #286, opened 2025-03-20
Added a pattern for load load store store struct copy on PPC

dbalatoni13:
With a new test case. If I'm not mistaken, this shouldn't have false posivites.

Being able to recognize struct copy with structs smaller than 8 bytes would require editing code somewhere else and in many cases having a correct context.

I wonder if the change to `parser` is fine.


simonlindholm (commented):
Looks nice! This only covers lengths 0, 1, 2 and 4 (mod 8), what about 3/5/6/7? Is this pattern also used for memcpy? Should we include the length in the called intrinsic?

dbalatoni13:
You're right, I'm going to add 3,5,6 and 7 (mod 8).
According to my testing memcpy (the one with a loop) is generated if the struct's size is more than 64 bytes. That one requires an IR pattern I think, because the order of the instructions is different on -O0 and -O4. 
As the whole struct is copied, I didn't see a reason to specify the length, but I can include it if you'd like.

dbalatoni13:
Bss/data section struct copies have a different pattern in -O4, but I'm not interested in that currently.

m2c/arch_ppc.py
```diff
@@ -265,6 +265,81 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+        "stw $b, (I+4)($d)",
+    )
+
+    def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
+        # Use the initial patterns first
+        m = matcher.try_match(self.pattern)
+        if m is None:
+            return None
+        i = 8
+        # Extend it only if necessary, increasing performance
```
simonlindholm, line 286:
```suggestion
```

(I don't see what the alternative to extending it like this would be)

m2c/arch_ppc.py
```diff
@@ -265,6 +265,81 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
         )
 
 
+class StructCopyPattern(AsmPattern):
+    """Recognizing struct copy when it starts with lwz lwz stw stw. Others
+    would cause false positves. Maybe we can find another way for those using
+    context?"""
```
simonlindholm, line 271: Can you write something here about what compiler versions/flags this applies to, to the extent you know about it? My first attempt at playing around with this pattern at decomp.me by picking compiler/compiler flags at random failed. :) (Of course, even then this is a good start and something we could extend in the future to cover more patterns as the need arises.)

simonlindholm (approved):
Looks great. Can you add the intrinsic to m2c_macros.h as well, or switch to one of the MEMCPY macros that are already mentioned there?

dbalatoni13:
Uhm those are either aligned or unaligned. I'm not sure which to pick. With this approach it would be hard to differentiate the intrinsic, right? I guess it might be better to create that new one.

m2c/arch_ppc.py
dbalatoni13: Yes, I'll experiment later today and update you.

simonlindholm:
From my experiments with decomp.me it seemed like if the alignment of the struct was 1 (by having only char members and no int ones) it would use only byte-level loads/stores, so the 2x lwz/stw at the start signal that the alignment is >= 4 I guess? But a new intrinsic is also fine.

dbalatoni13:
I'll experiment with it, but in the test case I included, I set the alignment to 1 and it's using 2x lwz/stw. Can you send a scratch?

simonlindholm:
Hesitant to create scratches for things that aren't actually real decomps (@ethteck it would be nice to have a compiler explorer mode), but I tried modifying https://decomp.me/scratch/eYyyU (a GameCube/Wii scratch with 4.3 build 151 (Wii MV 1.1) and -O4,p) adding the following at the end:
```c
#pragma push
#pragma pack(1)

typedef struct Test0 {
	int a;
	int b;
} Test0;

typedef struct Test0a {
	char a,b,c,d,e,f,g,h;
} Test0a;

#pragma pop

void test_0(Test0 *a, Test0 *b) {
	*a = *b;
}

void test_0_alt(Test0a *a, Test0a *b) {
	*a = *b;
}
```
and got the following output:
```asm
test_0:
lwz     r0,0(r4)
stw     r0,0(r3)
lwz     r0,4(r4)
stw     r0,4(r3)
blr     

test_0_alt:
lbz     r0,0(r4)
stb     r0,0(r3)
lbz     r0,1(r4)
stb     r0,1(r3)
lbz     r0,2(r4)
stb     r0,2(r3)
lbz     r0,3(r4)
stb     r0,3(r3)
lbz     r0,4(r4)
stb     r0,4(r3)
lbz     r0,5(r4)
stb     r0,5(r3)
lbz     r0,6(r4)
stb     r0,6(r3)
lbz     r0,7(r4)
stb     r0,7(r3)
blr     
```

dbalatoni13:
I see, I will check that out. Here's a scratch using Mario Party 4's preset: https://decomp.me/scratch/D20X1

dbalatoni13:
> Looks great. Can you add the intrinsic to m2c_macros.h as well, or switch to one of the MEMCPY macros that are already mentioned there?

What should it be defined as? Simply memcpy?

dbalatoni13:
I checked out your scratch and came to the conclusion that the difference is caused by C++.

simonlindholm:
> What should it be defined as? Simply memcpy?

I think aligned memcpy is the right one, given that your scratch changes codegen to copy byte-by-byte if you redefine the structs to only contain chars.

dbalatoni13:
Okay. What you're describing is because of -O0,p btw, in -O4,p it still uses load load store store.

simonlindholm:
> Okay. What you're describing is because of -O0,p btw, in -O4,p it still uses load load store store in that case.

Oh, okay. Then I guess a new M2C_STRUCT_COPY macro is fine, since we can't distinguish between aligned and unaligned. So your PR is good, it just needs to add the macro to m2c_macros.h.

dbalatoni13:
Okay. Should it be defined as dest = source?

simonlindholm:
Maybe define it as memcpy similar to M2C_MEMCPY_ALIGNED/UNALIGNED? Empty definition would also be fine

dbalatoni13:
Gotcha!

dbalatoni13:
I just realized that we have the source code of a similar version here: https://git.wuffs.org/MWCC/tree/compiler_and_linker/BackEnd/PowerPC/CodeGenerator/StructMoves.c?h=main

PR #287, opened 2025-03-20
Add option to disable the stack spilling logic

dbalatoni13:
Is the way I passed over the option fine?

### DIFF
--- a/m2c/translate.py
+++ b/m2c/translate.py
@@ -3191,7 +3191,6 @@ class NodeState:
     node: Node
     stack_info: StackInfo = field(repr=False)
     regs: RegInfo = field(repr=False)
-    stack_spill_detection: bool = field(repr=False)
 
     local_var_writes: Dict[LocalVar, Tuple[Register, Expression, bool]] = field(
         default_factory=dict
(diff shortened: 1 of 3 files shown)


simonlindholm:
I'm not a big fan of having the state live in so many places that are copies of each other. Can you put it on stack_info.global_info instead?

dbalatoni13:
Yeah, that would make more sense.

m2c/translate.py
```diff
@@ -3716,7 +3719,11 @@ def create_dominated_node_state(
     """
     stack_info = parent_state.stack_info
     new_regs = RegInfo(stack_info=stack_info)
-    child_state = NodeState(node=child, regs=new_regs, stack_info=stack_info)
+    child_state = NodeState(
+        node=child,
+        regs=new_regs,
+        stack_info=stack_info,
+    )
```
simonlindholm, line 3726:
```suggestion
    child_state = NodeState(node=child, regs=new_regs, stack_info=stack_info)
```

simonlindholm (approved):

simonlindholm:
Something I just realized the second after merging: we could update website.py as well to expose the option there as well (on https://simonsoftware.se/other/m2c.html). Since it's a niche option it's less important though.

dbalatoni13:
I can add it in the sorting PR if that's fine.

simonlindholm:
Sure!

dbalatoni13:
Only this one or the other two also?

simonlindholm:
The other two would be good too.

PR #288, opened 2025-03-21
Add options to sort bss backwards and local variables descending by register number

dbalatoni13:
These greatly help when decompiling Mario Party 4

### DIFF
--- a/m2c/if_statements.py
+++ b/m2c/if_statements.py
@@ -1467,7 +1467,10 @@ def get_function_text(function_info: FunctionInfo, options: Options) -> str:
                 type_decl = var.type.to_decl(var.format(fmt), fmt)
                 temp_decls.append(f"{type_decl};")
                 any_decl = True
-        for decl in sorted(temp_decls):
+
+        var_sort = (lambda d: d.split("_")[1]) if fmt.descending_regs else None
+
+        for decl in sorted(temp_decls, key=var_sort, reverse=fmt.descending_regs):
             function_lines.append(SimpleStatement(decl).format(fmt))
 
         for phi_var in function_info.stack_info.naive_phi_vars:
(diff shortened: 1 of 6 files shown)


m2c/if_statements.py
```diff
@@ -1467,7 +1467,10 @@ def get_function_text(function_info: FunctionInfo, options: Options) -> str:
                 type_decl = var.type.to_decl(var.format(fmt), fmt)
                 temp_decls.append(f"{type_decl};")
                 any_decl = True
-        for decl in sorted(temp_decls):
+
+        var_sort = (lambda d: d.split("_")[1][1:]) if fmt.descending_regs else None
```
simonlindholm, line 1471:
```suggestion
        var_sort = (lambda d: d.partition("_")[2]) if fmt.descending_regs else None
```
to include the full text after _ instead of just the second _-separated word, and include the first character of the register name in the sorting. Or was there a reason to drop it?

I guess float registers will still be sorted incorrectly, with f10 < f2.

dbalatoni13: Your suggestion seems nice to me. There wasn't a particular reason not to include first character. Regarding f10 < f2, that shouldn't be a big issue, as this feature is mainly so that saved registers are in the "right order" for specific MWCC environments. Do you have a suggestion though if it's bothering you?

simonlindholm: I think theoretically the right fix would be to put some ordering information in `Var`, getting it from Register at the point of construction. But let's leave that be for the moment.

Hm, wait, does my suggestion lead to interleaving
```
var_t1
temp_t1
var_t1_2
temp_t1_2
...
```
? Then your original version with `split` is better. I'd still drop the `[1:]` from it.

dbalatoni13: True, it would, okay.

dbalatoni13:
Sorry, I had a bit of a skill issue with git because I rebased during the PR 😅

dbalatoni13:
Ready to merge on my side.

PR #289, opened 2025-03-22
Comparison patterns for PPC

dbalatoni13:
I added a test case for how MWCC 1.3.2/2.6 handles a != 0 and a != b and code to handle it. In this PR I plan to fix all possible combinations from the previous and new test.

Feel free to give feedback to the commits in the meantime! As I wrote on Discord, I'm still learning when to use `early_unwrap_ints` over `early_unwrap`. I have a TODO comment and I'm also not sure why the input of `replace_clz_shift` is first passed to `fold_divmod`, should we do that here too?


dbalatoni13:
I'll add a comment about the version info and flags.

m2c/evaluate.py
```diff
@@ -860,6 +860,48 @@ def fold(
...
+        and or_left_expr.op == "-"
+        and isinstance(or_right_expr, BinaryOp)
+        and or_right_expr.op == "-"
+    ):
+        # (b - a) | (a - b)
+        a_left = early_unwrap(or_left_expr.right)
+        b_left = early_unwrap(or_left_expr.left)
+        a_right = early_unwrap(or_right_expr.left)
+        b_right = early_unwrap(or_right_expr.right)
+        # TODO check that it's not a float?
```
simonlindholm, line 898: sure, easy enough to add a check `not or_left_expr.is_floating() and not and_right_expr.is_floating()`. I don't think it's *actually* needed, but it's nice to keep the logic more obviously correct.

m2c/evaluate.py
```diff
@@ -860,6 +860,48 @@ def fold(
...
+        isinstance(or_left_expr, BinaryOp)
+        and or_left_expr.op == "-"
+        and isinstance(or_right_expr, BinaryOp)
+        and or_right_expr.op == "-"
+    ):
+        # (b - a) | (a - b)
+        a_left = early_unwrap(or_left_expr.right)
+        b_left = early_unwrap(or_left_expr.left)
+        a_right = early_unwrap(or_right_expr.left)
+        b_right = early_unwrap(or_right_expr.right)
```
simonlindholm, line 897: (I suspect these unwraps aren't needed, but I guess they don't hurt. I would inline them at the use site)

simonlindholm:
> I'm also not sure why the input of replace_clz_shift is first passed to fold_divmod, should we do that here too?

That's just a case of "try all the patterns one at a time", first fold_divmod, then replace_clz_shift, then (newly added) replace_or_shift. The way you've added it looks good to me.

m2c/evaluate.py
```diff
@@ -1155,6 +1197,8 @@ def handle_rlwinm(
 
         if simplify:
             lower_bits = replace_clz_shift(fold_divmod(lower_bits))
+            # This really only makes sense if the clz simplification didn't happen
```
simonlindholm, line 1200:
```suggestion
```

(this kind of "it only makes sense if so and so" is common, and the check for `>>` at the top of the function will take care of it)

simonlindholm (approved):
Looks good. Happy to merge this now and do the other comparisons in a follow-up if you feel like it, but also ok with awaiting those

m2c/evaluate.py
dbalatoni13: Okay, I was thinking of that.

m2c/evaluate.py
dbalatoni13: I'll add it.

dbalatoni13:
> That's just a case of "try all the patterns one at a time", first fold_divmod, then replace_clz_shift, then (newly added) replace_or_shift. The way you've added it looks good to me.
Oh, right. Should I leave it like this then or can I chain it one more time? I think the other new patterns will go to the same place, so I'm not sure.

dbalatoni13:
> Looks good. Happy to merge this now and do the other comparisons in a follow-up if you feel like it, but also ok with awaiting those

I won't take long, so I'd like to do it here to avoid cluttering PRs

simonlindholm:
> Oh, right. Should I leave it like this then or can I chain it one more time?

Either way is fine. Probably best to be consistent, i.e. either
```py
            lower_bits = replace_or_shift(replace_clz_shift(fold_divmod(lower_bits)))
```
or
```py
            lower_bits = fold_divmod(lower_bits)
            lower_bits = replace_clz_shift(lower_bits)
            lower_bits = replace_or_shift(lower_bits)
```
Maybe the latter is a bit more readable?

dbalatoni13:
> Maybe the latter is a bit more readable?
Yes, especially if considering that I'll have to add more.

dbalatoni13:
a < b and a > b generate the exact same code, I wonder which one I should go for.

dbalatoni13:
Hmm, if you look at a <= b in the comparison2 test, it's decompiled perfectly. But a >= b has casts, even though it's the exact same thing in the asm. This is how the expressions look like:
((arg1 >> 0x1F) + (arg0 >> 0x1FU) + M2C_CARRY)
(((s32) arg0 >> 0x1F) + ((u32) arg1 >> 0x1FU) + M2C_CARRY)

Do you think we can do something about that safely?

m2c/evaluate.py
```diff
@@ -952,6 +952,34 @@ def replace_xor_shift(expr: BinaryOp) -> BinaryOp:
...
+    These are generated by GW MW 2.6 for example.
+    """
+
+    if expr.is_floating():
+        return expr
+    
+    left_expr = early_unwrap_ints(expr.left)
+    if not (isinstance(left_expr, BinaryOp) and left_expr.op == "+"):
+        return expr
+    # We call this function only from carry_add_to so it's not necessary to check the right side
```
dbalatoni13, line 968: Now that I think about it, the expression might be changed by fold_divmod by this point. I think I should either change the order or just add an extra check here.

simonlindholm:
Btw, it would be worth adding a test for unsigned comparisons as well.

dbalatoni13:
Oh wow, that generates code that uses an instruction that's not even implemented yet.

dbalatoni13:
Actually, let's look at the patterns from the old comparison test and some other time. Looking forward to your feedback to merge this :)

simonlindholm:
I'll take a look tomorrow!

dbalatoni13:
Thanks! I just had this idea: if we know that a certain pattern is only generated if the args are u32 for example, can't we use this information to guess the types better? 🤔

dbalatoni13:
May I ask how to do that? :D

dbalatoni13:
I feel like `arg0 >> 31` and `(arg0 >> 31) ^ 1` might be too short to not create false positives. On the other hand: should I implement these two?:
```
temp_r0 = CLZ(arg0);
    global = ((1 << (temp_r0 & 0x1F)) & 1) | ((1 >> (0x20 - (temp_r0 & 0x1F))) & 1);
    global = (u32) (-(s32) arg0 & ~arg0) >> 0x1FU;
 ```

**Comment by dbalatoni13** [comment]
I ran this through the whole Mario Party 4 decomp and these new patterns caused 30 files to be changed. Only one of them is a false positive, but that function is a mess anyways. I noticed that the order in a < b actually matters when matching, because of the order the values are loaded in, we've sadly lost this info by the point we handle the expression as a whole.

**Comment by simonlindholm** [comment]
> May I ask how to do that? :D Though the problem is that in the a < b case we don't know which one from the two is u32. It generates this if either of them is u32.

I would suggest using as_uintish() on both operands, which will try to coerce them both to unsigned, falling back to a cast if it fails. It might be wrong (as you say only one has to be u32 for C to auto-convert to u32) but it still feels like a better guess than nothing at all?

> On the other hand: should I implement these three?:

Well, sure, sounds good to me. Do you want to do it in the same PR or another? (I'm already taking too long to get around to reviewing this one I feel like...)

File: tests/end_to_end/comparison2/orig.c [hunk 5]
```diff
@@ -0,0 +1,101 @@
...
+    global = (a >= 0);
+}
+
+void test(void) {
+    test_s32(1, 2, 3);
+    test_u32(1, 2, 3);
+    test_s16(1, 2, 3);
+    test_u16(1, 2, 3);
+    test_s8(1, 2, 3);
+    test_u8(1, 2, 3);
```
simonlindholm, line 100: looks like s8 and u8 tests give the same output as s16/u16, should they be removed?

m2c/evaluate.py
```diff
@@ -860,6 +895,125 @@ def fold(
...
+    elif early_unwrap(sub_left_xor_expr.right) == sub_right_right_expr:
+        # (b ^ a) ... & a
+        return BinaryOp.icmp(sub_left_xor_expr.left, "<", sub_left_xor_expr.right)
+
+    return expr
+
+
+def replace_shift_add_carry(expr: BinaryOp) -> BinaryOp:
+    """
+    Simplify the expressions matching `((b >> 31) + (a >> 31U)) + M2C_CARRY`
```
simonlindholm, line 992: (Hm, I wish M2C_CARRY expressions kept the addition/subtraction that resulted in the carry, it's tricky to understand why the pattern makes sense right now)

m2c/evaluate.py
```diff
@@ -860,6 +895,125 @@ def fold(
...
+
+    left_expr = early_unwrap_ints(expr.left)
+    if not (isinstance(left_expr, BinaryOp) and left_expr.op == "+"):
+        return expr
+
+    left_left_expr = early_unwrap(left_expr.left)
+    left_right_expr = early_unwrap(left_expr.right)
+    if not (
+        isinstance(left_left_expr, BinaryOp)
+        and early_unwrap(left_left_expr.right) == Literal(31)
```
simonlindholm, line 1008: It should generally be fine to skip early_unwrap's when comparing against literals, to the extent they exist wrappers around literals are generally removed very eagerly automatically and you never see them. Not that it hurts, but still

simonlindholm:
Actually, for comparisons specifically you can use BinaryOp.ucmp() for uint compares, which is shorthand for doing as_uintish() on each argument.

tests/end_to_end/comparison2/orig.c
dbalatoni13: Sure!

m2c/evaluate.py
dbalatoni13: Okay

simonlindholm:
The topic of PPC comparison patterns came up again a few months back in the GC/Wii decomp server's m2c channel, upon which I added support using IR patterns (similar to asm patterns but they support instruction reordering): https://github.com/matt-kempster/m2c/blob/e07f7e1c1d1d52f33ab3f7fedbe74b634404b943/m2c/arch_ppc.py#L231-L347 ... and only now did I come across this PR again and remember that it existed. Oops. I just pushed 708028435a2a5695fc0d06e14459995f58b064b6 to get coverage for the last two patterns from the tests added here, seemingly a compiler version diff.

Will close this, but thanks for the PR nonetheless!

PR #294, opened 2025-03-29
MWCC Struct copy improvements

dbalatoni13:
- Adjusted the pattern to make it possible for addresses to not start at zero
- Sometimes the last part uses different registers, this was adjusted in the pattern
- Use a store statement instead of the void function we had till now.  This helps with type resolution.


m2c/arch_ppc.py
```diff
@@ -1097,6 +1105,29 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+                assert isinstance(size_expr, Literal)
+                size = size_expr.value
+
+                stack_info = a.stack_info
+                dest = deref(dest_addr, a.regs, stack_info, size=size, store=False)
+                source = deref(source_addr, a.regs, stack_info, size=size, store=False)
+                if dest.type.is_struct() or source.type.is_struct():
+                    dest.type.unify(source.type)
+
+                # TODO are these four needed?
```
dbalatoni13, line 1124: Are these needed here? I copied them from another place

m2c/translate.py
```diff
@@ -2075,7 +2076,13 @@ def format(self, fmt: Formatter) -> str:
         ) or isinstance(dest, (ArrayAccess, LocalVar, SubroutineArg)):
             # Known destination; fine to elide some casts.
             source = elide_literal_casts(source)
-        return format_assignment(dest, source, fmt)
+        line = format_assignment(dest, source, fmt)
+        if self.commented_size is None:
+            return line
+        else:
+            return fmt.with_comments(
+                line, [f"size 0x{fmt.format_hex(self.commented_size)}"]
```
dbalatoni13, line 2084: This causes an extra level of identation, do you have a suggestion maybe how we could get rid of that?

dbalatoni13:
TODO: There's a bug that causes the copy to go onto `->unk0` instead of the argument.

dbalatoni13:
We do support addresses not starting at 0, but I wonder if that causes false positives on -O4.

dbalatoni13:
It also seems to make no difference whether we manually call `dest.type.unify(source.type)` in the code. It probably already happens somewhere else.

tests/end_to_end/struct-copy2/mwcc-o0p-out.c
```diff
@@ -5,35 +5,35 @@ void test(void) {
 
 }
 
-void test_0(s32 arg0, s32 arg1) {
-    M2C_STRUCT_COPY(arg0, arg1, 8);
+void test_0(? *arg0, ? *arg1) {
```
dbalatoni13, line 8: We should probably get these to be typed as pointers.

m2c/translate.py
```diff
@@ -2075,7 +2076,14 @@ def format(self, fmt: Formatter) -> str:
         ) or isinstance(dest, (ArrayAccess, LocalVar, SubroutineArg)):
             # Known destination; fine to elide some casts.
             source = elide_literal_casts(source)
-        return format_assignment(dest, source, fmt)
+        line = format_assignment(dest, source, fmt)
+        # Used in struct copy statements
```
simonlindholm, line 2080: Drop this comment, it's mostly irrelevant it is used and it's confusing what this comment refers to exactly

m2c/translate.py
simonlindholm: Eh, it's fine.

It looks like fmt.with_comments gives additional indentation though, we need to fix that somehow.

tests/end_to_end/struct-copy2/mwcc-o0p-out.c
simonlindholm: ?* is a pointer

m2c/translate.py
simonlindholm: Oh, wait, I must have been really tired when reading this, I thought you were referring to Python indentation...

I think the comments probably have to be carried all the way to if_statements somehow. Maybe Statement can grow a method for getting comments from a statement (defaulting to []), and SimpleStatement can check the output of that method?

PR #296, opened 2025-08-13
Add a special `#ifdef` for decomp.me convenience

thefoxcam:
This addresses a longstanding issue particularly affecting users trying to match Melee code on the m2c + decomp.me setup, where it's desirable to have the same context file able to both provide the correct type to m2c for the `user_data` of our custom `*_GObj` structs, as well as typedef the `*_GObj` structs back to `HSD_GObj`. Since it's undesirable to run a C preprocessor in m2c, a regular expression is used that mimics the functionality of an `#ifdef`, but only specific expressions are allowed to be used with it. Currently the only expression supported is hardcoded to be `M2CTX_DUAL`. Use of any other expression will raise a `Directives not supported yet` error like it currently does. The `#ifdef` also must contain an `#else` clause. For instance:

```c
#ifdef M2CTX_DUAL
typedef struct Fighter_GObj Fighter_GObj;
struct Fighter_GObj {
    /* +2C */ Fighter* user_data;
};
#else
typedef struct HSD_GObj Fighter_GObj;
#endif
```

If this kind of functionality is acceptable, I can write documentation for it, as other games may find this useful as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,20 @@ python3 ./m2c.py --visualize=c --context ctx.c -f my_fn my_asm.s > my_fn_c.svg
 python3 ./m2c.py --visualize=asm --context ctx.c -f my_fn my_asm.s > my_fn_asm.svg
 ```
 
+### Preprocessing
+
+There currently is a psuedo-macro in lieu of a full preprocessor that allows for the conditional switching of code in a context file. This allows for both m2c and i.e. a compiler to use the same context file if both need to define i.e. structs or typedefs slightly differently.
+
+```c
+#ifdef M2C
+...
+#else
+...
+#endif
+```
+
+Any other macros besides `#ifdef M2C` currently will fail, and you also need the `#else` between `#ifdef M2C` and `#endif` for the pattern to match.
+
 ### Migrating from `mips_to_c.py`
 
 This tool was originally known as `mips_to_c`. As part of the rename, deprecated command line arguments were removed.
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ python3 ./m2c.py --visualize=asm --context ctx.c -f my_fn my_asm.s > my_fn_asm.s
 
 ### Preprocessing
 
-There currently is a psuedo-macro in lieu of a full preprocessor that allows for the conditional switching of code in a context file. This allows for both m2c and i.e. a compiler to use the same context file if both need to define i.e. structs or typedefs slightly differently.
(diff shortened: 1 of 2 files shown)


simonlindholm:
This sounds like a nice idea to me! Can we let the macro name be just `M2C` maybe?

I am thinking of just running the preprocessor for real, but haven't yet looked into what options exist for e.g. disallowing includes, and it might take a little bit until I get time to look. Merging this in the meantime sounds like a good plan.

thefoxcam:
Yeah, sure. I'm a bit hesitant to make the name look like a "real" macro since you can only use it in this specific pattern, but since Melee will likely only be using it until a full preprocessor gets implemented and it'll get changed anyways, I'll change it. 

Here's some information on preprocessors (which also was discussed recently in the Melee decomp channel):

- The command-line tool for pcpp supports ignoring includes with `--passthru-includes <regex>`. For more control/speed the library also allows you to subclass the `Preprocessor` class directly, which I did in the [Melee context builder](https://github.com/doldecomp/melee/pull/1838/files#diff-5ffc9cb66e8e444c8e3c2dd1c7e7acfefd51171d8cba96f2ea99f3a95df10d96) to produce a context file that only contains the allowed macro. This is basically a compressed vesrion of the pcpp CLI code, found [here](https://github.com/ned14/pcpp/blob/master/pcpp/pcmd.py).

- When you do implement the full preprocessor it would be nice if the macro was defined by default either in m2c itself or decomp.me, since there's no way to adjust m2c flags from the decomp.me UI currently. Someone was working on a [PR](https://github.com/decompme/decomp.me/pull/1508) for that a few months ago, but it seems to be stalled.

README.md
```diff
@@ -271,6 +271,20 @@ python3 ./m2c.py --visualize=c --context ctx.c -f my_fn my_asm.s > my_fn_c.svg
 python3 ./m2c.py --visualize=asm --context ctx.c -f my_fn my_asm.s > my_fn_asm.svg
 ```
 
+### Preprocessing
+
+There currently is a psuedo-macro in lieu of a full preprocessor that allows for the conditional switching of code in a context file. This allows for both m2c and i.e. a compiler to use the same context file if both need to define i.e. structs or typedefs slightly differently.
```
simonlindholm, line 276:
```suggestion
There currently is a pseudo-macro in lieu of a full preprocessor that allows for the conditional switching of code in a context file. This allows for both m2c and e.g. a compiler to use the same context file if both need to define e.g. structs or typedefs slightly differently.
```

simonlindholm (approved):

PR #300, opened 2025-08-27
ARM: Fix subroutines getting appended with '/* extra? */' and 'Unable to find stack arg <offset> in block' for >= 4 args.

JaceCear:
I'm not sure whether this could break in some other place, but it seems like the correct behavior to me.


simonlindholm:
Thanks! I'm also a bit hazy on the details of this area of m2c tbh, too much inherited complexity from MIPS... but it makes sense to me.

PR #301, opened 2025-10-16
Add flag for specifying union fields

lukechampine:
Before:
```c
void ftKb_SpecialNNs_800FEDD0(Fighter_GObj* gobj)
{
    Fighter* temp_r31;
    Fighter* temp_r6;
    void* temp_r5;

    temp_r31 = gobj->user_data;
    Fighter_ChangeMotionState(gobj, 0x1B7, 0U, 0.0f, 1.0f, 0.0f, NULL);
    temp_r31->cmd_vars[3] = 0;
    temp_r31->cmd_vars[2] = 0;
    temp_r31->cmd_vars[1] = 0;
    temp_r31->cmd_vars[0] = 0;
    temp_r31->self_vel.y = 0.0f;
    temp_r6 = gobj->user_data;
    temp_r5 = temp_r6->dat_attrs;
    temp_r6->mv.ca.specials.grav = temp_r5->unk1A0;
    temp_r6->mv.ca.specialhi.vel.x = temp_r5->unk1A4;
    temp_r6->mv.ca.specialhi.vel.y = temp_r5->unk1A8;
    temp_r6->fv.kb.ns_flash_gobj = NULL;
    temp_r6->mv.co.walk.middle_anim_frame = temp_r5->unk1AC;
    ftAnim_8006EBA4(gobj);
}
```
After:
```c
// --union-field Fighter_FighterVars:kb \
// --union-field Fighter_MotionVars:kb \
// --union-field ftKb_MotionVars:specialn_ns \
// --void-var-type temp_r5:ftKb_DatAttrs

void ftKb_SpecialNNs_800FEDD0(Fighter_GObj* gobj)
{
    Fighter* temp_r31;
    Fighter* temp_r6;
    ftKb_DatAttrs* temp_r5;

    temp_r31 = gobj->user_data;
    Fighter_ChangeMotionState(gobj, 0x1B7, 0U, 0.0f, 1.0f, 0.0f, NULL);
    temp_r31->cmd_vars[3] = 0;
    temp_r31->cmd_vars[2] = 0;
    temp_r31->cmd_vars[1] = 0;
    temp_r31->cmd_vars[0] = 0;
    temp_r31->self_vel.y = 0.0f;
    temp_r6 = gobj->user_data;
    temp_r5 = temp_r6->dat_attrs;
    temp_r6->mv.kb.specialn_ns.frames_to_loop_charge_ground = temp_r5->specialn_ns_frames_to_loop_charge_ground;
[...]

PR #302, opened 2025-10-16
Add void-var-type flag

lukechampine:
Split out of #301


m2c/main.py
```diff
@@ -662,6 +674,23 @@ def parse_flags(flags: List[str]) -> Options:
...
+            print(
+                f"Error: Invalid void variable type override '{override}'. "
+                "Expected format: VAR_NAME:TYPE_NAME",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        var_name, type_name = parts
+        # Add implicit * if not already present (unless it's already a pointer type)
+        if not type_name.endswith("*"):
+            type_name = type_name + "*"
```
lukechampine: this is a bit controversial; the justification is that, unless you wrap it in quotes, `*` will be interpreted as a glob by your shell. I don't love this solution, though, so I'm open to better suggestions.

lukechampine:
fixed the test regressions 👍🏻

simonlindholm:
(I'll take a look at this once I find time!)

lukechampine:
Hmm. I'm only now realizing that there isn't always a variable you can reference. For example:
```c
void itLinkarrow_UnkMotion1_Phys(Item_GObj* gobj)
{
    Item* ip = GET_ITEM(gobj);
    f32 var_f1 = ip->xC4_article_data->x4_specialAttributes->unk1C;
    ip->x40_vel.y -= var_f1;
}
```
Here, the `x4_specialAttributes` value is a `void*` that should be cast to `itLinkArrowAttributes*`. But the decompiler reads straight through it into `unk1C`, so there's no intermediate variable we can fix. Do you have any thoughts on how to handle situations like this?

lukechampine:
thanks! I did indeed end up adding a struct field version of this on my fork: https://github.com/lukechampine/m2c/commit/374f69b7c300c0c97f8beadce30fd331f186da49
I prefer this approach over naming vars anyway, since you can do it preemptively instead of "decompile, find the name of the variable, decompile again."

Claude implemented it, so I suspect the approach is far from optimal, but it does appear to work ‾\\\_(ツ)\_/‾

PR #312, opened 2026-05-26
Implement IDO signed division by power of two with a large addend pattern

Drahsid:
We spotted an unusual instance of `lui ... addiu` in some Mischief Makers code. I was able to match it with a division, and noticed it was related to power of two.

The way this is different from existing cases is that it appears when there is another immediate/literal in the operation. In our circumstance, it was this: `(gActors[arg0].unk_160 - 0x400000) / 32`

The way this works within ido is that ugen lowers signed constant division to a pseudo-instruction which is later expanded by as1, at which point, it explicitly uses lui + addiu for power of two division.

I tried to be careful with not having false positives, so the code might be a bit more verbose than it needs to be. I also added multiple tests to make sure it worked in more complicated scenarios.
I'm pretty sure this pattern is unique to IDO, so I gated it.


Drahsid:
> **P1 Badge Preserve clobber when addend register equals bias register**
> 
> This rewrite is unsafe when the addend register is also the `addiu` bias destination (commonly `$at`): the original sequence overwrites that register on the negative path, but the replacement re-emits `li` and then leaves the register holding the constant. Any later read of that register after the label can decompile incorrectly (e.g., `move $v0, $at` becomes a constant return). A quick repro with this exact pattern shows `IdoDivP2LargeAddendPattern` simplifying to `return -0x400000;`, which is not equivalent to the original conditional value.


This made me realize there was a preexisting bug. Consider:
```
glabel test
/* 000000 3C01FFC0 */  lui   $at, 0xffc0
/* 000004 00811021 */  addu  $v0, $a0, $at
/* 000008 04410003 */  bgez  $v0, .L0018
/* 00000C 00027143 */   sra   $t6, $v0, 5
/* 000010 2441001F */  addiu $at, $v0, 0x1f
/* 000014 00017143 */  sra   $t6, $at, 5
.L0018:
/* 000018 03E00008 */  jr    $ra
/* 00001C 00201025 */   move  $v0, $at
```

The current version of m2c incorrectly outputs:
```c
s32 test(s32 arg0) {
    return 0xFFC00000;
}
```

This is missing the control flow.

Drahsid:
I fixed the aforementioned bug, and also covered some previously missed ido cases in these two commits

simonlindholm:
This looks very complicated. (Is it AI-generated?) Generally asm patterns are meant for simple pieces of sequential asm; when the patterns start to be non-consecutive getting all checks for equivalence right starts getting too complex for the low level that they operate on. As alternatives we have IR patterns and handling in the translation layer, I wonder if that could be used here instead. Admittedly that is unable to deal with branches at the moment, though it would be nice to grow support for that for IR patterns.

Let me ask first though, what are some examples of the asm you're trying to match?

Drahsid:
This originally started as 2 much more simple patterns but I realized that there could be some rare circumstances where the instructions rescheduled and it ended up becoming more complex. I ended up grafting what I had to cover new cases over time. Considering this is already a rare care, it may make more sense to only use basic patterns to keep it simple.

You can find the initial example which lead me to investigate this within this scratch: [https://decomp.me/scratch/qubf4](https://decomp.me/scratch/qubf4) on line 129, offset 8EC. This can be observed if you compile any similarly shaped code (large immediate, power of 2 denominator)

I didn't use AI for this (I suppose beside the random codex PR review jumpscare)

simonlindholm:
> You can find the initial example which lead me to investigate this within this scratch: https://decomp.me/scratch/qubf4 on line 129, offset 8EC. This can be observed if you compile any similarly shaped code (large immediate, power of 2 denominator)

Aha. This looks like it's actually covered by the existing DivP2Pattern1 pattern:
```py
    pattern = make_pattern(
        "bgez $i, .A",
        "sra $o, $i, N",
        "addiu $at, $i, ((1 << N) - 1)",
        "sra $o, $at, N",
        ".A:",
    )
```
but the `addiu` in your asm is `%lo(D_FFC0001F)` rather than 0x1f, so it doesn't fire. (I don't like the idea of having m2c make the assumption that D_FFC0001F == 0xffc0001f.) I suppose this is a spimdisasm bug. The guess of 0xffc0001f as being a symbol is wrong, since the addiu happens conditionally:
```asm
    /* 3D814 8003CC14 3C01FFC0 */  lui        $at, %hi(D_FFC0001F)
    /* 3D818 8003CC18 8CEB0164 */  lw         $t3, 0x164($a3)
    /* 3D81C 8003CC1C 00616021 */  addu       $t4, $v1, $at
    /* 3D820 8003CC20 05810003 */  bgez       $t4, .L8003CC30
    /* 3D824 8003CC24 000C7943 */   sra       $t7, $t4, 5
    /* 3D828 8003CC28 2581001F */  addiu      $at, $t4, %lo(D_FFC0001F)
    /* 3D82C 8003CC2C 00017943 */  sra        $t7, $at, 5
  .L8003CC30:
```
@AngheloAlf 

> I didn't use AI for this (I suppose beside the random codex PR review jumpscare)

Sorry about that (and about the suspicion), it's not something I've enabled intentionally. Might just be something GitHub forces upon everyone nowadays... Maybe Quate can turn it off.

Do you have other cases of division that have come up in practice that this PR fixes?

Drahsid:
> Do you have other cases of division that have come up in practice that this PR fixes?

No examples from any game disassembly, just synthetic tests I wrote after understanding the pattern.

Out of the tests I added, these have questionable results in current m2c:
- `tests/end_to_end/ido-division-large-addend/compiled-loop-label.s`, `tests/end_to_end/ido-division-large-addend/compiled-reuse-addend.s`, `tests/end_to_end/ido-division-large-addend/compiled-scheduled-after.s`, `tests/end_to_end/ido-division-large-addend/manual.s`, `tests/end_to_end/ido-division-large-addend/no-wildcard.s`, `tests/end_to_end/ido-division-large-addend/scheduled-two-out.c`
It correctly determines the div pattern in these but the immediate and arithmetic is flipped (not exactly wrong output, just not as clear)
- `tests/end_to_end/ido-division-large-addend/compiled-shift16.s` and `tests/end_to_end/ido-division-large-addend/compiled-shift17.s`
Fails to recover the div entirely and shows the division pattern instead.
- `tests/end_to_end/ido-division-large-addend/clobbered-bias-read.s`
Is missing the entire control flow; not exactly the issue I was targeting but is a bug. Maybe worth splitting out, though this only seems to occur in tiny functions.


> Sorry about that (and about the suspicion), it's not something I've enabled intentionally. Might just be something GitHub forces upon everyone nowadays... Maybe Quate can turn it off.

It may have been my end, I do have a codex subscription, though I never intentionally enabled such a feature.

AngheloAlf:
> but the `addiu` in your asm is `%lo(D_FFC0001F)` rather than 0x1f, so it doesn't fire. (I don't like the idea of having m2c make the assumption that D_FFC0001F == 0xffc0001f.) I suppose this is a spimdisasm bug. The guess of 0xffc0001f as being a symbol is wrong, since the addiu happens conditionally:
> 
> ```assembly
>     /* 3D814 8003CC14 3C01FFC0 */  lui        $at, %hi(D_FFC0001F)
>     /* 3D818 8003CC18 8CEB0164 */  lw         $t3, 0x164($a3)
>     /* 3D81C 8003CC1C 00616021 */  addu       $t4, $v1, $at
>     /* 3D820 8003CC20 05810003 */  bgez       $t4, .L8003CC30
>     /* 3D824 8003CC24 000C7943 */   sra       $t7, $t4, 5
>     /* 3D828 8003CC28 2581001F */  addiu      $at, $t4, %lo(D_FFC0001F)
>     /* 3D82C 8003CC2C 00017943 */  sra        $t7, $at, 5
>   .L8003CC30:
> ```
> 
> @AngheloAlf

I agree spimdisasm shouldn't be pairing those two.
I'm working on a fix, but idk when I'll have it ready because the naive approach broke some edge cases.
Hopefully I can have it ready tomorrow.

simonlindholm:
> tests/end_to_end/ido-division-large-addend/clobbered-bias-read.s

Ah yes, this happens because we match the division pattern and remove it, but then we read from $at at the end, the the assignment to it has been removed. We do mark the division instruction as clobbering $at in this case, but we don't act on that in any way beyond printing a warning if you run with --debug ("in test, regs were read before being written to"). Should probably actually apply clobbers in translate.py, so that this generates an error about a read from unset $at, rather than silent bad codegen...

We could prevent the division pattern from matching by checking that the output register of `addiu $at, $v0, 0x1f` has only one use before accepting the match. I don't feel like it's really worth doing though, I don't think this ever comes up in compiler-generated asm (it would require the compiler to use $at as something other than an assembler temporary). There are other bugs that could be fixed in the existing patterns, too, like failing to check `input reg != output reg` in DivP2Pattern1, and `M == (1 << N) - 1` in DivP2Pattern2. Probably worth fixing though I don't care overly much.

> tests/end_to_end/ido-division-large-addend/compiled-shift16.s and tests/end_to_end/ido-division-large-addend/compiled-shift17.s

Those are fair. I think they're low priority (how often do you divide by such a large power of two, rather than bit shift?), but I'm okay taking them if you feel like adding code for them. In that case I'd prefer a static pattern like the existing ones, though (I guess it becomes an additional 4 patterns, for the combinations of 2^16 vs 2^17+, and input reg = output reg vs input reg != output reg; could reduce that a little by adding a pattern for first folding `lui $at, N`, `addiu $at, $at, M` -> `li $at, (N << 16) + M`).

>  It correctly determines the div pattern in these but the immediate and arithmetic is flipped (not exactly wrong output, just not as clear)

Ugh, yeah, I've been thinking of how to best deal with cases like `arg0 + 0xFFC00000` vs `arg0 - 0x400000`. Sometimes you do want the former, and it's not clear to me where in the code this kind of folding would go. ARM suffers even more from it, since it does 32-bit loads from literal pools even for somewhat smaller constants.

Drahsid:
> Those are fair. I think they're low priority (how often do you divide by such a large power of two, rather than bit shift?)

Yeah I think this is pretty rare, but the patterns are at least pretty straightforward.

> Ugh, yeah, I've been thinking of how to best deal with cases like `arg0 + 0xFFC00000` vs `arg0 - 0x400000`. Sometimes you do want the former, and it's not clear to me where in the code this kind of folding would go.

Thinking on this more, I wonder if we can make a guess based on the assumed signedness of the non-literal(s) in the operation. In this case, if arg0 is unsigned, the former is alright, otherwise, the other option might seem more correct. I feel like this might still be a bit complex, do you have any better ideas?


For `tests/end_to_end/ido-division-large-addend/clobbered-bias-read.s` I thought that there might be a way to produce this after seeing the bug, but it seems pretty unlikely (and if it does exist, it's only on ido 5.3).

matt-kempster:
I can turn off the Codex code review action (I didn't intentionally turn it on, I guess it turned itself on when I connected my GitHub account to Codex at one point). But seeing as it found a legitimate bug in this PR, I am tempted to keep it on for now. Please let me know if you have a strong opinion to the contrary.

AngheloAlf:
fyi I released a new rabbitizer version, it fixes the issue about wrongly symbolizing those constants.

image

Drahsid:
Since the main issue here was fixed by the rabbitizer update, I'm going to close this PR and maybe open some new ones later regarding the other issues.

PR #315, opened 2026-06-16
client-side browser implementation

mkst:
e.g. https://mkst.github.io/m2c/

Tried to be as non-invasive as possible. Obviously this could replace the server-side CGI version (and then can do some cleanup e.g. website.py)...


m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+    source: str, context: Optional[str] = None, flags: Optional[List[str]] = None
+) -> BrowserResult:
+    """Run m2c from string inputs, for Pyodide/browser use."""
+    if not _FUNCTION_START_RE.search(source):
+        source = "glabel foo\n" + source
+
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+    is_visualize = False
+    was_gc_enabled = gc.isenabled()
```
simonlindholm: nothing ever disables gc, so no need for this

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+    gc_thresholds = gc.get_threshold()
+
+    try:
+        with tempfile.TemporaryDirectory(prefix="m2c-browser-") as tmpdir:
+            base_path = Path(tmpdir)
+            asm_path = base_path / "input.s"
+            asm_path.write_text(source, encoding="utf-8")
+
+            argv = list(flags or [])
+            argv.append("--no-cache")
```
simonlindholm:
```suggestion
            argv.append("--no-cache")
            argv.append("--visualize-format=dot")
```

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+
+            argv.append(str(asm_path))
+
+            with contextlib.redirect_stdout(stdout), contextlib.redirect_stderr(stderr):
+                options = parse_flags(argv)
+                is_visualize = options.visualize_flowgraph is not None
+                if is_visualize:
+                    options.visualize_format = (
+                        main_module.Options.VisualizeFormatEnum.DOT
+                    )
```
simonlindholm:
```suggestion
```

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+        outputType = result.get("output_type");
+      } finally {
+        if (result) {
+          result.destroy();
+        }
+      }
+      document.body.dataset.m2cReturncode = String(returncode);
+      document.body.dataset.m2cOutput = output;
+
+      if (returncode === 0 && outputType === "dot") {
```
simonlindholm: `returncode === 0 && action === "visualize"` to get rid of the need for outputType

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+      showTextOutput(formatError(err));
+    } finally {
+      resetButtonLabels();
+      buttonEl.disabled = false;
+      visualizeEl.disabled = false;
+    }
+  }
+
+  restoreState();
+  if (!localStorage.mips_to_c_saved_options || localStorage.mips_to_c_saved_options.indexOf("\"dark\"") === -1) {
```
simonlindholm: why search for a string here instead of using the parsed json?

browser/index.html
```diff
@@ -0,0 +1,100 @@
...
+  <div>
+    Output:
+  </div>
+  <div class="output-pane">
+    <textarea id="output" name="outputframe" spellcheck="false" readonly></textarea>
+    <div id="output-graph"></div>
+  </div>
+</div>
+<script type="module">
+import { instance } from "https://cdn.jsdelivr.net/npm/@viz-js/viz@3.28.0/+esm";
```
simonlindholm: can we make the build pull down these third-party dependencies and verify their hashes?

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+        if (result) {
+          result.destroy();
+        }
+      }
+      document.body.dataset.m2cReturncode = String(returncode);
+      document.body.dataset.m2cOutput = output;
+
+      if (returncode === 0 && outputType === "dot") {
+        if (!window.m2cVizReady) {
+          throw new Error("Viz.js was not loaded");
```
simonlindholm: I think there's a race condition here, where the module script runs asynchronously, so this can legitimately happen. Could we have a promise for the entire viz.js load, that rejects if it fails to load?

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+        returncode = result.get("returncode");
+        output = result.get("output");
+        outputType = result.get("output_type");
+      } finally {
+        if (result) {
+          result.destroy();
+        }
+      }
+      document.body.dataset.m2cReturncode = String(returncode);
+      document.body.dataset.m2cOutput = output;
```
simonlindholm: why this?

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+        source: sourceEl.value,
+        context: contextEl.value,
+        flags: flags
+      }));
+      var result = null;
+      var returncode;
+      var output;
+      var outputType;
+      try {
+        result = await pyodide.runPythonAsync("decompile_from_json(m2c_options_json).copy()");
```
simonlindholm: if decompile_from_json is made to return a plain dict, pyodide will convert that into a JS object which reduces some complexity here (e.g. avoids the need for `destroy()`)

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+      }
+    } catch (err) {
+      console.error(err);
+      var message = formatError(err);
+      setBusyButton("decompile", "Failed");
+      showTextOutput(message);
+    }
+  }
+
+  async function runM2c(action) {
```
simonlindholm: make `action` a bool `visualize` I guess

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+    r"[A-Za-z_.$][\w.$]*:"
+    r")",
+    re.MULTILINE,
+)
+_RECURSION_LIMIT_SET = False
+
+
+def _ensure_recursion_limit() -> None:
+    global _RECURSION_LIMIT_SET
+    if not _RECURSION_LIMIT_SET:
```
simonlindholm: instead of a lazy global for this, make the JS side do it on startup

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+    output: str
+    output_type: str
+
+
+_FUNCTION_START_RE = re.compile(
+    r"^\s*(?:"
+    r"(?:glabel|dlabel|arm_func_start|thumb_func_start|"
+    r"non_word_aligned_thumb_func_start|ARM_FUNC_START|THUMB_FUNC_START|"
+    r"NON_WORD_ALIGNED_THUMB_FUNC_START)\s+\S+|"
+    r"\.fn\s+\S+|"
```
simonlindholm: what happened to `.global`?

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+    r"[A-Za-z_.$][\w.$]*:"
```
simonlindholm: This will match even local labels, which means the auto-glabel insertion will almost never trigger

m2c/browser.py
simonlindholm, line 1: This file doesn't make sense without the surrounding JS; maybe we can embed it as a string in the JS file instead?

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+        sys.setrecursionlimit(min(2**31 - 1, 10 * sys.getrecursionlimit()))
+        _RECURSION_LIMIT_SET = True
+
+
+def decompile(
+    source: str, context: Optional[str] = None, flags: Optional[List[str]] = None
+) -> BrowserResult:
+    """Run m2c from string inputs, for Pyodide/browser use."""
+    if not _FUNCTION_START_RE.search(source):
+        source = "glabel foo\n" + source
```
simonlindholm: This logic can be moved to JS I guess, especially if it makes it easier to deduplicate the list in _FUNCTION_START_RE with the function select dropdown logic

m2c/browser.py
```diff
@@ -0,0 +1,128 @@
...
+
+def _ensure_recursion_limit() -> None:
+    global _RECURSION_LIMIT_SET
+    if not _RECURSION_LIMIT_SET:
+        sys.setrecursionlimit(min(2**31 - 1, 10 * sys.getrecursionlimit()))
+        _RECURSION_LIMIT_SET = True
+
+
+def decompile(
+    source: str, context: Optional[str] = None, flags: Optional[List[str]] = None
```
simonlindholm: no need for defaults for these, and also no reason to make them accept None

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+    if (action === "visualize") {
+      visualizeEl.value = message;
+    } else {
+      buttonEl.value = message;
+    }
+  }
+
+  function formatError(err) {
+    if (err && err.stack) {
+      return err.stack;
```
simonlindholm: I saw this once and got a bit confused, the error message ought to be included and it ought to say "error" somewhere

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+    functionEl.innerHTML = "";
+
+    var allOption = document.createElement("option");
+    allOption.value = "all";
+    allOption.textContent = "all functions";
+    functionEl.appendChild(allOption);
+
+    var matches = sourceEl.value.matchAll(/^\s*(?:(?:glabel|dlabel|arm_func_start|thumb_func_start|non_word_aligned_thumb_func_start|ARM_FUNC_START|THUMB_FUNC_START|NON_WORD_ALIGNED_THUMB_FUNC_START)\s+(\S+)|\.fn\s+(\S+)|([A-Za-z_.$][\w.$]*):)/gm);
+    for (var match of matches) {
+      var name = match[1] || match[2] || match[3];
```
simonlindholm: dlabel is not a function label.

Why separate .fn from the rest of the regex?

same thing about `label:` as in the other comment. It could make some sense to support the `label:` syntax for defining functions, since m2c does allow it, but then we should match m2c's logic for when a label is treated as a function label, as set down in asm_file.py ("doesn't start with `.` and doesn't match `re_local_label` ").

browser/app.js
```diff
@@ -0,0 +1,371 @@
...
+    graphEl.replaceChildren(svgElement);
+  }
+
+  function normalizeDotForBrowser(dotSource) {
+    return dotSource
+      .replace(
+        'node [shape="rect", fontname="Monospace"];',
+        'node [shape="rect", fontname="Courier", margin="0.12,0.08"];'
+      )
+      .replace('edge [fontname="Monospace"];', 'edge [fontname="Courier"];');
```
simonlindholm: `.replace` only replaces the first hit, unless you use a regex with `/g`.

What's the impact of this replacement, visually?

simonlindholm (commented):
Visualization does not work for me at https://mkst.github.io/m2c/, I get the following error:
```
renderString@https://cdn.jsdelivr.net/npm/@viz-js/viz@3.28.0/+esm:15:1238570
renderSVGElement@https://cdn.jsdelivr.net/npm/@viz-js/viz@3.28.0/+esm:15:1238696
runM2c@https://mkst.github.io/m2c/app.js:332:29
```

mkst:
Will return with updates. The tldr is that I trusted too much in codex despite telling it to simplify.

Not sure why the visualization fails, it works for me on mobile & laptop 🤔

simonlindholm:
The "visualize crash" seems input dependent, I get it when using just "nop" as input. I guess it might be including `Warning: missing "jr $ra" in last block of foo (initial).` or similar in the output?

Some UI notes:
- the output survives across a reload. Feels a bit weird, and my spontaneous reaction is to clear the output, but the text area is uneditable.
- maybe ought to focus the output on decompile, so you can ctrl+a, ctrl+c?
- when decompilation is fast (as is usually the case), it's distracting that the Decompile button momentarily changes size and shifts the rest of the UI around it. Not sure what to suggest as alternative UI though...

browser/index.html
```diff
@@ -0,0 +1,106 @@
...
+    <div id="output-graph"></div>
+  </div>
+</div>
+<script>
+window.m2cVizReady = new Promise(function (resolve, reject) {
+  window.m2cResolveViz = resolve;
+  window.m2cRejectViz = reject;
+});
+</script>
+<script type="module">
```
simonlindholm, line 92:
```suggestion
<script type="module" onerror="window.m2cRejectViz()">
```

Alternatively, untested, but might work to do:
```js
window.m2cVizReady = (async function() {
  let { instance } = await import("./dist/vendor/viz/3.28.0/viz.mjs");
  return await instance();
})()
```

simonlindholm:
(lmk if you feel done and want another round of review; so far it felt like most of the comments still apply)

browser/app.js
mkst: Without the replace, the boxes are too small for the text, but perhaps there's a smarter solve 
image

simonlindholm: Worth a comment in the function, probably. Are both the Monospace -> Courier and margin changes necessary?

mkst:
Had another go at it :)

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+
+class BrowserResult(TypedDict):
+    returncode: int
+    output: str
+
+
+def decompile(source: str, context: str, flags: List[str]) -> BrowserResult:
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+    is_visualize = False
```
simonlindholm:
```suggestion
```

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    output = stdout.getvalue()
+    if is_visualize and returncode == 0:
+        dot_start = output.find("digraph {")
+        if dot_start != -1:
+            output = output[dot_start:]
+
+    return {
+        "returncode": returncode,
+        "output": output,
+    }
```
simonlindholm: Could we move this function into main.py and make it return a dataclass instead? (so we get type checking back; I forgot about when making the suggestion to move it into JS)

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+
+
+def decompile_from_json(options_json: str) -> BrowserResult:
```
simonlindholm: and then we keep this function in JS and make it responsible solely for marshalling data to/from JS-friendly formats:
```py
def decompile_from_json(options_json: str) -> str:
    try:
        options = json.loads(options_json)
        assert isinstance(options, dict)
        source = options["source"]
        context = options["context"]
        flags = options["flags"]
        assert isinstance(source, str)
        assert isinstance(context, str)
        assert isinstance(flags, list)
        assert all(isinstance(flag, str) for flag in flags)
        res = decompile_for_browser(source, context, flags)
    except Exception as e:
        res = Blah(returncode=1, output=f"Internal error: {e}")
    return json.dumps({"returncode": res.returncode, "output": res.output})
```

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    "glabel",
+    "arm_func_start",
+    "thumb_func_start",
+    "non_word_aligned_thumb_func_start",
+    "ARM_FUNC_START",
+    "THUMB_FUNC_START",
+    "NON_WORD_ALIGNED_THUMB_FUNC_START",
+    ".fn"
+  ];
+  var localLabelRe = /^(?:loc_|locret_|def_|lbl_|LAB_|switchD_|jump_|_[0-9A-Fa-f]{7,8}(?:_.*)?$)/;
```
simonlindholm: Add a comment next to `re_local_label` in asm_file.py to keep it in sync with this copy, and rename it `reLocalLabel` to match

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    } else if (err && typeof err === "object") {
+      try {
+        message = JSON.stringify(err);
+      } catch (jsonErr) {
+        message = String(err);
+      }
+    } else {
+      message = String(err);
+    }
+    return message.indexOf("Error") === -1 && message.indexOf("error") === -1 ? "Error: " + message : message;
```
simonlindholm: Even if the error comes with a stack trace its message is still important. Also I don't see the harm in prepending "Error:" always.

```suggestion
    var message = "Error: ";
    if (err && err.message) {
        message += err.message;
    } else if (err && typeof err === "object") {
      try {
        message = JSON.stringify(err);
      } catch (jsonErr) {
        message = String(err);
      }
    } else {
      message = String(err);
    }
    if (err && err.stack) {
        message += "\n\n" + err.stack;
    }
    return message;
```

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    if (!labelMatch) {
+      return "";
+    }
+    var label = labelMatch[1].replace(/^"|"$/g, "");
+    if (label.charAt(0) === "." || localLabelRe.test(label)) {
+      return "";
+    }
+    return label;
+  }
+
```
simonlindholm, line 112: To avoid line-splitting code duplication, add a function `getFunctionNames(source)` and replace `hasFunctionStart(x)` by `getFunctionNames(x).length > 0`, and make `updateFunctions` use it too

I guess `parseFunctionName` can be inlined into the new `parseFunctionNames` function

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+      } catch (jsonErr) {
+        message = String(err);
+      }
+    } else {
+      message = String(err);
+    }
+    return message.indexOf("Error") === -1 && message.indexOf("error") === -1 ? "Error: " + message : message;
+  }
+
+  function stripLineComment(line) {
```
simonlindholm: inline this function into its user

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    return line.replace(/[#;@].*$/, "").replace(/\/\/.*$/, "").trim();
+  }
+
+  function parseFunctionName(line) {
+    var stripped = stripLineComment(line);
+    var parts = stripped.split(/\s+/);
+    var directive = parts[0];
+    if (functionStartDirectives.indexOf(directive) !== -1 && parts[1]) {
+      return parts[1].replace(/,$/, "");
```
simonlindholm: It bugs me a little that this handles `glabel foo, local` but not `glabel foo,local` -- whitespace is just the wrong thing to split on for the arguments to the directive.

(We also don't handle quoted arguments, and `#;@` in strings terminate the parsed line. That's less important though.)

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    var labelMatch = stripped.match(/^([a-zA-Z0-9_.$]+|"[a-zA-Z0-9_.$<>@,-]+"):/);
```
simonlindholm: make this label a global and put it next to `localLabelRe`, it also needs to be covered by the code duplication comment

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    if (!labelMatch) {
+      return "";
+    }
+    var label = labelMatch[1].replace(/^"|"$/g, "");
```
simonlindholm: eh, just use the same regex as Python, making this `labelMatch[1] || labelMatch[2]`

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    var savedContext = localStorage.m2c_saved_context;
+    var savedOptions = localStorage.m2c_saved_options;
+
+    if (savedSource) sourceEl.value = savedSource;
+    if (savedContext) contextEl.value = savedContext;
+
+    if (!savedOptions) {
+      return {};
+    }
+    try {
```
simonlindholm, line 190:
```js
if (savedOptions) {
    try {
        ...
        return options;
    } catch (err) {
        console.warn(...);
    }
}
return {};
```
to reduce code duplication

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    });
+  }
+
+  function sourceWithDefaultFunction() {
+    return hasFunctionStart(sourceEl.value) ? sourceEl.value : "glabel foo\n" + sourceEl.value;
+  }
+
+  function updateFunctions() {
+    var previous = functionEl.value;
+    functionEl.innerHTML = "";
```
simonlindholm:
```suggestion
    functionEl.replaceChildren();
```
to match code elsewhere (alternatively: `x.textContent = "";` everywhere)

browser/index.html
```diff
@@ -0,0 +1,102 @@
...
+<link rel="stylesheet" href="style.css">
+</head>
+<body>
+<form>
+<div class="main">
+  <div>
+    Assembly:
+  </div>
+  <div style="flex: 20;">
+    <textarea id="source" name="source" spellcheck="false"></textarea>
```
simonlindholm: there is no longer a need for `name=` for any of these

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    if (savedSource) sourceEl.value = savedSource;
+    if (savedContext) contextEl.value = savedContext;
+
+    if (!savedOptions) {
+      return {};
+    }
+    try {
+      var options = JSON.parse(savedOptions);
+      for (var key in options) {
+        var el = document.getElementsByName(key)[0];
```
simonlindholm: `getElementById` to match `saveState` and to get rid of now-pointless `name` attributes

(note that `comment_style` differed between name and id and thus failed to restore)

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+  }
+
+  function buildFlags() {
+    var flags = [];
+    var globals = document.getElementById("globals").value;
+    var target = document.getElementById("target").value;
+    var commentStyle = document.getElementById("comment-style").value;
+    var regvarsSelect = regvarsSelectEl.value;
+
+    flags.push("--globals", globals);
```
simonlindholm: might as well do
```suggestion
    flags.push("--globals=" + globals);
```
to avoid command injection, not that it really matters now that it's local

same elsewhere

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    if (document.getElementById("extraswitchindent").checked) flags.push("--indent-switch-contents");
+    if (document.getElementById("leftptr").checked) flags.push("--pointer-style", "left");
+    if (document.getElementById("zfillconstants").checked) flags.push("--zfill-constants");
+    if (document.getElementById("noifs").checked) flags.push("--gotos-only");
+    if (document.getElementById("noswitches").checked) flags.push("--no-switches");
+    if (document.getElementById("nounkinference").checked) flags.push("--no-unk-inference");
+    if (document.getElementById("stackstructs").checked) flags.push("--stack-structs");
+    if (document.getElementById("nostackspill").checked) flags.push("--no-stack-spill");
+    if (document.getElementById("descendingregs").checked) flags.push("--descending-regs");
+    if (document.getElementById("backwardsbss").checked) flags.push("--backwards-bss");
```
simonlindholm: loop over
```js
var boolFlags = {
    "void": "--void",
    "leftptr": "--pointer-style=left",
    ...
}
```
to make this more declarative maybe

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+          if (!pyodide.FS.analyzePath(current).exists) {
+            throw err;
+          }
+        }
+      }
+    }
+
+    mkdirp(appRoot + "/m2c");
+    mkdirp(appRoot + "/m2c_pycparser");
+    mkdirp(appRoot + "/m2c_pycparser/ply");
```
simonlindholm: feels like this isn't needed given the mkdirp below. Alternatively, these `mkdirp`s can be `mkdir` and the mkdirp below can be removed, to remove the need for the `mkdirp` function

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+    if (document.getElementById("backwardsbss").checked) flags.push("--backwards-bss");
+
+    return flags;
+  }
+
+  function writeBrowserFiles(files) {
+    function mkdirp(path) {
+      var parts = path.split("/");
+      var current = "";
+      for (var i = 0; i < parts.length; i += 1) {
```
simonlindholm: for (let part of parts)

(we should use `let` throughout really, my only reason for using `var` is because it's what I grew up with)

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+      }
+
+      setBusyButton("decompile", "Loading...");
+      await loadScript(window.M2C_VENDOR_PATHS.pyodideScript);
+      pyodide = await loadPyodide({
+        indexURL: window.M2C_VENDOR_PATHS.pyodideIndexURL
+      });
+      setBusyButton("decompile", "Installing...");
+      writeBrowserFiles(window.M2C_PYTHON_FILES);
+      await pyodide.runPythonAsync("import json\nimport sys\nsys.path.insert(0, '/tmp/m2c-browser-root')\nsys.setrecursionlimit(min(2**31 - 1, 10 * sys.getrecursionlimit()))\nfrom m2c.browser import decompile_from_json\n");
```
simonlindholm, line 352: `-style multiline string

actually, maybe all of this can be put into `browserPython` and that variable can be evaled here, to avoid the need for writing `m2c/browser.py`

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+      showTextOutput(message);
+    }
+  }
+
+  async function runM2c(visualize) {
+    if (!ready) {
+      return;
+    }
+
+    saveState();
```
simonlindholm: shouldn't be necessary

browser/app.js
```diff
@@ -0,0 +1,518 @@
...
+  function parseFunctionName(line) {
+    var stripped = stripLineComment(line);
+    var parts = stripped.split(/\s+/);
+    var directive = parts[0];
+    if (functionStartDirectives.indexOf(directive) !== -1 && parts[1]) {
+      return parts[1].replace(/,$/, "");
+    }
+
+    var labelMatch = stripped.match(/^([a-zA-Z0-9_.$]+|"[a-zA-Z0-9_.$<>@,-]+"):/);
+    if (!labelMatch) {
```
simonlindholm: the python code has this as a while rather than an if to handle things on the same line as local labels; won't matter but might as well match it (once parseFunctionName gets inlined into parseFunctionNames, and moving the local label parse code above the one for global labels)

browser/app.js
simonlindholm: An alternative to fixing up all this logic would be to call into Python code, where it would be easier to reuse code. Though that adds complications around Pyodide needing to be loaded.

browser/build_bundle.py
```diff
@@ -0,0 +1,164 @@
...
+        output_path.parent.mkdir(parents=True, exist_ok=True)
+        output_path.write_bytes(data)
+
+    write_vendor_lock(lock)
+
+
+def update_vendor_files() -> None:
+    lock = read_vendor_lock()
+
+    for relative_path, metadata in sorted(lock["files"].items()):
```
simonlindholm, line 104: This shouldn't pass mypy; `object` does not have `.items()`. Can you update mypy.ini to make sure this file gets type-checked, and fix the typing here somehow? E.g. by defining a `LockJson` type as a TypedDict around a dict around another TypedDict, and doing `typing.cast` in `read_vendor_lock` out of laziness, or something more proper if you want.

browser/build_bundle.py
```diff
@@ -0,0 +1,164 @@
...
+        expected_hash = metadata["sha256"]
+
+        if output_path.exists():
+            data = output_path.read_bytes()
+            actual_hash = sha256(data)
+            if actual_hash == expected_hash:
+                continue
+            print(f"Hash mismatch for {output_path}, re-downloading...")
+
+        print(f"Downloading {metadata['url']}...")
```
simonlindholm, line 115: variable for `metadata["url"]`

browser/build_bundle.py
```diff
@@ -0,0 +1,164 @@
...
+VENDOR_PATHS_JS = DIST_DIR / "vendor-paths.js"
+
+INCLUDE_DIRS = (
+    ROOT / "m2c",
+    ROOT / "m2c_pycparser",
+    ROOT / "m2c_pycparser" / "ply",
+)
+
+
+def iter_python_files() -> list[Path]:
```
simonlindholm, line 23: `iter_` in the name is misleading when it returns a list

m2c/main.py
```diff
@@ -40,6 +43,43 @@ class Valid:
...
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+
+    with tempfile.TemporaryDirectory(prefix="m2c-browser-") as tmpdir:
+        base_path = Path(tmpdir)
+        asm_path = base_path / "input.s"
+        asm_path.write_text(source, encoding="utf-8")
+
+        argv = flags + ["--no-cache", "--visualize-format=dot"]
+        if context:
```
simonlindholm, line 62: No point in making this conditional

browser/build_bundle.py
```diff
@@ -0,0 +1,164 @@
...
+    return parser.parse_args()
+
+
+def main() -> None:
+    args = parse_args()
+    if args.update_vendor_lock:
+        update_vendor_lock()
+
+    update_vendor_files()
+    lock = read_vendor_lock()
```
simonlindholm, line 144: I think it's kinda silly that `read_vendor_lock()` is called from several functions, better just read the lock file once, right after `parse_args()`, and pass it around to `update_vendor_lock()` and `update_vendor_files()`

browser/app.js
```diff
@@ -0,0 +1,437 @@
+(function () {
+  let pyodide = null;
+  let ready = false;
+  let appRoot = "/tmp/m2c-browser-root";
```
simonlindholm, line 4: `let` -> `const` for most of these. Within small functions where you can see that a variable doesn't change I feel like it doesn't matter, but within a huge one like here it's nice.

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+    if (err && err.stack) {
+      message += "\n\n" + err.stack;
+    }
+    return message;
+  }
+
+  function getFunctionNames(source) {
+    let names = [];
+    for (let line of source.split(/\r?\n/)) {
+      let stripped = line.replace(/[#;@].*$/, "").replace(/\/\/.*$/, "").trim();
```
simonlindholm, line 116: m2c does not support `//`-style comments elsewhere, though it does support `/* ... */`. Would be nice to be consistent. Other things we could fix at the same time:
- `!` indicates a comment for sh2
- quoted symbol names are allowed to contain comment syntax (can be done using the `re_comment_or_string` trick)

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+    }
+    if (err && err.stack) {
+      message += "\n\n" + err.stack;
+    }
+    return message;
+  }
+
+  function getFunctionNames(source) {
+    let names = [];
+    for (let line of source.split(/\r?\n/)) {
```
simonlindholm, line 115:
```suggestion
    for (let line of source.split("\n")) {
```
since CR's cannot occur in browser textareas, and also would be immediately stripped anyway

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+      let labelMatch;
+      while ((labelMatch = stripped.match(reLabel)) !== null) {
+        let label = labelMatch[1] || labelMatch[2];
+        stripped = stripped.slice(labelMatch[0].length).trim();
+        if (label.charAt(0) !== "." && !reLocalLabel.test(label)) {
+          names.push(label);
+        }
+      }
+      let directiveMatch = stripped.match(/^(\S+)\s+([^,\s]+)/);
+      if (directiveMatch && functionStartDirectives.indexOf(directiveMatch[1]) !== -1) {
```
simonlindholm, line 126:
```suggestion
      if (directiveMatch && functionStartDirectives.includes(directiveMatch[1])) {
```

browser/app.js
simonlindholm: still relevant

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+        names.push(directiveMatch[2]);
+      }
+    }
+    return names;
+  }
+
+  function sourceWithDefaultFunction() {
```
simonlindholm, line 133: Inline this function into its caller, to avoid a function that reads a global variable without clear context for it. Or pass sourceEl.value as an argument to it, making it pure

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+    });
+  }
+
+  async function initPyodide() {
+    try {
+      if (!window.M2C_VENDOR_PATHS) {
+        throw new Error("vendor-paths.js was not loaded");
+      }
+      if (!window.M2C_PYTHON_FILES) {
+        throw new Error("m2c.generated.js was not loaded");
```
simonlindholm, line 342: It's not called that any more.

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+    let options = {};
+
+    for (let id of optionIds) {
+      let el = document.getElementById(id);
+      options[id] = el.type === "checkbox" ? (el.checked ? "yes" : "no") : el.value;
+    }
+
+    localStorage.m2c_saved_source = sourceEl.value;
+    localStorage.m2c_saved_context = contextEl.value;
+    localStorage.m2c_saved_options = JSON.stringify(options);
```
simonlindholm, line 176: Previously, saving source/context/options was split into three different functions (or, well, inlined at their respective uses). I think that might possibly be a good thing to keep given how contexts can be gigantic and localStorage writes can lag a bit

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+    let target = document.getElementById("target").value;
+    let commentStyle = document.getElementById("comment-style").value;
+    let regvarsSelect = regvarsSelectEl.value;
+
+    flags.push("--globals=" + globals);
+    flags.push("--target=" + target);
+
+    if (commentStyle === "none") {
+      flags.push("--comment-style=none");
+    } else if (commentStyle.indexOf("oneline") === 0) {
```
simonlindholm, line 257:
```suggestion
    } else if (commentStyle.startsWith("oneline")) {
```

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+      flags.push("--comment-style=oneline");
+    } else {
+      flags.push("--comment-style=multiline");
+    }
+
+    if (commentStyle.indexOf("unaligned") !== -1) {
```
simonlindholm, line 263:
```suggestion
    if (commentStyle.includes("unaligned")) {
```

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+        if (pyodide.FS.analyzePath(current).exists) {
+          continue;
+        }
+        try {
+          pyodide.FS.mkdir(current);
+        } catch (err) {
+          if (!pyodide.FS.analyzePath(current).exists) {
+            throw err;
+          }
+        }
```
simonlindholm, line 312:
```suggestion
        if (!pyodide.FS.analyzePath(current).exists) {
          pyodide.FS.mkdir(current);
        }
```
since this is a memfs

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+
+  function writeBrowserFiles(files) {
+    function mkdirp(path) {
+      let parts = path.split("/");
+      let current = "";
+      for (let i = 0; i < parts.length; i += 1) {
+        if (!parts[i]) {
+          continue;
+        }
+        current += "/" + parts[i];
```
simonlindholm, line 302:
```suggestion
      for (let part of parts) {
        if (!part) {
          continue;
        }
        current += "/" + part;
```

browser/app.js
```diff
@@ -0,0 +1,437 @@
...
+`;
+  let functionStartDirectives = [
+    "glabel",
+    "arm_func_start",
+    "thumb_func_start",
+    "non_word_aligned_thumb_func_start",
+    "ARM_FUNC_START",
+    "THUMB_FUNC_START",
+    "NON_WORD_ALIGNED_THUMB_FUNC_START",
+    ".fn"
```
simonlindholm, line 49: This also needs to be kept in sync with Python, can you add a comment on both sides?

simonlindholm (commented):
Looks pretty good

mkst:
I'm going get back to this soon - but I will do some kind of refactor with the intention of being able to share the .js portion with decomp.me (`/decompile` is our heaviest endpoint, so want to see whether we can run that clientside).

m2c/asm_file.py
```diff
@@ -824,6 +826,7 @@ def process_label(label: str, *, kind: LabelKind) -> None:
                     process_label(args[0], kind=LabelKind.JUMP_TARGET)
 
             elif directive in (
+                # Keep in sync with functionStartDirectives in browser/app.js.
```
simonlindholm, line 829: Could you move this list out into a variable? (It could contain `.fn` too.) Feels weird to have comments in the middle of an expression like this, and it would allow the JS comment to have a name to refer to.

simonlindholm (approved):

m2c/asm_file.py
simonlindholm: Or actually, I think this is probably fine as is. I'll just merge!

PR #316, opened 2026-07-14
SuperH scaffolding for empty functions

sozud:
This PR adds basic scaffolding to allow decompiling an empty function for the SuperH architecture. The target is specifically SH-2 and GCC. The goal is to enhance the support from here in small steps so that it's easily reviewable. This PR adds a test that can be run as follows:
```
python3 m2c.py --target sh2-gcc-c -f test tests/end_to_end/sh-empty/sh-empty.s
```
Expected output is:
```
void test(void) {

}
```


m2c/options.py
```diff
@@ -57,13 +58,16 @@ def arch(self) -> Target.ArchEnum:
                 return Target.ArchEnum.ARM
             elif self == Target.PlatformEnum.GBA:
                 return Target.ArchEnum.ARM
+            elif self == Target.PlatformEnum.SH2:
+                return Target.ArchEnum.SH
```
simonlindholm: sh vs sh2, should we be consistent? (I don't have a good feeling for how they relate)

m2c/arch_sh.py
```diff
@@ -0,0 +1,142 @@
...
+
+    stack_pointer_reg = Register("r15")
+    frame_pointer_regs = [Register("r14")]
+    return_address_reg = Register("pr")
+
+    base_return_regs = [(Register("r0"), False), (Register("r1"), False)]
+    all_return_regs = [Register("r0"), Register("r1")]
+
+    argument_regs = [Register(r) for r in ["r4", "r5", "r6", "r7"]]
+    simple_temp_regs = [Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13"]]
```
simonlindholm: r8 is surely not both a temp reg and a saved reg

m2c/arch_sh.py
```diff
@@ -0,0 +1,142 @@
...
+    re_comment = r"!.*"
+    supports_dollar_regs = False
+
+    home_space_size = 0
+
+    stack_pointer_reg = Register("r15")
+    frame_pointer_regs = [Register("r14")]
+    return_address_reg = Register("pr")
+
+    base_return_regs = [(Register("r0"), False), (Register("r1"), False)]
```
simonlindholm: Is there a document describing the ABI you're targetting? All I can find is https://users.informatik.haw-hamburg.de/~krabat/FH-Labor/gnupro/7_GNUPro_Embedded_Development/embCalling_conventions_for_00000036.html, which doesn't mention anything about r1.

m2c/arch_sh.py
```diff
@@ -0,0 +1,142 @@
...
+            clobbers=clobbers,
+            outputs=outputs,
+            eval_fn=eval_fn,
+            is_return=is_return,
+            has_delay_slot=has_delay_slot,
+        )
+
+    def arg_name(self, loc: ArgLoc) -> str:
+        if loc.offset is not None:
+            return f"arg{loc.offset // 4}"
```
```suggestion
            return f"arg{loc.offset // 4 + 4}"
```
I assume, given home_space = 0

simonlindholm (commented):
Exciting!

m2c/arch_sh.py
```diff
@@ -0,0 +1,142 @@
...
+    return_address_reg = Register("pr")
+
+    base_return_regs = [(Register("r0"), False), (Register("r1"), False)]
+    all_return_regs = [Register("r0"), Register("r1")]
+
+    argument_regs = [Register(r) for r in ["r4", "r5", "r6", "r7"]]
+    simple_temp_regs = [Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13"]]
+    temp_regs = argument_regs + simple_temp_regs
+    saved_regs = [
+        Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15"]
```
simonlindholm: "pr" too I think

m2c/options.py
sozud: I don't have a ton of background but the most likely other platform here would be Dreamcast which has a SH4. I think most of the difference is SH4 has an FPU and MMU but is otherwise a superset. I'll change it to SH2 since I don't have plans to add SH4 support

m2c/arch_sh.py
sozud: I'm basing this off of https://llvm-gcc-renesas.com/manuals/SH-ABI-Specification.html
and this bit here:
```
The registers and there usage is given below:
R0-R1:         Return value, caller saves
```

simonlindholm: Aha:
>  The scalar function value no larger than 4 bytes in size is returned in R0 and FR0 and values no larger than 8 bytes are returned in R0 and R1.
>
> If R0 and R1 are used to return the function value, R0 contains the most significant part and R1 contains the least significant part for big-endian mode and vice-versa for little-endian mode.

(Is the platform you are interested in big or little endian?)

I think we ought to match MIPS here (which is in a similar spot) and set base_return_regs = [r0], all_return_regs = [r0, r1].

m2c/arch_sh.py
sozud: Ok looks like this now, I think this is more correct
```
    argument_regs = [Register(r) for r in ["r4", "r5", "r6", "r7"]]
    simple_temp_regs = [Register(r) for r in ["r0", "r1", "r2", "r3"]]
    temp_regs = argument_regs + simple_temp_regs

    saved_regs = [
        Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15", "pr"]
    ]
```

m2c/arch_sh.py
sozud: Saturn is big-endian, I made the base_return_reg / sall_return_regs adjustment like you suggested

m2c/arch_sh.py
sozud: I'm not totally sure what this does is but I applied your suggestion. My understanding is that the superh ABI doesn't define a home space.

m2c/arch_sh.py
sozud: Ok added

m2c/arch_sh.py
simonlindholm: It means that the argument at intial_sp + 4*i gets the name "arg{i+4}", arg0..arg3 being taken by register args

m2c/arch_sh.py
```diff
@@ -0,0 +1,143 @@
...
+
+    base_return_regs = [(Register("r0"), False)]
+    all_return_regs = [Register("r0"), Register("r1")]
+
+    argument_regs = [Register(r) for r in ["r4", "r5", "r6", "r7"]]
+    simple_temp_regs = [Register(r) for r in ["r0", "r1", "r2", "r3"]]
+    temp_regs = argument_regs + simple_temp_regs
+
+    saved_regs = [
+        Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15", "pr"]
```
```suggestion
        Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "pr"]
```

simonlindholm (approved):

sozud:
I reformatted but looks like it needs approval to run

m2c/arch_sh.py
```diff
@@ -0,0 +1,144 @@
...
+    base_return_regs = [(Register("r0"), False)]
+    all_return_regs = [Register("r0"), Register("r1")]
+
+    argument_regs = [Register(r) for r in ["r4", "r5", "r6", "r7"]]
+    simple_temp_regs = [Register(r) for r in ["r0", "r1", "r2", "r3"]]
+    temp_regs = argument_regs + simple_temp_regs
+
+    saved_regs = [
+        Register(r)
+        for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15", "pr"]
```
```suggestion
        Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "pr"]
```

PR #317, opened 2026-07-14
sh2: implement preamble and return value

sozud:
This implements mov and mov.l instructions to support gcc function boilerplate and a simple return value like:
```
/* 0x0600DE2C 0x2FE6 */ mov.l r14, @-r15
/* 0x0600DE2E 0x6EF3 */ mov r15, r14
/* 0x0600DE30 0x6EF6 */ mov.l @r15+, r14
/* 0x0600DE32 0x000B */ rts
/* 0x0600DE34 0xE001 */ mov #1, r0
```
The @ instructions are indirect and can have a pre or post-increment.

A test is added for this case.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -94,6 +94,7 @@ class Sh2Arch(Arch):
         clobbers: List[Location] = []
         outputs: List[Location] = []
         is_return = False
+        is_load = False
         is_store = False
         has_delay_slot = False
         eval_fn: Optional[Callable[[NodeState, InstrArgs], object]] = None
@@ -142,6 +143,7 @@ class Sh2Arch(Arch):
             outputs=outputs,
             eval_fn=eval_fn,
             is_return=is_return,
+            is_load=is_load,
             is_store=is_store,
             has_delay_slot=has_delay_slot,
         )
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -5,9 +5,12 @@ from .error import DecompFailure
 from .options import Target
 from .asm_instruction import (
     Argument,
+    AsmAddressMode,
     AsmInstruction,
+    AsmLiteral,
     AsmState,
     Register,
+    Writeback,
 )
 from .instruction import (
     Instruction,
@@ -21,6 +24,7 @@ from .translate import (
     ArgLoc,
     Cast,
     Expression,
+    Literal,
     NodeState,
 )
 
@@ -32,6 +36,7 @@ class Sh2Arch(Arch):
 
     re_comment = r"!.*"
     supports_dollar_regs = False
+    supports_at_addressing = True
 
     home_space_size = 0
 
@@ -49,6 +54,7 @@ class Sh2Arch(Arch):
     saved_regs = [
         Register(r) for r in ["r8", "r9", "r10", "r11", "r12", "r13", "r14", "pr"]
     ]
+
 
     all_regs = (
         saved_regs
@@ -87,6 +93,7 @@ class Sh2Arch(Arch):
         clobbers: List[Location] = []
         outputs: List[Location] = []
         is_return = False
+        is_store = False
         has_delay_slot = False
         eval_fn: Optional[Callable[[NodeState, object], object]] = None
 
@@ -97,6 +104,30 @@ class Sh2Arch(Arch):
             has_delay_slot = True
         elif mnemonic == "nop":
             assert len(args) == 0
+        elif mnemonic == "mov":
+            assert len(args) == 2 and isinstance(args[1], Register)
+            outputs = [args[1]]
+            if isinstance(args[0], Register):
+                inputs = [args[0]]
+                eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), a.reg(0))
+            else:
+                assert isinstance(args[0], AsmLiteral)
+                eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), Literal(a.imm_value(0)))
+        elif mnemonic == "mov.l":
+            assert len(args) == 2
+            if isinstance(args[0], Register):
+                assert isinstance(args[1], AsmAddressMode)
+                assert args[1].base == cls.stack_pointer_reg
+                assert args[1].writeback == Writeback.PRE
+                inputs = [args[0], args[1].base]
+                is_store = True
+            else:
+                assert isinstance(args[0], AsmAddressMode)
+                assert isinstance(args[1], Register)
+                assert args[0].base == cls.stack_pointer_reg
+                assert args[0].writeback == Writeback.POST
+                inputs = [args[0].base]
+                outputs = [args[1]]
         else:
             raise DecompFailure(f"Unable to parse instruction: {mnemonic}")
 
@@ -109,6 +140,7 @@ class Sh2Arch(Arch):
             outputs=outputs,
             eval_fn=eval_fn,
             is_return=is_return,
+            is_store=is_store,
             has_delay_slot=has_delay_slot,
         )
 
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -37,6 +37,7 @@ class Sh2Arch(Arch):
     re_comment = r"!.*"
     supports_dollar_regs = False
     supports_at_addressing = True
+    has_delay_slots = True
 
     home_space_size = 0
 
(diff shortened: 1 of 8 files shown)


m2c/asm_instruction.py
```diff
@@ -579,15 +579,31 @@ def expect(n: str) -> str:
                 else:
                     value = BinOp(op, value, rhs)
         elif tok == "@":
+            if value is None and getattr(arch, "supports_at_addressing", False):
```
```suggestion
            if value is None and arch.supports_at_addressing:
```
and add a default value of `supports_at_addressing` on ArchAsmParsing

m2c/options.py
```diff
@@ -118,7 +118,7 @@ def parse(name: str) -> Target:
                 compiler = Target.CompilerEnum.MWCC
             elif arch == Target.ArchEnum.ARM:
                 compiler = Target.CompilerEnum.GCC
-            elif arch == Target.ArchEnum.SH:
+            elif arch == Target.ArchEnum.SH2:
```
simonlindholm, line 121: huh, I really wonder why CI did not catch this, `mypy` finds it locally for me and it looks like we run that there... I pushed this fix directly to master (edit: and found the issue with CI: fa8194e27a4b50499b528683d3a7f43fd19c65cc)

m2c/flow_graph.py
```diff
@@ -604,7 +606,7 @@ def process_no_delay_slots(item: Union[Instruction, Label]) -> None:
             block_builder.new_block()
 
     for item in body_iter:
-        if arch.arch == Target.ArchEnum.MIPS:
+        if arch.arch in (Target.ArchEnum.MIPS, Target.ArchEnum.SH2):
             process_mips(item)
```
simonlindholm: rename to `process_delay_slots`. This function will need some changes btw to handle non-mips things, like the `assert not item.is_jump(), "all MIPS jumps have a delay slot"` thing at the start (from what I gather superh does not use delay slots for conditional jumps)

tests/end_to_end/sh-return-one/sh-return-one.s
```diff
@@ -0,0 +1,7 @@
+glabel test
+/* 0x0600DE2C 0x2FE6 */ mov.l r14, @-r15
+/* 0x0600DE2E 0x6EF3 */ mov r15, r14
+/* 0x0600DE30 0x6EF6 */ mov.l @r15+, r14
+/* 0x0600DE32 0x000B */ rts
+/* 0x0600DE34 0xE001 */ mov #1, r0
+/* 0x0600DE36 */ .word 0x0009
```
simonlindholm, line 7: What's this 0x0009, out of curiosity?

Also, what disassembler are you using for this? It would be nice to eventually add tests for all existing tests/end_to_end/*/orig.c files. (One option for that to avoid a dependency on a disassembler would be to run `gcc -S`, like we do for ARM.)

m2c/flow_graph.py
```diff
@@ -604,7 +606,7 @@ def process_no_delay_slots(item: Union[Instruction, Label]) -> None:
             block_builder.new_block()
 
     for item in body_iter:
-        if arch.arch == Target.ArchEnum.MIPS:
+        if arch.arch in (Target.ArchEnum.MIPS, Target.ArchEnum.SH2):
```
simonlindholm: add a bool `arch.has_delay_slots`

m2c/asm_instruction.py
```diff
@@ -579,15 +579,31 @@ def expect(n: str) -> str:
...
+                    writeback = Writeback.PRE
+                word = parse_word(arg_elems)
+                base = replace_bare_reg(AsmGlobalSymbol(word), arch, asm_state)
+                assert isinstance(base, Register)
+                if arg_elems and arg_elems[0] == "+":
+                    assert writeback is None
+                    expect("+")
+                    writeback = Writeback.POST
+                value = AsmAddressMode(base, AsmLiteral(0), writeback)
             # A relocation (e.g. (...)@ha or (...)@l).
```
simonlindholm: move this comment within the else

m2c/arch_sh.py
```diff
@@ -97,6 +103,30 @@ def parse(
...
+                assert args[1].writeback == Writeback.PRE
+                inputs = [args[0], args[1].base]
+                is_store = True
+            else:
+                assert isinstance(args[0], AsmAddressMode)
+                assert isinstance(args[1], Register)
+                assert args[0].base == cls.stack_pointer_reg
+                assert args[0].writeback == Writeback.POST
+                inputs = [args[0].base]
+                outputs = [args[1]]
```
simonlindholm, line 132: might as well set `is_load = True` (even though that property is not used for anything outside arm atm)

m2c/asm_instruction.py
sozud: Ok done

m2c/flow_graph.py
sozud: Done

tests/end_to_end/sh-return-one/sh-return-one.s
sozud: 0x0009 is nop on this architecture, it's used as padding after a function depending on the length. The disassebmer is the one I wrote here https://github.com/sozud/saturn-splitter/blob/main/rust-dis/src/main.rs

Ok I can look into integrating the target compiler and generating these files, are we fine with using dosbox? It's a DOS compiler.

m2c/flow_graph.py
sozud: Done

m2c/asm_instruction.py
sozud: Done

m2c/arch_sh.py
sozud: Done

tests/end_to_end/sh-return-one/sh-return-one.s
simonlindholm: Sure, that's fine. We're already very inconsistent about running all the compilers for newly added tests, it's all kinda best effort and dependent on which compilers the person adding the test happened to have installed at the time, and that's fine -- most tests only really need a single compiler, and generally the e2e tests are kinda just smoke tests anyway (but it's useful to have at least a bunch for each architecture). Most of the interesting test coverage comes from running m2c on large corpuses.

m2c/arch_sh.py
```diff
@@ -109,6 +142,7 @@ def parse(
             outputs=outputs,
             eval_fn=eval_fn,
             is_return=is_return,
+            is_store=is_store,
```
simonlindholm, line 147: and `is_load=is_load,` here

m2c/flow_graph.py
```diff
@@ -413,8 +413,10 @@ def build_blocks(
     fragment: bool,
     debug_patterns: bool,
 ) -> List[Block]:
-    if arch.arch == Target.ArchEnum.MIPS:
+    if arch.arch in (Target.ArchEnum.MIPS, Target.ArchEnum.SH2):
```
simonlindholm:
```suggestion
    if arch.has_delay_slots:
```

m2c/asm_instruction.py
```diff
@@ -579,15 +581,31 @@ def expect(n: str) -> str:
...
+                base = replace_bare_reg(AsmGlobalSymbol(word), arch, asm_state)
+                assert isinstance(base, Register)
+                if arg_elems and arg_elems[0] == "+":
+                    assert sh_writeback is None
+                    expect("+")
+                    sh_writeback = Writeback.POST
+                value = AsmAddressMode(base, AsmLiteral(0), sh_writeback)
+            elif not top_level:
                 # Parse a+b@l as (a+b)@l, not a+(b@l)
                 break
```
simonlindholm: makes more sense to move this into the else, after the comment about it being a relocation

m2c/arch_sh.py
sozud: Added

m2c/flow_graph.py
sozud: Done

m2c/asm_instruction.py
sozud: Moved

tests/end_to_end/sh-return-one/sh-return-one.s
sozud: Ok I figure this makes sense as a separate PR, I can see about getting this integrated

simonlindholm: For sure. It's more useful when the arch support is nearing completion, the only upsides of doing it earlier is that it gets rid of the need to write tests for individual PRs (e.g. the tests in both this and the earlier PR I think we can remove), and that it's a nice feeling watching test outputs grow successively better.

PR #318, opened 2026-07-15
sh2: function args, use gcc for test asm

sozud:
This adds some basic support for arguments enabling
```
s32 _test(s32 arg0) {
    return arg0;
}
```

It also replaces the test asm with sh-gcc output and adds some infrastructure to generate and check the asm. I didn't include the actual parts to get the compiler running since none of the other compilers are set up that way.


tests/add_test.py
```diff
@@ -177,13 +200,15 @@ def do_compilation_step(
...
     ]
-    subprocess.run(args)
+    subprocess.run(args, check=True)
 
 
 def run_compile(in_file: Path, out_file: Path, compiler: Compiler) -> None:
     flags_str = " ".join(compiler.cc_command)
     logger.info(f"Compiling {in_file} to {out_file} using: {flags_str}")
-    if compiler.name == "agbcc":
+    if compiler.name == "sh-gcc":
```
simonlindholm, line 209: Can this reuse the agbcc codepath, with first preprocessing to a .i file? No big deal if not, just saves two lines of code

m2c/arch_sh.py
```diff
@@ -163,8 +163,37 @@ def function_abi(
         for_call: bool,
     ) -> Abi:
         known_slots: List[AbiArgSlot] = []
+        candidate_slots: List[AbiArgSlot] = []
         possible_slots: List[AbiArgSlot] = []
 
+        if fn_sig.params_known:
```
simonlindholm, line 169: This should also copy stuff about struct return type, >4-byte arguments, and varargs from arm. (If varargs even works the same way? I hear some ABIs put all varargs on the stack.) Could do that later though

m2c/arch_sh.py
```diff
@@ -163,8 +163,37 @@ def function_abi(
...
+                AbiArgSlot(ArgLoc(None, i, Register(f"r{i + 4}")), Type.intptr())
+                for i in range(4)
+            ]
+
+        # argument registers are filled in order, so using a higher register implies
+        # that the preceding registers may also contain arguments
+        highest_used = -1
+        for i, reg in enumerate(self.argument_regs):
+            if likely_regs.get(reg, False):
+                highest_used = i
```
simonlindholm, line 192: You'll want to copy some logic from ARM here, with e.g. breaking if `reg not in likely_regs`. One of the effects the old code wants to achieve is that when calling a function, if `r5` contains a value (because it's been used as a temp reg) but `r4` does not, the function will not be called as `foo(ERROR(read from unset reg r4), the_r5_value)`. It's also nice to keep the code similar to other architectures for maintainability reasons (even though I know the existing code is rather a mess).

sozud: I'm not really able to test any of these abi features because I would have to implement more instructions first. So I'd be just guessing about the correctness basically. Is it fine to keep your two comments as TODOs for the moment and I come back to them once I have the right infrastructure in place?

tests/add_test.py
sozud: So the way I set up the build was just using the real gcc frontend. Would you like me to set it up using modern cpp and tthe .i file approach? I think there's possibly some additional complexity due to dos/unix line endings but I think it can work that way too.

m2c/arch_sh.py
simonlindholm: Sure, sounds reasonable

tests/add_test.py
simonlindholm: I guess I'm just surprised at seeing a compiler output asm but accept as input an unpreprocessed file. But if it's correct then I suppose it's fine. Are line endings an issue here, as in all the orig.c files need to have CRLF line endings for the compiler to work?

sozud: I'm calling gcc.exe so it's the regular frontend calling cpp. Is orig.c supposed to be the post-processed source?

simonlindholm: Well, we're never using any macros so the difference doesn't matter. (I'm so confused, why does the regular frontend emit asm without any special -S flag? But oh well)

sozud: Oh I updated my setup and this PR to optionally use modern cpp but I see this already got merged. I can include the change with the next one.

For more clarity what I have setup is docker container with dosemu and the compiler in it. I run a script like this inside the container that has the arguments. I didn't include the toolchain setup since it's complex and I thought there would be a lot of back and forth about it.

https://gist.github.com/sozud/27e2093171cf334da5f7e813f270f210#file-gistfile1-txt-L42

simonlindholm: Aha, that explains it! We ought to link that somewhere, so it's possible for others to add additional tests in the future. No for modern cpp fanciness if it needs extra work, I really was just aiming to save two lines of code.

sozud: Ok I'll put together a PR for this and if it's bothersome to merge it can just be a draft for people to look at.

PR #319, opened 2026-07-16
sh2: add ops: add, sub, tst, bt.s, bra

sozud:
This adds some additional ops to handle

```
int test(int a, int b) { return a + b; }
int minus(int a, int b) { return a - b; }
```

and
```
int test(int x) {
    if (x == 0) {
        return 1;
    }
    return 2;
}
```

The branching test output is wrong but I'm not sure if that's a problem with my implementation or just how m2c works.

```
s32 _test(s32 arg0) {
    // conditional swapped
    if (arg0 != 0) {
        return 2;     
    }
    return 1;
}
```

I also combined the add_test.py codepath since my tooling can handle .i files now.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from dataclasses import replace
 from typing import Callable, Dict, List, Optional
 
 from .error import DecompFailure
@@ -26,12 +27,13 @@ from .translate import (
     BinaryOp,
     Cast,
     Expression,
+    InstrMap,
     InstrArgs,
     Literal,
     NodeState,
 )
 
-from .evaluate import condition_from_expr
+from .evaluate import condition_from_expr, handle_add, handle_sub
 
 from .types import FunctionSignature, Type
 
@@ -137,7 +139,7 @@ class Sh2Arch(Arch):
                 inputs = [args[0].base]
                 outputs = [args[1]]
                 is_load = True
-        elif mnemonic in ("add", "sub"):
+        elif mnemonic in cls.instrs_arithmetic:
             assert (
                 len(args) == 2
                 and isinstance(args[0], Register)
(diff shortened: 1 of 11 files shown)


m2c/arch_sh.py
```diff
@@ -131,6 +137,52 @@ def parse(
...
+                len(args) == 2
+                and isinstance(args[0], Register)
+                and isinstance(args[1], Register)
+            )
+            inputs = [args[0], args[1]]
+            outputs = [args[1]]
+            op = "+" if mnemonic == "add" else "-"
+            eval_fn = lambda s, a: s.set_reg(
+                a.reg_ref(1), BinaryOp.intptr(a.reg(1), op, a.reg(0))
+            )
```
simonlindholm, line 152: This is fine for now, but will eventually want to use the functions from evaluate.py for handling add/sub which also deal with taking addresses to struct members, combining small and large literal addends, that sort of thing. You'll probably also want to move to the format used by the other arch's with lookup tables that map intrinsics to handler lambdas, to keep code size down.

simonlindholm:
> The branching test output is wrong but I'm not sure if that's a problem with my implementation or just how m2c works.

It's just how m2c works. We can look into it later once we get more data (would need changes to if_statements.py most likely).

m2c/arch_sh.py
sozud: Ok adjusted this to use InstrMap and the handlers

m2c/arch_sh.py
```diff
@@ -202,6 +203,18 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
             has_delay_slot=has_delay_slot,
         )
 
+    instrs_arithmetic: InstrMap = {
+        # sh2 format is src, dst
+        # add handler is dest, left, right
```
simonlindholm, line 208: Oh, this is rough... Wondering if we should change all our helpers to accept individual arguments instead of InstrArgs, or if we should perhaps just be lazy and go the route you're taking (which could be made rather concise if we add a helper on InstrArgs so it's just `handle_add(a.reorder_args(1, 1, 0))`)

PR #320, opened 2026-07-16
sh2: add minimal ABI coverage

sozud:
This is a follow-up to https://github.com/matt-kempster/m2c/pull/318 to try and improve the function ABI handling.

I'm testing these cases:

5 arguments, the 5th will go on the stack.
```
int test(int a, int b, int c, int d, int e) {
    return a + b + c + d + e;
}
```

Hidden struct return, ret in r2, value in r4
```
// 12 byte struct, 
typedef struct {
    int a;
    int b;
    int c;
} Large;

Large test(int value) {
    Large ret = {value, value + 1, value + 2};
    return ret;
}
```

16 bit value in literal pool (not really abi related but sh2 only has 8-bit immediate intructions)
```
typedef signed int s32;

s32 test(void) { return 2345; }
```

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -33,7 +33,14 @@ from .translate import (
     NodeState,
 )
 
-from .evaluate import condition_from_expr, handle_add, handle_sub
+from .evaluate import (
+    condition_from_expr,
+    handle_add,
+    handle_addi,
+    handle_load,
+    handle_sub,
+    make_store,
+)
 
 from .types import FunctionSignature, Type
 
@@ -123,29 +130,44 @@ class Sh2Arch(Arch):
             else:
                 assert isinstance(args[0], AsmLiteral)
                 eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), Literal(a.imm_value(0)))
-        elif mnemonic == "mov.l":
+        elif mnemonic in ("mov.l", "mov.w"):
             assert len(args) == 2
             if isinstance(args[0], Register):
                 assert isinstance(args[1], AsmAddressMode)
-                assert args[1].base == cls.stack_pointer_reg
-                assert args[1].writeback == Writeback.PRE
                 inputs = [args[0], args[1].base]
                 is_store = True
+                if args[1].writeback is None:
+                    def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                        store = make_store(a, Type.reg32(likely_float=False))
+                        if store is not None:
+                            s.store_memory(store, a.reg_ref(0))
             else:
-                assert isinstance(args[0], AsmAddressMode)
                 assert isinstance(args[1], Register)
-                assert args[0].base == cls.stack_pointer_reg
-                assert args[0].writeback == Writeback.POST
-                inputs = [args[0].base]
+                if isinstance(args[0], AsmAddressMode):
+                    inputs = [args[0].base]
                 outputs = [args[1]]
                 is_load = True
+                if not (
+                    isinstance(args[0], AsmAddressMode)
+                    and args[0].writeback is not None
+                ):
+                    load_type = (
+                        Type.s16()
+                        if mnemonic == "mov.w"
+                        else Type.reg32(likely_float=False)
+                    )
+                    eval_fn = lambda s, a: s.set_reg(
+                        a.reg_ref(1),
+                        handle_load(
+                            replace(a, raw_args=[a.raw_arg(1), a.raw_arg(0)]),
+                            type=load_type,
+                        ),
+                    )
         elif mnemonic in cls.instrs_arithmetic:
-            assert (
-                len(args) == 2
-                and isinstance(args[0], Register)
-                and isinstance(args[1], Register)
-            )
-            inputs = [args[0], args[1]]
+            assert len(args) == 2 and isinstance(args[1], Register)
+            inputs = [args[1]]
+            if isinstance(args[0], Register):
+                inputs.insert(0, args[0])
             outputs = [args[1]]
             eval_fn = lambda s, a: s.set_reg(
                 a.reg_ref(1), cls.instrs_arithmetic[mnemonic](a)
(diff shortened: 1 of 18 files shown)


tests/end_to_end/sh-abi-struct-return/context.c
```diff
@@ -0,0 +1,7 @@
+typedef struct {
+    int a;
+    int b;
+    int c;
+} Large;
+
+Large _test(int value);
```
simonlindholm, line 7: (future work: we ought to strip leading underscores as a demangling step, both for output and for correlating asm symbols with the context)

m2c/translate.py
simonlindholm: Move the sh framepointer code next to here

m2c/arch_sh.py
```diff
@@ -123,29 +130,46 @@ def parse(
...
-        elif mnemonic == "mov.l":
+        elif mnemonic in ("mov.l", "mov.w"):
             assert len(args) == 2
             if isinstance(args[0], Register):
                 assert isinstance(args[1], AsmAddressMode)
-                assert args[1].base == cls.stack_pointer_reg
-                assert args[1].writeback == Writeback.PRE
                 inputs = [args[0], args[1].base]
                 is_store = True
+                if args[1].writeback is None:
```
simonlindholm, line 139: else keep asserting that it's the stack pointer, I guess

m2c/asm_instruction.py
```diff
@@ -582,8 +582,24 @@ def expect(n: str) -> str:
...
+                        asm_state,
+                        top_level=False,
+                        do_replace_bare_reg=False,
+                    )
+                    expect(",")
+                    word = parse_word(arg_elems)
+                    base = replace_bare_reg(AsmGlobalSymbol(word), arch, asm_state)
+                    assert isinstance(base, Register)
+                    expect(")")
+                    value = AsmAddressMode(base, constant_fold(addend, asm_state), None)
```
```suggestion
                    value = AsmAddressMode(base, addend, None)
```
since `parse_arg_elems` already constant folds by itself (unless you pass `do_constant_fold=False`)

m2c/evaluate.py
```diff
@@ -468,13 +468,19 @@ def load_rodata_constant() -> Optional[Expression]:
...
             # Do allow non-zero offsets: they occur in raw agbcc output which
             # we use for tests.
             return None
-        if not is_arm and (not type.is_likely_float() or expr.offset != 0):
+        if (
+            not is_arm
+            and not is_sh
+            and (not type.is_likely_float() or expr.offset != 0)
+        ):
             # For non-ARM, only allow float constants and offset 0.
```
simonlindholm:
```suggestion
            # Outside of ARM/SH, only allow float constants and offset 0.
```

m2c/arch_sh.py
sozud: Do you mean like this?
```
                assert isinstance(args[1], AsmAddressMode)
                inputs = [args[0], args[1].base]
                is_store = True
                if args[1].writeback is None:

                    def eval_fn(s: NodeState, a: InstrArgs) -> None:
                        store = make_store(a, Type.reg32(likely_float=False))
                        if store is not None:
                else:
                    assert isinstance(args[1], AsmAddressMode)
```

m2c/asm_instruction.py
sozud: Done

m2c/evaluate.py
sozud: Done

m2c/translate.py
sozud: Done

tests/end_to_end/sh-abi-struct-return/context.c
sozud: Alright can look into this next

m2c/translate.py
```diff
@@ -699,6 +720,13 @@ def get_stack_info(
                         allowed_callee_save_gaps.add(stack_offset + 4)
         elif arch_mnemonic == "ppc:mflr" and inst.args[0] == Register("r0"):
             info.is_leaf = False
+        elif (
+            arch_mnemonic == "sh2:mov"
+            and inst.args[0] == arch.stack_pointer_reg
+            and isinstance(inst.args[1], Register)
+            and inst.args[1] in arch.frame_pointer_regs
+        ):
+            info.frame_pointer_reg = inst.args[1]
```
simonlindholm: Sorry, I think I managed to make a confusing suggestion -- I meant to move *this* block specifically up there, so that the two blocks whose bodies both go `info.frame_pointer_reg = inst.args[1]` are next to each other (operand order making them unfortunately hard to merge).

You could do the same for the block that goes `info.allocated_stack_size += -inst.args[0].value` I guess.

BTW, unrelated, but you'll in relation to this code also at some point want to look into how your compiler handles allocating large stack frames. `./tests/add_test.py tests/end_to_end/big-stack/orig.c` to get the test for that added (you can comment the rest of the architectures out so those don't update at the same time).

m2c/arch_sh.py
simonlindholm:
```py
                assert args[1].base == cls.stack_pointer_reg
                assert args[1].writeback == Writeback.PRE
```
like you had it before

m2c/translate.py
sozud: I only see one instance of `info.frame_pointer_reg = inst.args[1]`, not two blocks? Sorry for the confusion

OK I can look into the large stack one

simonlindholm: There is
```py
        elif (
            arch_mnemonic in ("mips:move", "arm:mov", "ppc:mr")
            and isinstance(inst.args[0], Register)
            and inst.args[0] in arch.frame_pointer_regs
            and inst.args[1] == arch.stack_pointer_reg
        ):
            # "move fp, sp" very likely means the code is compiled with frame
            # pointers enabled; thus fp should be treated the same as sp.
            info.frame_pointer_reg = inst.args[0]
```
and your newly added version
```py
        elif (
            arch_mnemonic == "sh2:mov"
            and inst.args[0] == arch.stack_pointer_reg
            and isinstance(inst.args[1], Register)
            and inst.args[1] in arch.frame_pointer_regs
        ):
            info.frame_pointer_reg = inst.args[1]
```
that do materially the same thing and thus ought to next to each other

sozud: Ok done

m2c/arch_sh.py
sozud: Ok done

PR #321, opened 2026-07-17
sh2: support jump tables, demangle asm

sozud:
This adds support for 3 cases, a jump table non-zero-based:
```
int test(int value) {
    switch (value) {
    case 10:
        return 11;
    case 11:
        return 42;
    case 12:
        return 19;
    case 13:
        return 73;
    default:
        return -7;
    }
}
```

zero-based jump table
```
int test(int value) {
    switch (value) {
    case 0:
        return 11;
    case 1:
        return 42;
    case 2:
        return 19;
    case 3:
        return 73;
    default:
        return -7;
    }
}
```

And a small switch that doesn't generate a jump table
```
int test(int value) {
    switch (value) {
    case 0:
        return 11;
    case 1:
        return 42;
    case 2:
        return 19;
    default:
        return -7;
    }
}
```

This PR also removes the `_` from the test cases, and normalizes the line endings. Previously there were dos line endings mixed with unix ones.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -18,7 +18,7 @@ from .asm_instruction import (
     Writeback,
     get_jump_target,
 )
-from .asm_pattern import AsmMatcher, AsmPattern, Replacement
+from .asm_pattern import AsmMatch, Replacement, SimpleAsmPattern, make_pattern
 from .instruction import (
     Instruction,
     InstructionMeta,
@@ -50,35 +50,24 @@ from .evaluate import (
 from .types import FunctionSignature, Type
 
 
-class ShJumpTablePattern(AsmPattern):
-    def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
-        body = matcher.input[matcher.index : matcher.index + 7]
-        instructions = [x for x in body if isinstance(x, Instruction)]
-        if len(instructions) != 7:
-            return None
-        mov, double, mova, load, add, jump, nop = instructions
-        if [x.mnemonic for x in instructions] != [
-            "mov",
-            "add",
-            "mova",
-            "mov.w",
-            "add",
-            "jmp",
-            "nop",
-        ]:
-            return None
-        if (
-            len(mov.args) != 2
-            or not isinstance(mov.args[0], Register)
-            or not isinstance(mov.args[1], Register)
-            or double.args != [mov.args[1], mov.args[1]]
-            or len(mova.args) != 2
-            or not isinstance(mova.args[0], AsmGlobalSymbol)
-        ):
-            return None
+class JumpTablePattern(SimpleAsmPattern):
+    pattern = make_pattern(
+        "mov $x, $i",
+        "add $i, $i",
+        "mova _, $b",
+        "mov.w",
+        "add $i, $b",
+        "jmp",
+        "nop",
+    )
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        mova = m.body[2]
+        assert isinstance(mova, Instruction)
+        assert isinstance(mova.args[0], AsmGlobalSymbol)
 
         table_name = mova.args[0].symbol_name
-        table = matcher.asm_data.values.get(table_name)
+        table = m.asm_data.values.get(table_name)
         if table is None:
             return None
         targets: List[AsmGlobalSymbol] = []
@@ -96,10 +85,10 @@ class ShJumpTablePattern(AsmPattern):
             return None
         return Replacement(
             [
-                AsmInstruction("tablejmp.fictive", [mov.args[0], *targets]),
+                AsmInstruction("tablejmp.fictive", [m.regs["x"], *targets]),
                 AsmInstruction("nop", []),
             ],
-            len(body),
+            len(m.body),
         )
 
 
@@ -350,7 +339,7 @@ class Sh2Arch(Arch):
         "sub": lambda a: handle_sub(a.reg(1), a.reg(0)),
     }
 
-    asm_patterns = [ShJumpTablePattern()]
+    asm_patterns = [JumpTablePattern()]
 
     def arg_name(self, loc: ArgLoc) -> str:
         if loc.offset is not None:
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -31,6 +31,8 @@ from .translate import (
     ArgLoc,
     BinaryOp,
     Cast,
+    ErrorExpr,
+    ExprStmt,
     Expression,
     InstrMap,
     InstrArgs,
@@ -55,9 +57,9 @@ class JumpTablePattern(SimpleAsmPattern):
         "mov $x, $i",
         "add $i, $i",
         "mova _, $b",
-        "mov.w",
+        "mov.w",  # "mov.w @($b,$i),$i"
         "add $i, $b",
-        "jmp",
+        "jmp",  # "jmp @$b"
         "nop",
     )
 
@@ -177,13 +179,6 @@ class Sh2Arch(Arch):
             else:
                 assert isinstance(args[0], AsmLiteral)
                 eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), Literal(a.imm_value(0)))
-        elif mnemonic == "mova":
-            assert (
-                len(args) == 2
-                and isinstance(args[0], AsmGlobalSymbol)
-                and isinstance(args[1], Register)
-            )
-            outputs = [args[1]]
         elif mnemonic in ("mov.l", "mov.w"):
             assert len(args) == 2
             if isinstance(args[0], Register):
@@ -312,7 +307,11 @@ class Sh2Arch(Arch):
                 s.set_reg(Register("condition_bit"), condition)
 
         else:
-            raise DecompFailure(f"Unable to parse instruction: {mnemonic}")
+            instr_str = str(AsmInstruction(mnemonic, args))
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                error = ErrorExpr(f"unknown instruction: {instr_str}")
+                s.write_statement(ExprStmt(error))
 
         return Instruction(
             mnemonic=mnemonic,
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -53,17 +53,11 @@ from .types import FunctionSignature, Type
 class ShJumpTablePattern(AsmPattern):
     def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
         body = matcher.input[matcher.index : matcher.index + 7]
-        if len(body) != 7 or not all(isinstance(x, Instruction) for x in body):
+        instructions = [x for x in body if isinstance(x, Instruction)]
+        if len(instructions) != 7:
             return None
-        mov, double, mova, load, add, jump, nop = body
-        assert isinstance(mov, Instruction)
-        assert isinstance(double, Instruction)
-        assert isinstance(mova, Instruction)
-        assert isinstance(load, Instruction)
-        assert isinstance(add, Instruction)
-        assert isinstance(jump, Instruction)
-        assert isinstance(nop, Instruction)
-        if [x.mnemonic for x in body] != [
+        mov, double, mova, load, add, jump, nop = instructions
+        if [x.mnemonic for x in instructions] != [
             "mov",
             "add",
             "mova",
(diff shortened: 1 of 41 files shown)


m2c/arch_sh.py
```diff
@@ -45,6 +50,59 @@
 from .types import FunctionSignature, Type
 
 
+class ShJumpTablePattern(AsmPattern):
```
simonlindholm: You can use a SimpleAsmPattern for this, which should make this shorter/more readable.

The mov.w is a bit annoying parsing-wise. Simple solution is to leave it as just `"mov.w"` which the pattern matcher will accept as "any arguments", and then if needed manually check stuff in `replace` (SimpleAsmPattern's replacement for `match`).

Alternative, probably better, solution is to add `supports_at_addressing = True` to `NaiveParsingArch` and make the `@` parsing code able to parse `$`-prefixed registers by using `parse_arg_elem` instead of `parse_word` + `replace_bare_reg`.

(Maybe the setup there ought to change so we can use our actual arch for parsing instead of this fake one that requires `$`-prefixed registers, but that's a more involved change)

m2c/arch_sh.py
```diff
@@ -130,6 +188,13 @@ def parse(
             else:
                 assert isinstance(args[0], AsmLiteral)
                 eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), Literal(a.imm_value(0)))
+        elif mnemonic == "mova":
+            assert (
+                len(args) == 2
+                and isinstance(args[0], AsmGlobalSymbol)
+                and isinstance(args[1], Register)
+            )
+            outputs = [args[1]]
```
simonlindholm: I assume you'll implement this at a later point? (Handling it like this makes it silently ignored which is worse than an error, but as long as this is just a temporary state that's totally fine)

m2c/translate.py
```diff
@@ -727,6 +747,13 @@ def get_stack_info(
                         allowed_callee_save_gaps.add(stack_offset + 4)
         elif arch_mnemonic == "ppc:mflr" and inst.args[0] == Register("r0"):
             info.is_leaf = False
+        elif (
+            arch_mnemonic == "sh2:mov"
+            and inst.args[0] == arch.stack_pointer_reg
+            and isinstance(inst.args[1], Register)
+            and inst.args[1] in arch.frame_pointer_regs
+        ):
+            info.frame_pointer_reg = inst.args[1]
```
simonlindholm: This part got accidentally duplicated

m2c/arch_sh.py
```diff
@@ -241,6 +350,8 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
         "sub": lambda a: handle_sub(a.reg(1), a.reg(0)),
     }
 
+    asm_patterns = [ShJumpTablePattern()]
```
simonlindholm: remove Sh from the name since this file is all sh stuff

tests/add_test.py
```diff
@@ -203,6 +204,17 @@ def do_compilation_step(
...
 
+def demangle_sh_asm(asm_text: str) -> str:
+    symbols = re.findall(r"(?m)^\s*\.global\s+(_[A-Za-z][A-Za-z0-9_]*)\s*$", asm_text)
+    for symbol in symbols:
+        asm_text = re.sub(
+            rf"(?<![A-Za-z0-9_]){re.escape(symbol)}(?![A-Za-z0-9_])",
+            symbol[1:],
+            asm_text,
+        )
+    return re.sub(r"[ \t]+(?=\r?$)", "", asm_text, flags=re.MULTILINE)
```
simonlindholm, line 215: Ah, I was thinking some demangling in actual m2c, since I imagine your decomp also suffers from leading underscores? For now this is a nice change though, better than cluttering *-flags.txt files.

tests/end_to_end/sh-switch-jumptable-nonzero/orig.c
```diff
@@ -0,0 +1,14 @@
...
+    case 11:
+        return 42;
+    case 12:
+        return 19;
+    case 13:
+        return 73;
+    default:
+        return -7;
+    }
+}
```
simonlindholm, line 14: Instead of adding new tests it would be nice if you could make use of existing ones, like `tests/end_to_end/switch/orig.c` in this case.

tests/end_to_end/sh-switch-jumptable-nonzero/sh2-gcc-o2-flags.txt
```diff
@@ -0,0 +1 @@
+--context context.c --target sh2-gcc-c
```
simonlindholm: doesn't need a separate context file, can use orig.c (now that we deal with underscores)

m2c/arch_sh.py
sozud: Ok adjusted to use SimpleAsmPattern. I'm thinking we can do the deeper changes if it turns out I need more patterns that use @ instructions?

m2c/arch_sh.py

m2c/translate.py
sozud: Bad rebase, fixed

tests/add_test.py
sozud: I suppose the only real effect for us is that user_syms.txt needs to be like `_name` so it hasn't been too bothersome.

tests/end_to_end/sh-switch-jumptable-nonzero/orig.c
sozud: Ok right now I don't have multiplication or division so I can't use `tests/end_to_end/switch/orig.c`. Would it be fine to do a pass to reduce/simplify the e2e tests later?

tests/end_to_end/sh-switch-jumptable-nonzero/sh2-gcc-o2-flags.txt
sozud: If I remove the contexts I get diffs e.g.

```
-s32 test(s32 a, s32 b, s32 c, s32 d, s32 e) {
-    return a + b + c + d + e;
+s32 test(s32 arg0, s32 arg1, s32 arg2, s32 arg3, s32 arg4) {
+    return arg0 + arg1 + arg2 + arg3 + arg4;
 }
```

```
-s32 test(s32 value) {
-    switch (value) {
+s32 test(u32 arg0) {
+    switch (arg0) {
     case 0:
         return 0xB;
     case 1:
```

```
-Large test(Large *__return__, s32 value) {
+void test(s32 arg0) {
     s32 sp0;
     s32 sp4;
     s32 sp8;
     s32 temp_r1;
     s32 temp_r1_2;
+    void *temp_r0;
 
-    sp0 = value;
-    temp_r1 = value + 1;
+    temp_r0 = M2C_ERROR(/* Read from unset register $r2 */);
+    sp0 = arg0;
+    temp_r1 = arg0 + 1;
     sp4 = temp_r1;
-    temp_r1_2 = value + 2;
+    temp_r1_2 = arg0 + 2;
     sp8 = temp_r1_2;
-    __return__->a = value;
-    __return__->b = temp_r1;
-    __return__->c = temp_r1_2;
-    return (Large) __return__;
+    temp_r0->unk0 = arg0;
+    temp_r0->unk4 = temp_r1;
+    temp_r0->unk8 = temp_r1_2;
 }
```

Do you mean just remove it for this specific test?

m2c/arch_sh.py
sozud: Yes probably if gcc emits it outside of the jump table pattern (I'm  not sure). Is there a way to warn? I have to have handling here or it gets stuck

tests/end_to_end/sh-switch-jumptable-nonzero/sh2-gcc-o2-flags.txt
simonlindholm: I mean --context orig.c

tests/end_to_end/sh-switch-jumptable-nonzero/orig.c
simonlindholm: Good point, sure

tests/add_test.py
simonlindholm: Hm, you'll have problems when you want to use context files though, right?

m2c/arch_sh.py
simonlindholm: You'll want to have an eval_fn that does mimics the "unknown instruction:" case of arm/mips/ppc (you'll want one of those as a fallback anyway). Can probably just have that fallback and remove this explicit handling code

m2c/arch_sh.py
simonlindholm: Sure, sounds good

m2c/arch_sh.py
```diff
@@ -45,6 +50,48 @@
 from .types import FunctionSignature, Type
 
 
+class JumpTablePattern(SimpleAsmPattern):
+    pattern = make_pattern(
+        "mov $x, $i",
+        "add $i, $i",
+        "mova _, $b",
+        "mov.w",
```
```suggestion
        "mov.w",  # "mov.w @($b,$i),$i"
```

m2c/arch_sh.py
```diff
@@ -45,6 +50,48 @@
...
+        "add $i, $b",
+        "jmp",
```
```suggestion
        "jmp",  # "jmp @$b"
```

m2c/arch_sh.py
```diff
@@ -45,6 +50,48 @@
...
+        targets: List[AsmGlobalSymbol] = []
+        for entry in table.data:
+            if (
+                not isinstance(entry, AsmSymbolicData)
+                or not isinstance(entry.data, BinOp)
+                or entry.data.op != "-"
+                or not isinstance(entry.data.lhs, AsmGlobalSymbol)
+                or entry.data.rhs != AsmGlobalSymbol(table_name)
+            ):
+                return None
```
simonlindholm, line 84: (I wonder if you should `break` instead, to handle trailing undisassembled data, like nul bytes)

m2c/translate.py
```diff
@@ -638,6 +638,26 @@ def get_stack_info(
...
+            and inst.args[0] in arch.saved_regs
+            and isinstance(inst.args[1], AsmAddressMode)
+            and inst.args[1].base == arch.stack_pointer_reg
+            and inst.args[1].writeback == Writeback.PRE
+        ):
+            info.allocated_stack_size += 4
+            info.callee_save_regs.add(inst.args[0])
+            callee_saved_offsets.append(-info.allocated_stack_size)
+            if inst.args[0] == arch.return_address_reg:
+                info.is_leaf = False
```
simonlindholm: also part of the bad rebase

m2c/arch_sh.py
sozud: Added

m2c/arch_sh.py
sozud: Added

m2c/arch_sh.py
sozud: The idea is this could run on random data that appears to be a jump table at first but actually isn't? I think if the disassembler is right you'd only have .data entries so you'd never get here?

m2c/arch_sh.py

m2c/translate.py
sozud: Fixed

tests/end_to_end/sh-switch-jumptable-nonzero/sh2-gcc-o2-flags.txt
sozud: Ok removed

tests/add_test.py
sozud: Yeah I suppose you're right, never thought about this since m2c never has had any sh2 support. Do any other backends need to demangle as an example?

sozud: I think actually I need to revert this part then otherwise the demangling you'd need on a real project can't be exercised

simonlindholm: No other arch needs demangling currently, though the x86 PR in flight (#303) also needs it. Haven't verified whether it does it in the best way.

m2c/arch_sh.py
simonlindholm: I was thinking something like
```
.word L2-L1
.word L3-L1
.word 0
```
with end-of-file padding, but looking at the asm it's an inline literal pool next to the jump itself, so that cannot happen. In theory maybe the first case could start with a loop and GCC want to over-align the start label, but eh. Let's revisit if it ever comes up I guess.

tests/end_to_end/sh-switch-jumptable/sh2-gcc-o2.s
```diff
@@ -0,0 +1,52 @@
...
+	mov.w	@(r0,r1),r1
+	add	r1,r0
+	jmp        @r0
+	nop
+	.align 2
+L8:
+	.word	L3-L8
+	.word	L4-L8
+	.word	L5-L8
+	.word	L6-L8
```
simonlindholm, line 34: Hm, .word here means a two-byte value, right? That's different from arm/mips/ppc, we ought to make `parse_file` aware of the discrepancy.

(I'm confused at the `.align 2`, surely `.align 1` would make more sense.)

PR #322, opened 2026-07-18
Sh2 function calls and better handling of _

sozud:
This adds support for performing a function call, for example
```
int callee(int first, int second);
extern int global;

int test(int value) {
    return callee(value, value + 1) + global;
}
```

This also improves the handling of `_`. The simplest solution seemed to be adding `c_symbol_name` to Arch and letting that be overridden by SH.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -38,6 +38,7 @@ from .translate import (
     InstrArgs,
     Literal,
     NodeState,
+    as_u32,
 )
 
 from .evaluate import (
@@ -316,7 +317,7 @@ class Sh2Arch(Arch):
             )
             target_reg = args[0].base
             inputs = [*cls.argument_regs, target_reg]
-            outputs = [reg for reg, _ in cls.base_return_regs]
+            outputs = list(cls.all_return_regs)
             clobbers = list(cls.temp_regs)
             function_target = target_reg
             has_delay_slot = True
@@ -454,5 +455,8 @@ class Sh2Arch(Arch):
         return {
             Register("r0"): Cast(
                 expr, reinterpret=True, silent=True, type=Type.intptr()
+            ),
+            Register("r1"): as_u32(
+                Cast(expr, reinterpret=True, silent=False, type=Type.u64())
             ),
         }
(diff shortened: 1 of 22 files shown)


m2c/arch_sh.py
```diff
@@ -277,6 +308,19 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
 
+        elif mnemonic == "jsr":
+            assert (
+                len(args) == 1
+                and isinstance(args[0], AsmAddressMode)
+                and args[0].addend == AsmLiteral(0)
+            )
+            target_reg = args[0].base
+            inputs = [*cls.argument_regs, target_reg]
+            outputs = [reg for reg, _ in cls.base_return_regs]
```
simonlindholm: should be able to use `all_return_regs` here, like other archs

and then either add `r1` to `function_return`, with some u64 cast or something (this almost never comes up in practice, but it's there to marginally improve output and not just generate a "read from unset register" when it does), or remove `r1` from `all_return_regs`

m2c/translate.py
```diff
@@ -4244,7 +4249,7 @@ def address_of_gsym(self, sym_name: str) -> AddressOf:
                     demangled_str = str(demangled_symbol)
 
             sym = self.global_symbol_map[sym_name] = GlobalSymbol(
-                symbol_name=sym_name,
+                symbol_name=c_sym_name,
```
simonlindholm: I'm thinking we should maybe rename `symbol_name` -> `c_symbol_name` now to make it easier to not accidentally introduce uses of it where it's used to look up asm-level data (or other way around: using non-C names for looking up symbols in a context)

simonlindholm (commented):
Very clean!

m2c/translate.py
sozud: Ok renamed

m2c/evaluate.py
```diff
@@ -490,8 +490,7 @@ def load_rodata_constant() -> Optional[Expression]:
         ):
             return None
 
-        sym_name = target.expr.symbol_name
-        ent = args.stack_info.global_info.asm_data_value(sym_name)
```
simonlindholm, line 494: hah, I feel vindicated on my review comment, good catch! (and nice cleanup too)

m2c/arch_sh.py
sozud: Ok adjusted this to be like mips

simonlindholm (approved):

PR #323, opened 2026-07-18
sh2: five argument handling, shift/rotate, binary ops

sozud:
This adds abi fixing to handle five arguments as well as most src/dest binary operations like shifts, rotates, add, etc.


m2c/arch_sh.py
```diff
@@ -384,6 +444,44 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+
+    instrs_shift: InstrMap = {
+        "shlr": lambda a: BinaryOp.uint(a.reg(0), ">>", Literal(1)),
+        "shar": lambda a: BinaryOp.sint(a.reg(0), ">>", Literal(1)),
+        "shll2": lambda a: BinaryOp.uint(a.reg(0), "<<", Literal(2)),
+        "shlr2": lambda a: BinaryOp.uint(a.reg(0), ">>", Literal(2)),
+        "shll8": lambda a: BinaryOp.uint(a.reg(0), "<<", Literal(8)),
+        "shlr8": lambda a: BinaryOp.uint(a.reg(0), ">>", Literal(8)),
+        "shll16": lambda a: BinaryOp.uint(a.reg(0), "<<", Literal(16)),
+        "shlr16": lambda a: BinaryOp.uint(a.reg(0), ">>", Literal(16)),
```
simonlindholm: I'm guessing you'll want to fold `a << 2 << 8` into `a << 10`, and (I assume sh needs this) also multiplication being performed as a combination. To this end you could call `fold_mul_chains` on the constructed BinaryOp, with an optional parameter set to indicate that `allow_sll` is to be ignored within `fold`.

shll will want to use BinaryOp.int rather than BinaryOp.uint.

m2c/arch_sh.py
```diff
@@ -203,6 +209,19 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
                         if store is not None:
                             s.store_memory(store, a.reg_ref(0))
 
+                elif (
+                    args[1].writeback == Writeback.PRE and args[0] not in cls.saved_regs
+                ):
```
simonlindholm, line 215: Oh noo, not stack pointer being moved in the middle of functions... How does this work with six arguments? (Every SubroutineArg getting location 0 feels dubious.) Does it always restore the stack immediately afterwards?

m2c/arch_sh.py
```diff
@@ -259,6 +278,36 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
                 a.reg_ref(1), cls.instrs_arithmetic[mnemonic](a)
             )
+        elif mnemonic in cls.instrs_source_dest:
+            assert (
+                len(args) == 2
+                and isinstance(args[0], Register)
+                and isinstance(args[1], Register)
+            )
+            inputs = [args[0]]
+            if mnemonic in ("and", "or", "xor"):
```
simonlindholm: could these be moved to instrs_arithmetic, so that the dimension we split on is whether we read from dest register or not?

m2c/arch_sh.py
sozud: Ok added a few more tests and adjusted to use fold_mul_chains and int

m2c/arch_sh.py
sozud: It pushes the arguments in reverse order and does mov r14,r15 after the call. I added push_subroutine_arg to handle this and tests for 6,7,8 args

m2c/arch_sh.py
sozud: Ok moved and renamed to be more clear

m2c/evaluate.py
simonlindholm: allow_sll_chains here as well

tests/end_to_end/sh-shifts/sh2-gcc-o2-out.c
```diff
@@ -45,3 +45,39 @@ u32 test_rotl(u32 value) {
...
+s32 test_shal3(s32 value) {
+    return value * 8;
+}
+
+s32 test_shar2(s32 value) {
+    return (s32) (value >> 1) >> 1;
+}
+
+s32 test_shar3(s32 value) {
+    return (s32) ((s32) (value >> 1) >> 1) >> 1;
```
simonlindholm: ah yeah, should probably add some handling for folding this as well

tests/end_to_end/sh-shifts/sh2-gcc-o2-flags.txt
```diff
@@ -1 +1 @@
---context orig.c --target sh2-gcc-c --function test_shal --function test_shlr --function test_shar --function test_shll2 --function test_shlr2 --function test_shll8 --function test_shlr8 --function test_shll16 --function test_shlr16 --function test_rotl --function test_rotr
+--context orig.c --target sh2-gcc-c --function test_shal --function test_shlr --function test_shar --function test_shll2 --function test_shlr2 --function test_shll8 --function test_shlr8 --function test_shll16 --function test_shlr16 --function test_rotl --function test_rotr --function test_shll3 --function test_shll10 --function test_shll18 --function test_shal3 --function test_shar2 --function test_shar3 --function test_mul3 --function test_mul5 --function test_mul10
```
simonlindholm: -f instead of --function to make it shorter

tests/end_to_end/sh-shifts/orig.c
```diff
@@ -45,3 +45,39 @@ unsigned test_rotl(unsigned value) {
...
+    return value * 3;
+}
+
+signed test_mul5(signed value) {
+    return value * 5;
+}
+
+signed test_mul10(signed value) {
+    return value * 10;
+}
```
simonlindholm, line 83: There is an existing test that multiplies by all numbers up to 20 or something, you could add that one for sh. (It's possible that it does other things as well, in addition to multiplying by constants, but that's fine, the output will just contain M2C_ERROR markers for a while)

m2c/evaluate.py
sozud: Added with a manual test since I'm not sure gcc emits something that exercises this

tests/end_to_end/sh-shifts/sh2-gcc-o2-out.c
sozud: Ok added fold_shift_right

tests/end_to_end/sh-shifts/sh2-gcc-o2-flags.txt
sozud: Done

tests/end_to_end/sh-shifts/orig.c
sozud: Ok added, missing instructions for the moment

m2c/evaluate.py
simonlindholm: Ah, hm. I guess it just adds false positives then

PR #324, opened 2026-07-19
sh2: support multiplication and division

sozud:
Adds a couple more instructions and a SimpleAsmPattern for gcc division helpers


m2c/arch_sh.py
```diff
@@ -102,6 +102,39 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+class DivisionHelperPattern(SimpleAsmPattern):
+    pattern = make_pattern("mov.l _, $t", "jsr")
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        load, call = m.body
+        assert isinstance(load, Instruction) and isinstance(call, Instruction)
+        assert isinstance(load.args[0], AsmGlobalSymbol)
+        if not (
+            isinstance(call.args[0], AsmAddressMode)
+            and call.args[0].base == m.regs["t"]
```
simonlindholm: again make a comment in the pattern about how it's supposed to look (I think it may be nice to actually implement the thing I suggested)

m2c/arch_sh.py
```diff
@@ -454,6 +516,12 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
         ),
     }
 
+    instrs_multiply: InstrMap = {
+        "mul.l": lambda a: BinaryOp.int(a.reg(1), "*", a.reg(0)),
+        "muls": lambda a: BinaryOp.sint(a.reg(1), "*", a.reg(0)),
+        "mulu": lambda a: BinaryOp.uint(a.reg(1), "*", a.reg(0)),
```
simonlindholm: muls and mulu treat their inputs as 16-bit, so this should probably be something like
```py
BinaryOp.uint(as_u16(a.reg(1)), "*", as_u16(a.reg(0))
```
 (and we could implement normalization for the spellings muls.w, mulu.w)

m2c/arch_sh.py
```diff
@@ -364,6 +417,15 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
             function_target = target_reg
             has_delay_slot = True
             eval_fn = lambda s, a: s.make_function_call(a.regs[target_reg], outputs)
+        elif mnemonic in ("sdiv.fictive", "udiv.fictive"):
+            assert args == [Register("r4"), Register("r5")]
+            inputs = [Register("r4"), Register("r5")]
+            outputs = [Register("r0")]
+            has_delay_slot = True
```
simonlindholm: That's funny, a normal instruction with a delay slot. I don't think the delay slot rewriting code supports this at the moment; it assumes that all normal instructions with delay slots are function calls, and tests for them with `item.function_target is not None`. I think it should be possible to change that to `not item.is_jump()` (in the two places where that's checked).

sozud: You were right that this wasn't working, added another test and adjusted how the delay slots work. The issue is that we jsr to the helper so it does need a delay slot.

m2c/arch_sh.py
sozud: Ok added the casts and normalization

m2c/arch_sh.py
sozud: Ok changed the parsing as previously requested

m2c/asm_instruction.py
```diff
@@ -345,13 +353,7 @@ def expect(n: str) -> str:
         elif tok == "$" and arch.supports_dollar_regs:
             # Register.
             assert value is None
-            word = parse_word(arg_elems)
-            reg = word[1:]
-            if "$" in reg:
-                # If there is a second $ in the word, it's a symbol
-                value = AsmGlobalSymbol(word)
-            else:
-                value = asm_state.reg_formatter.parse_and_store(reg, arch)
```
simonlindholm, line 354: Eh, I'd revert this part, I think it makes sense for the core parser loop to still have this logic inlined here. Then either we could have `parse_arg_elem` do
```py
                    return parse_arg_elems(
                        arg_elems,
                        arch,
                        asm_state,
                        top_level=False,
                        precedence=MIN_PRECEDENCE,  # = 0
                    )
```

or we could add a simple function just for parsing sh registers (which wouldn't need the `"$" in reg` check since it's just for patterns that we control). Slight advantage of the latter is that it gives slightly more control over parsing and doesn't accidentally let through `@- r0` with a space.

m2c/flow_graph.py
```diff
@@ -547,10 +547,14 @@ def process_delay_slots(item: Union[Instruction, Label]) -> None:
             block_builder.new_block()
             block_builder.set_label(Label([temp_label]))
             block_builder.add_instruction(nop.clone())
-        elif item.function_target is not None:
+        elif not item.is_jump():
             # Move the delay slot instruction to before the call so it
             # passes correct arguments.
-            if len(item.args) >= 2 and item.args[1] in next_item.outputs:
+            if (
+                item.function_target is not None
```
simonlindholm: Hm, right, there's a subtle difference here, where with function calls, register arguments to the instruction are evaluated before the delay slot (which is why we cannot allow reordering here), while with the division intrinsics, the instruction *can* read register arguments from the delay slot. We should either mention this difference in a comment here, or alternatively (maybe a better idea, since it's more local and involves less thinking about this overcomplicated rewriting pass) change the asm pattern to additionally consume an instruction and do the reordering itself. It can be just

```py
pattern = make_pattern("mov.l _, $t", "jsr @$t", "*")
...
delay_slot = m.wildcard_items[0]
if not isinstance(delay_slot, Instruction):
    return None
return Replacement([delay_slot, new_ins], 3)
```

m2c/asm_instruction.py
sozud: Ok reverted that and added a helper for superh format

m2c/flow_graph.py
sozud: Ok went with your second suggestion

m2c/asm_instruction.py
```diff
@@ -335,6 +336,15 @@ def expect(n: str) -> str:
         assert g in n, f"Expected one of {list(n)}, got {g} (rest: {arg_elems})"
         return g
 
+    def parse_sh_register() -> Register:
+        word = parse_word(arg_elems)
+        if word.startswith("$"):
+            assert arch.supports_dollar_regs
```
```suggestion
        if word.startswith("$") and arch.supports_dollar_regs:
```

simonlindholm (approved):

PR #325, opened 2026-07-19
sh2: support direct register memory transfers

sozud:
This adds a number of `mov` instructions needed to manipulate pointers like: `*ptr`, `*--ptr`, `*ptr--` etc.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -338,14 +338,10 @@ class Sh2Arch(Arch):
             else:
                 assert isinstance(args[1], Register)
                 if isinstance(args[0], AsmAddressMode):
-                    if args[0].base == cls.stack_pointer_reg:
-                        # ex. mov.l @r15+, r14
-                        # we don't do the writeback because setup_initial_registers
-                        # sets sp to a GlobalSymbol that never gets reassigned.
-                        # stack accesses are constant offsets from it and there's
-                        # nowhere to += 4. The frame size comes from get_stack_info
-                        # which doesn't consult the epilogue
-                        assert args[0].writeback in (None, Writeback.POST)
+                    # We intentionally ignore args[0].writeback. It can only be non-None
+                    # for stack pointer writes (Sh2AddrModeWritebackPattern gets rid of
+                    # it for other registers), and then only during the epilogue, which
+                    # we don't emit any code for.
                     inputs = [args[0].base]
                 outputs = [args[1]]
                 is_load = True
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -140,18 +140,15 @@ class Sh2AddrModeWritebackPattern(AsmPattern):
                 f"for instruction: {instr}"
             )
 
+        assert addr.addend == AsmLiteral(0)
+
         new_args = list(instr.args)
-        new_args[addr_arg] = replace(addr, writeback=None, addend=AsmLiteral(0))
+        new_args[addr_arg] = replace(addr, writeback=None)
+        stride = Sh2Arch.mov_stride(instr.mnemonic)
         if addr.writeback == Writeback.PRE:
             return Replacement(
                 [
-                    AsmInstruction(
-                        "add",
-                        [
-                            AsmLiteral(-Sh2Arch.mov_stride(instr.mnemonic)),
-                            addr.base,
-                        ],
-                    ),
+                    AsmInstruction("add", [AsmLiteral(-stride), addr.base]),
                     AsmInstruction(instr.mnemonic, new_args),
                 ],
                 1,
@@ -160,10 +157,7 @@ class Sh2AddrModeWritebackPattern(AsmPattern):
         return Replacement(
             [
                 AsmInstruction(instr.mnemonic, new_args),
-                AsmInstruction(
-                    "add",
-                    [AsmLiteral(Sh2Arch.mov_stride(instr.mnemonic)), addr.base],
-                ),
+                AsmInstruction("add", [AsmLiteral(stride), addr.base]),
             ],
             1,
         )
@@ -246,16 +240,16 @@ class Sh2Arch(Arch):
         ]
     )
 
-    @classmethod
-    def mov_type(cls, mnemonic: str) -> Type:
+    @staticmethod
+    def mov_type(mnemonic: str) -> Type:
         return {
             "mov.b": Type.s8(),
             "mov.w": Type.s16(),
             "mov.l": Type.reg32(likely_float=False),
         }[mnemonic]
 
-    @classmethod
-    def mov_stride(cls, mnemonic: str) -> int:
+    @staticmethod
+    def mov_stride(mnemonic: str) -> int:
         return {
             "mov.b": 1,
             "mov.w": 2,
@@ -328,10 +322,8 @@ class Sh2Arch(Arch):
                         if store is not None:
                             s.store_memory(store, a.reg_ref(0))
 
-                elif (
-                    args[1].writeback == Writeback.PRE
-                    and args[1].base == cls.stack_pointer_reg
-                ):
+                elif args[1].base == cls.stack_pointer_reg:
+                    assert args[1].writeback == Writeback.PRE
 
                     def eval_fn(s: NodeState, a: InstrArgs) -> None:
                         source_raw = a.regs.get_raw(a.reg_ref(0))
@@ -339,11 +331,21 @@ class Sh2Arch(Arch):
                         if not s.stack_info.should_save(source_raw, None):
                             s.push_subroutine_arg(a.reg(0))
 
-                    assert args[1].base == cls.stack_pointer_reg
-                    assert args[1].writeback == Writeback.PRE
+                else:
+                    # writeback on other registers is rewritten into an explicit
+                    # add by Sh2AddrModeWritebackPattern, so no eval_fn is needed.
+                    pass
             else:
                 assert isinstance(args[1], Register)
                 if isinstance(args[0], AsmAddressMode):
+                    if args[0].base == cls.stack_pointer_reg:
+                        # ex. mov.l @r15+, r14
+                        # we don't do the writeback because setup_initial_registers
+                        # sets sp to a GlobalSymbol that never gets reassigned.
+                        # stack accesses are constant offsets from it and there's
+                        # nowhere to += 4. The frame size comes from get_stack_info
+                        # which doesn't consult the epilogue
+                        assert args[0].writeback in (None, Writeback.POST)
                     inputs = [args[0].base]
                 outputs = [args[1]]
                 is_load = True
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -18,7 +18,14 @@ from .asm_instruction import (
     Writeback,
     get_jump_target,
 )
-from .asm_pattern import AsmMatch, Replacement, SimpleAsmPattern, make_pattern
+from .asm_pattern import (
+    AsmMatch,
+    AsmMatcher,
+    AsmPattern,
+    Replacement,
+    SimpleAsmPattern,
+    make_pattern,
+)
 from .instruction import (
     Instruction,
     InstructionMeta,
@@ -38,7 +45,6 @@ from .translate import (
     InstrArgs,
     Literal,
     NodeState,
-    StoreStmt,
     UnaryOp,
     as_s16,
     as_type,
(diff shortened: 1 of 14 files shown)


m2c/arch_sh.py
```diff
@@ -233,7 +232,14 @@ def parse(
...
                 assert isinstance(args[0], AsmLiteral)
                 eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), Literal(a.imm_value(0)))
-        elif mnemonic in ("mov.l", "mov.w"):
+        elif mnemonic == "mova":
+            assert (
+                len(args) == 2
+                and isinstance(args[0], AsmGlobalSymbol)
+                and isinstance(args[1], Register)
+            )
+            outputs = [args[1]]
```
simonlindholm: We ought to add an eval_fn that just shows an error. (Or does something more sane)

m2c/arch_sh.py
```diff
@@ -242,17 +248,46 @@ def parse(
...
+                        }[mnemonic]
+                        store = make_store(a, store_type)
                         if store is not None:
                             s.store_memory(store, a.reg_ref(0))
 
                 elif (
-                    args[1].writeback == Writeback.PRE and args[0] not in cls.saved_regs
+                    args[1].writeback == Writeback.PRE
+                    and args[1].base == cls.stack_pointer_reg
+                    and args[0] not in cls.saved_regs
```
simonlindholm: Failed to notice this before, but this heuristic is definitely imperfect, saved registers can also hold arguments. `make_store` uses `stack_info.should_save()` to check this, which bases its result on whether the register still contains a "saved_reg_r8" or similar placeholder from function entry; I think we could do the same here.

m2c/arch_sh.py
```diff
@@ -242,17 +248,46 @@ def parse(
...
                 ):
 
                     def eval_fn(s: NodeState, a: InstrArgs) -> None:
                         s.push_subroutine_arg(a.reg(0))
 
+                elif (
+                    args[1].writeback == Writeback.PRE
+                    and args[1].base != cls.stack_pointer_reg
+                ):
```
simonlindholm: An alternative to this which might be nicer (doesn't duplicate as much code) is to do like arm and use an asm pattern to rewrite writebacks into an add + a non-writeback instruction (AddrModeWritebackPattern).

m2c/arch_sh.py
```diff
@@ -242,17 +248,46 @@ def parse(
...
+                        size = {"mov.b": 1, "mov.w": 2, "mov.l": 4}[mnemonic]
+                        s.set_reg(
+                            writeback_base,
+                            BinaryOp.intptr(a.regs[writeback_base], "-", Literal(size)),
+                        )
+                        store_type = {
+                            "mov.b": Type.s8(),
+                            "mov.w": Type.s16(),
+                            "mov.l": Type.reg32(likely_float=False),
+                        }[mnemonic]
```
simonlindholm: Should probably have a function for resolving "b"/"w"/"l" to a type

m2c/arch_sh.py
sozud: removed, came back from a bad rebase

m2c/arch_sh.py
sozud: Ok added a test and reworked it with should_save

m2c/arch_sh.py

m2c/arch_sh.py
sozud: I think the issue here is that `@-r15` is a special case so preprocessing could be error-prone?

simonlindholm: Yeah, with current stack pointer setup it makes sense to not do it for `@-r15` and `@r15+` specifically. I was thinking for all other regs. (Arm similarly doesn't do it for pre/post-increments of sp, instead it rewrites those into `push`/`pop`.)

m2c/arch_sh.py
```diff
@@ -242,16 +249,41 @@ def parse(
...
+                    args[1].writeback == Writeback.PRE
+                    and args[1].base == cls.stack_pointer_reg
                 ):
 
                     def eval_fn(s: NodeState, a: InstrArgs) -> None:
-                        s.push_subroutine_arg(a.reg(0))
+                        source_raw = a.regs.get_raw(a.reg_ref(0))
+                        assert source_raw is not None
+                        if not s.stack_info.should_save(
+                            source_raw, a.memory_ref(1).offset
```
simonlindholm: FWIW I think the second argument we pass here doesn't matter -- it's only used if `isinstance(expr, PassedInArg) and expr.loc.offset is not None` which is only the case for MIPS with home-space registers. `a.memory_ref(1).offset` feels wrong to me wrong though, would rather go with `-4` or `None` or something.

m2c/arch_sh.py
sozud: Do you mean like this? I'm probably not understanding since this seems worse than the current version https://github.com/matt-kempster/m2c/commit/535befbc9c0a433c4d19f0a8fd164f43e0f03e5a

m2c/arch_sh.py
sozud: Tried changing this to `None` and I think we can't treat everything as savable?
```
   s8 test_loadb_predec(s8 **ptr) {
       s8 *temp_r1;
  -    *ptr = temp_r1;
  +    temp_r1 = *ptr;
  +    *ptr = temp_r1 + 1;
       return *temp_r1;
   }

   s16 test_loadw_predec(s16 **ptr) {
       s16 *temp_r1;

  -    temp_r1 = *ptr - 2;
  -    *ptr = temp_r1;
  +    temp_r1 = *ptr;
  +    *ptr = temp_r1 + 2;
       return *temp_r1;
   }
```

m2c/arch_sh.py
simonlindholm: Yes, except:
- do it for both pre and post indexing
- remove the `if isinstance(args[0], AsmAddressMode) and args[0].writeback is not None and args[0].base != cls.stack_pointer_reg and args[0].writeback == Writeback.PRE:` case
- and the `elif args[1].writeback == Writeback.PRE and args[1].base != cls.stack_pointer_reg:` one

and the rewriting could also apply to `STS.L`/`STC.L`, although I'm not sure you care about that. Admittedly, if you don't, the win is not as large as with arm, where almost any instruction can use addressing modes with writeback, and it may not be worth it (but if so it would still be worth grouping the various `mov` cases into a single `eval_fn` so you don't duplicate writeback handling).

An `add` instruction will also be able to lower to something smarter than `BinaryOp.intptr(a.regs[writeback_base], "+", Literal(size))`, though that could be done with your current version too if you just call the right function.

Behavior of special cases like `mov.l @-r0, r0` may dictate which version becomes simplest.

sozud: Ok so if I apply the suggestion I get output like this:
```
 void test_storel_predec(s32 **ptr, s32 value) {
     s32 *temp_r1;
 
-    temp_r1 = *ptr - 4;
-    *temp_r1 = value;
-    *ptr = temp_r1;
+    temp_r1 = *ptr;
+    temp_r1->unk-4 = value;
+    *ptr = temp_r1 - 4;
 }
 
 s8 test_loadb_predec(s8 **ptr) {
     s8 *temp_r1;
 
-    temp_r1 = *ptr - 1;
-    *ptr = temp_r1;
-    return *temp_r1;
+    temp_r1 = *ptr;
+    *ptr = temp_r1 - 1;
+    return temp_r1->unk-1;
 }
 
 s16 test_loadw_predec(s16 **ptr) {
     s16 *temp_r1;
 
-    temp_r1 = *ptr - 2;
-    *ptr = temp_r1;
-    return *temp_r1;
+    temp_r1 = *ptr;
+    *ptr = temp_r1 - 2;
+    return temp_r1->unk-2;
 }
```
Is that syntax preferable in m2c?

simonlindholm: That looks worse, certainly, but it also seems kinda accidental. Apparently the difference is due to BinaryOp(..., "-", Literal(4)) (with current code) vs BinaryOp(..., "+", Literal(-4)) (with the asm rewrite). The latter feels more correct, we ought to always normalize to that form.

I pushed a commit to master that should avoid generating the ->unk-4's.

m2c/arch_sh.py
simonlindholm: I'm confused, I tried changing the argument to None and I'm not seeing a difference in output (and I still feel like there shouldn't be one, reading the code). I think your diff might come from another change?

sozud: Ok not sure what changed but seems to be fine

m2c/arch_sh.py
sozud: Ok added AddrModeWritebackPattern now

m2c/arch_sh.py
```diff
@@ -104,6 +110,65 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+
+        new_args = list(instr.args)
+        new_args[addr_arg] = replace(addr, writeback=None, addend=AsmLiteral(0))
+        if addr.writeback == Writeback.PRE:
+            return Replacement(
+                [
+                    AsmInstruction(
+                        "add",
+                        [
+                            AsmLiteral(-Sh2Arch.mov_stride(instr.mnemonic)),
```
simonlindholm: use a temp for `Sh2Arch.mov_stride(instr.mnemonic)` so it formats nicer

m2c/arch_sh.py
```diff
@@ -181,6 +246,22 @@ def c_symbol_name(self, asm_name: str) -> str:
         ]
     )
 
+    @classmethod
```
simonlindholm: `staticmethod` so you don't need to have the `cls` parameter

m2c/arch_sh.py
```diff
@@ -242,19 +323,22 @@ def parse(
...
+                        store_type = cls.mov_type(mnemonic)
+                        store = make_store(a, store_type)
                         if store is not None:
                             s.store_memory(store, a.reg_ref(0))
 
                 elif (
-                    args[1].writeback == Writeback.PRE and args[0] not in cls.saved_regs
+                    args[1].writeback == Writeback.PRE
+                    and args[1].base == cls.stack_pointer_reg
                 ):
```
```py
else:
    assert args[1].writeback == Writeback.PRE
    assert args[1].base == cls.stack_pointer_reg
    ...
```
and remove the redundant asserts below the `eval_fn`

m2c/arch_sh.py
```diff
@@ -263,22 +347,14 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
                     inputs = [args[0].base]
```
simonlindholm, line 345: assert that this is either no writeback, or that it's a read from sp (in which case I guess you're ignoring this because it's part of the epilogue and doesn't matter anyway?)

m2c/arch_sh.py
```diff
@@ -104,6 +110,65 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+            return None
+
+        if addr.base in (instr.args[0], instr.args[1]):
+            raise DecompFailure(
+                "Writeback with base register also used as load/store value, "
+                f"for instruction: {instr}"
+            )
+
+        new_args = list(instr.args)
+        new_args[addr_arg] = replace(addr, writeback=None, addend=AsmLiteral(0))
```
simonlindholm: the addend is already 0, right? could be asserted?

m2c/arch_sh.py
sozud: Ok added an assert and an explanation about the epilogue

m2c/arch_sh.py
sozud: ok asserting this

m2c/arch_sh.py
sozud: ok done

m2c/arch_sh.py
sozud: done

m2c/arch_sh.py
sozud: done

m2c/arch_sh.py
simonlindholm: Hm, right, we can't assert no writeback directly, because we need to support the pre-rewrite syntax. I guess what I asked you to do was kinda useless then, and if statement becomes kinda weird when all it does it `assert args[0].writeback in (None, Writeback.POST)` which will be true even outside the if.

m2c/arch_sh.py
```diff
@@ -242,43 +317,46 @@ def parse(
...
                 assert isinstance(args[1], Register)
                 if isinstance(args[0], AsmAddressMode):
+                    if args[0].base == cls.stack_pointer_reg:
+                        # ex. mov.l @r15+, r14
+                        # we don't do the writeback because setup_initial_registers
+                        # sets sp to a GlobalSymbol that never gets reassigned.
+                        # stack accesses are constant offsets from it and there's
+                        # nowhere to += 4. The frame size comes from get_stack_info
+                        # which doesn't consult the epilogue
+                        assert args[0].writeback in (None, Writeback.POST)
```
```suggestion
                    # We intentionally ignore args[0].writeback. It can only be non-None
                    # for stack pointer writes (Sh2AddrModeWritebackPattern gets rid of
                    # it for other registers), and then only during the epilogue, which
                    # we don't emit any code for.
```

PR #328, opened 2026-07-22
sh2: support immediate logical operations

sozud:
Support for operations like:
```
unsigned test(unsigned value) {
    return value & 0x5A;
}

unsigned test_or(unsigned value) {
    return value | 0x5A;
}

unsigned test_xor(unsigned value) {
    return value ^ 0x5A;
}

int test_tst(unsigned value) {
    return (value & 0x5A) == 0;
}
```


simonlindholm:
I note that test_tst compiled to
```
	and	#90,r0
	tst	r0,r0
```
and there was no handling added for the tst instruction, I don't know if that was intentional

PR #330, opened 2026-07-22
sh2: recognize modulo helper sequences

sozud:
This adds `ShModuloPattern` to help with
```
signed test(signed lhs, signed rhs) {
    return lhs % rhs;
}

unsigned test_unsigned(unsigned lhs, unsigned rhs) {
    return lhs % rhs;
}
```
and recognize calls to `___sdivsi3` and `___udivsi3`

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -23,6 +23,7 @@ from .asm_pattern import (
     AsmMatcher,
     AsmPattern,
     Replacement,
+    ReplacementPart,
     SimpleAsmPattern,
     make_pattern,
 )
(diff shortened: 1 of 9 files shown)


simonlindholm (commented):
Can you use a SimpleAsmPattern?

sozud:
> Can you use a SimpleAsmPattern?

Ok done

m2c/arch_sh.py
```diff
@@ -195,6 +196,44 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+        "sub $m, $l",
+        "rts?",
+        "mov $l, $r0",
+    )
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        # Both helpers return the quotient in r0, which the sequence multiplies back
+        # and subtracts to recover the remainder. Testing showed gcc kept the dividend
+        # in a callee-saved register only for the signed helper, which is what we key
+        # on to tell the two apart.
```
simonlindholm: I feel like this has got to be incidental. You can look at the name of the called function, right?

m2c/arch_sh.py
```diff
@@ -195,6 +196,44 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+        "mov $a, $l",
+        "mov.l _, $t",
+        "jsr @$t",
+        "mov $b, $r",
+        "mul.l $r, $r0",
+        "lds.l @$r15+, $pr",
+        "mov.l @$r15+, $r14",
+        "sts $macl, $m",
+        "sub $m, $l",
+        "rts?",
```
simonlindholm: These optional instructions are a warning sign, I learned; they usually mean the assembler used by the compiler can perform minor reorderings and easily break the pattern. I imagine a branch or other delay slot instruction could also go in this slot, or a maybe even an arbitrary instruction. I suggest doing some grepping over a corpus and see if you can actually confirm it to be reliable. IDO has similar issues with unreliable patterns, and for more complex things like modulo I think we detect it in evaluate.py instead, folding `a - a / b * b` into `a % b`

m2c/arch_sh.py
```diff
@@ -195,6 +196,44 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
 
 
+class ShModuloPattern(SimpleAsmPattern):
+    pattern = make_pattern(
+        "mov $a, $l",
+        "mov.l _, $t",
+        "jsr @$t",
+        "mov $b, $r",
+        "mul.l $r, $r0",
+        "lds.l @$r15+, $pr",
```
simonlindholm: Seems very weird for this to be part of the pattern. Isn't this part of the function epilogue reordered up here?

m2c/arch_sh.py
sozud: Alright added literal_pool_target to look up the name

m2c/arch_sh.py
sozud: Alright I went through sotn-decomp and added several additional tests that don't follow this pattern. The new pattern is
```

    patterns = [
        make_pattern("mov.l _, $t", "jsr @$t", "*"),
        make_pattern("mov.l _, $t", "*", "jsr @$t", "*"),
        make_pattern("mov.l _, $t", "*", "*", "jsr @$t", "*"),
```

m2c/arch_sh.py
sozud: Removed from the pattern

m2c/arch_sh.py
```diff
@@ -165,34 +167,57 @@ def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
...
+class DivisionHelperPattern(AsmPattern):
+    # gcc normally emits the literal pool load right before the call, but the
+    # scheduler is free to slot unrelated instructions into the gap, so allow a
+    # couple. Matching is still safe because the pool entry has to resolve to a
+    # known division helper.
+    patterns = [
+        make_pattern("mov.l _, $t", "jsr @$t", "*"),
+        make_pattern("mov.l _, $t", "*", "jsr @$t", "*"),
+        make_pattern("mov.l _, $t", "*", "*", "jsr @$t", "*"),
+    ]
```
simonlindholm, line 202: Hm, this is rather unfortunate. It would be easy to imagine more than two instructions appearing here, maybe the mov moving to a different basic block, and false positives from branch instructions and instructions overwriting `$t` (even though it's probably not super likely in practice). I wonder if we should just move this detection to the function call `eval_fn` instead.

m2c/arch_sh.py
```diff
@@ -165,34 +167,57 @@ def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
...
+        return None
+    return target.as_symbol_without_addend()
+
+
+def sh_sub(lhs: Expression, rhs: Expression) -> Expression:
+    """Subtract, recognizing the `x - (x / y) * y` shape gcc emits for modulo."""
+    expr = handle_sub(lhs, rhs)
+    if isinstance(expr, BinaryOp):
+        return fold_divmod(expr)
+    return expr
```
simonlindholm, line 190: I think it would be fine to move fold_divmod into handle_sub, to get this fix for other architectures as well

tests/end_to_end/sh-modulo-helpers/manual.s
```diff
@@ -0,0 +1,177 @@
+	.file	"manual.s"
+	.text
+	.align 2
+
+! The quotient is used straight out of r0, with no intervening move, and the
+! remainder feeds an add rather than being returned.
```
simonlindholm, line 6: This reads AI-generated. Is it? (And have other parts of the PR series been, too?) I would really have appreciated up-front disclosure if so.

tests/end_to_end/sh-modulo-helpers/manual.s
```diff
@@ -0,0 +1,177 @@
...
+	sub	r1,r0
+	mov.l	@r15+,r13
+	mov.l	@r15+,r11
+	mov.l	@r15+,r9
+	lds.l	@r15+,pr
+	rts
+	nop
+	.align 2
+L105:
+	.long	___udivsi3
```
simonlindholm, line 177: I feel like these tests aren't meaningful, given the solution settled on no longer cares about exact asm scheduling

tests/end_to_end/sh-modulo-helpers/manual.s
sozud: Yes I used AI assistance with this series. My understanding of this domain is low and I didn't take enough responsibility for the quality here, which put a burden on you. I'm sorry. I'll close this. Let me know if you want me to put up a revert PR for any of the other commits.

simonlindholm: No, the other PRs have been fine, I wouldn't have merged them otherwise (and I appreciate you sending them one by one, unlike the mastodon x86 PR which just got closed without anything to show for it).

PR #331, opened 2026-07-29
sh2: support positive comparisons

sozud:
This adds the `cmp/pl` and `cmp/pz` instructions which doc`If Rn > 0, 1 → T` and `If Rn ≥ 0, 1 → T` respectively. This allows some additional types of branching like
```
signed test(signed value) {
    if (value < 0) {
        return -1;
    }
    if (value > 0) {
        return 1;
    }
    return 0;
}
```
Disclosure: AI assistance was used while developing and reviewing this change. I verified the implementation and tests myself.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -510,6 +510,18 @@ class Sh2Arch(Arch):
                     BinaryOp.icmp(value, "==", Literal(0)),
                 )
 
+        elif mnemonic in ("cmp/pl", "cmp/pz"):
+            assert len(args) == 1 and isinstance(args[0], Register)
+            inputs = [args[0]]
+            outputs = [Register("condition_bit")]
+            eval_fn = lambda s, a: s.set_reg(
+                Register("condition_bit"),
+                BinaryOp.scmp(
+                    a.reg(0),
+                    ">" if mnemonic == "cmp/pl" else ">=",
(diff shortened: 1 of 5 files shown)


m2c/arch_sh.py
```diff
@@ -510,6 +510,18 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+            inputs = [args[0]]
+            outputs = [Register("condition_bit")]
+            eval_fn = lambda s, a: s.set_reg(
+                Register("condition_bit"),
+                BinaryOp.scmp(
+                    a.reg(0),
+                    ">" if mnemonic == "cmp/pl" else ">=",
+                    Literal(0),
+                ),
+            )
```
simonlindholm, line 519: Could you move this + the other `cmp/` comparisons into an `InstrMap` table for compactness?

sozud: Ok done. There's two tables since one uses a single register vs. the other uses two registers

simonlindholm: Eh, I think they could use a single table. No need to assert that len == 1 and so on beforehand, you can just do
```
inputs = [arg in inst.args if isinstance(arg, Register)]
```

sozud: Applied

PR #332, opened 2026-07-29
sh2: sign-extend mov immediates

sozud:
This implements `MOV #imm,Rn`, which does `imm → Sign extension → Rn`.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -311,7 +311,13 @@ class Sh2Arch(Arch):
                 eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), a.reg(0))
             else:
                 assert isinstance(args[0], AsmLiteral)
-                eval_fn = lambda s, a: s.set_reg(a.reg_ref(1), Literal(a.imm_value(0)))
+
+                def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                    value = a.imm_value(0)
+                    if 0x80 <= value <= 0xFF:
+                        value -= 0x100
+                    s.set_reg(a.reg_ref(1), Literal(value))
+
         elif mnemonic in ("mov.b", "mov.l", "mov.w"):
             assert len(args) == 2
             if isinstance(args[0], Register):
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -631,8 +631,11 @@ class Sh2Arch(Arch):
             handle_add(replace(a, raw_args=[a.raw_arg(1), a.raw_arg(1), a.raw_arg(0)]))
             if isinstance(a.raw_arg(0), Register)
             else handle_addi(
-                replace(a, raw_args=[a.raw_arg(1), a.raw_arg(1), a.raw_arg(0)]),
-                imm=a.s8_imm(0),
+                a.reg_ref(1),
+                a.reg_ref(1),
+                a.reg(1),
+                a.s8_imm(0),
+                a,
             )
         ),
         "sub": lambda a: handle_sub(a.reg(1), a.reg(0)),
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -630,13 +630,7 @@ class Sh2Arch(Arch):
         "add": lambda a: (
             handle_add(replace(a, raw_args=[a.raw_arg(1), a.raw_arg(1), a.raw_arg(0)]))
             if isinstance(a.raw_arg(0), Register)
-            else handle_addi(
-                a.reg_ref(1),
-                a.reg_ref(1),
-                a.reg(1),
-                a.s8_imm(0),
-                a,
-            )
+            else handle_addi(a.reg_ref(1), a.reg_ref(1), a.reg(1), a.s8_imm(0), a)
         ),
         "sub": lambda a: handle_sub(a.reg(1), a.reg(0)),
         "and": lambda a: BinaryOp.int(a.reg(1), "&", a.reg_or_imm(0)),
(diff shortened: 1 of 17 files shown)


simonlindholm:
Can you add a function `s8_imm` on InstrArgs, just like the existing s16_imm/u16_imm? Also, doesn't this apply to all instructions that use immediates, not just mov?

sozud:
> Can you add a function s8_imm on InstrArgs, just like the existing s16_imm/u16_imm? Also, doesn't this apply to all instructions that use immediates, not just mov?

I added s8_imm. I checked and the logical immediates don't sign extend, but cmp/eq and addi do, so I fixed those and reworked the tests.

m2c/arch_sh.py
simonlindholm: These other cmp/ instructions ought to be `a.reg(0)`, I guess

m2c/arch_sh.py
```diff
@@ -616,7 +617,11 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
     instrs_compare: InstrMap = {
         "cmp/pl": lambda a: BinaryOp.scmp(a.reg(0), ">", Literal(0)),
         "cmp/pz": lambda a: BinaryOp.scmp(a.reg(0), ">=", Literal(0)),
-        "cmp/eq": lambda a: BinaryOp.icmp(a.reg(1), "==", a.reg_or_imm(0)),
+        "cmp/eq": lambda a: BinaryOp.icmp(
+            a.reg(1),
+            "==",
+            a.reg(0) if isinstance(a.raw_arg(0), Register) else a.s8_imm(0),
```
simonlindholm: should probably add a `reg_or_s8_imm` function

m2c/evaluate.py
```diff
@@ -313,7 +313,11 @@ def handle_xori(args: InstrArgs) -> Expression:
     return BinaryOp.int(left=left, op="^", right=right)
 
 
-def handle_addi(args: InstrArgs, arm: bool = False) -> Expression:
+def handle_addi(
+    args: InstrArgs,
+    arm: bool = False,
+    imm: Optional[Expression] = None,
+) -> Expression:
```
simonlindholm, line 340: Ugh, this (and the caller) is growing into a sprawling mess. It made sense to pass just `args` here when it was only MIPS and PPC, but now with ARM and SH I'm starting to feel like we should just merge `handle_add` and `handle_addi`, and pass in all of output_reg, source_reg, source, imm explicitly... Doing that merge may be going too far for this PR, but at least making parameters explicit for `handle_addi` would be good. And instead of `args` we ought to pass in `stack_info`.

m2c/arch_sh.py
sozud: Added

m2c/arch_sh.py

m2c/evaluate.py
sozud: I tried this and I think explicit parameters and stack_info are a large enough change that it ought to be a separate PR. What do you think? https://github.com/matt-kempster/m2c/commit/ccb028be38dba62f1e3ab9dd862a72054ffde88f

simonlindholm: Hm, maybe. I forgot about `args.instruction_ref` -- can you revert back to passing the InstructionArgs, which should also make the diff slightly smaller?

sozud: Ok applied the smaller version to this PR

m2c/evaluate.py
```diff
@@ -313,18 +314,20 @@ def handle_xori(args: InstrArgs) -> Expression:
...
+def handle_addi(
+    output_reg: Register,
+    source_reg: Register,
+    source: Expression,
+    imm: Expression,
+    args: InstrArgs,
+    *,
+    arm: bool = False,
+) -> Expression:
+    ref = args.maybe_gprel_imm(2) if len(args.raw_args) > 2 else None
```
simonlindholm: Hm, this is awful, can we split out `def handle_addi_mips(args: InstrArgs) -> Expression:` and put this just there?

m2c/evaluate.py
simonlindholm, line 354: We can define
```py
arm = (args.stack_info.global_info.platform.arch == ArchEnum.ARM)
```
so we don't have to pass in the `arm` argument

m2c/arch_sh.py
```diff
@@ -616,19 +617,27 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
 
     instrs_read_modify_write: InstrMap = {
         # sh2 format is src, dst
         # add handler is dest, left, right
         "add": lambda a: (
-            handle_add if isinstance(a.raw_arg(0), Register) else handle_addi
-        )(replace(a, raw_args=[a.raw_arg(1), a.raw_arg(1), a.raw_arg(0)])),
+            handle_add(replace(a, raw_args=[a.raw_arg(1), a.raw_arg(1), a.raw_arg(0)]))
+            if isinstance(a.raw_arg(0), Register)
+            else handle_addi(
```
simonlindholm: should fit on one line

sozud: Black automatically reformats it to multiline if I try to change it

m2c/evaluate.py
sozud: Done

m2c/evaluate.py
sozud: Done

m2c/arch_sh.py
simonlindholm: even if you remove the trailing comma? (which is otherwise a cue to `black` to keep things split up)

sozud: Ok removing the comment puts it on one line

simonlindholm: nice, sorry for the nitpick

PR #333, opened 2026-07-29
sh2: evaluate mova addresses

sozud:
Implements `MOVA @(disp,PC),R0`. I don't think my version of gcc can emit this outside of a jump table so this has a manual test.

### DIFF
--- a/tests/end_to_end/sh-mova/context.c
+++ b/tests/end_to_end/sh-mova/context.c
@@ -1,0 +1,1 @@
+void *test(void);
--- a/tests/end_to_end/sh-mova/context.c
+++ b/tests/end_to_end/sh-mova/context.c
@@ -1,1 +1,0 @@
-void *test(void);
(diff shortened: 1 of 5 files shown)


tests/end_to_end/sh-mova/context.c
```diff
@@ -0,0 +1 @@
+void *test(void);
```
simonlindholm: This context file is not meaningful (I think this goes for most e2e test sh context.c files tbh, and for the ones that matter, --context orig.c can be used instead)

sozud: Ok removed. Removing the context gives this output
```
+? (*test(void))() {
     return literal;
 }
 ```
 so I adjusted the asm to give a normal looking function

simonlindholm: Semi-broken-looking test output would also be fine if you wanted to leave it, it just means we can look at test outputs to get inspiration for what to fix, and we get to be happy when we accidentally fix it in the future (by not treating `literal` as a function).

PR #336, opened 2026-07-30
sh2: support double-width multiply

sozud:
Supports dmuls.l/dmulu.l. Fixes these cases

```
 s32 test(s32 lhs, s32 rhs) {
-    return (s32) (((s64) lhs * (s64) rhs) >> 0x20);
+    M2C_ERROR(/* unknown instruction: dmuls.l $r5, $r4 */);
+    return M2C_ERROR(/* Read from unset register $mach */);
 }
 
 u32 test_unsigned(u32 lhs, u32 rhs) {
-    return (u32) (((u64) lhs * (u64) rhs) >> 0x20);
+    M2C_ERROR(/* unknown instruction: dmulu.l $r5, $r4 */);
+    return M2C_ERROR(/* Read from unset register $mach */);
 }
 ```

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -55,6 +55,7 @@ from .translate import (
 from .evaluate import (
     condition_from_expr,
     fn_op,
+    fold_divmod,
     fold_mul_chains,
     fold_shift_right,
     handle_add,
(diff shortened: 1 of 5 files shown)


simonlindholm:
I think you want to implement these like MIPS (with MULT_HI/MULTU_HI + `fold_divmod`); it looks like 100% of the uses in sotn are part of division patterns

sozud:
> I think you want to implement these like MIPS (with MULT_HI/MULTU_HI + fold_divmod); it looks like 100% of the uses in sotn are part of division patterns

Ok I adjusted this to work more like MIPS. I didn't go deep into the folding since there's a lot of complexity in handling the various patterns

simonlindholm:
> Ok I adjusted this to work more like MIPS. I didn't go deep into the folding since there's a lot of complexity in handling the various patterns

Makes sense, this is something we also haven't got around to for all existing arch/compiler combinations. Definitely worth adding sh2 versions of the existing division tests though just so we can track the current output state.

PR #340, opened 2026-08-02
sh2: handle doubled long-branch jump tables

sozud:
This adds another jump table handler for tables where the index is doubled. In isolation we could get rid of "tablejmp.doubled.fictive" by removing the `add r1,r1` but other doubled table types require the doubled index to be available. I'm intending to add these in a follow-up.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -51,6 +51,7 @@ from .translate import (
     as_type,
     as_u16,
     as_u32,
+    early_unwrap,
 )
 
 from .evaluate import (
@@ -76,8 +77,8 @@ from .evaluate import (
 from .types import FunctionSignature, Type
 
 
-def jump_table_targets(m: AsmMatch, mova_index: int) -> Optional[List[AsmGlobalSymbol]]:
-    mova = m.body[mova_index]
+def jump_table_targets(m: AsmMatch) -> Optional[List[AsmGlobalSymbol]]:
+    mova = m.body[0]
     assert isinstance(mova, Instruction)
     assert isinstance(mova.args[0], AsmGlobalSymbol)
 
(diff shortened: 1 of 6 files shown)


simonlindholm:
This feels wrong to me, we're undoing a multiplication that we introduced ourselves (and only for delay slot purposes, making it kind of a hack in the first place). Can you show some examples of (or wip m2c commits for) the other doubled table types you're referring to?

sozud:
Ok this is the example https://github.com/matt-kempster/m2c/commit/8a8383fbf7df9efd14d88b497f0369a137e48a59
```

	add	r2,r2               ; double the index
	mov	r2,r0            
	mov	#3,r1               
	cmp/hi	r1,r5
	bt.s	L8
	mov.w	r6,@(r0,r4)     ; doubled index gets used here
	mov	r2,r1               ; copy index for the jump
	mova	L13,r0
	mov.w	@(r0,r1),r1   
	add	r1,r0
	jmp        @r0          ; jump
```
So I don't think we can remove the add instruction locally

simonlindholm:
I see, that's annoying. Feels like we could have just one pattern
```
        "mova _, $b",
        "mov.w @($b,$i),$i",
        "add $i, $b",
        "jmp @$b",
```
that gets replaced by `tablejmp.doubled.fictive`, and then we can either use IR patterns to convert that into `tablejmp.fictive`, or `early_unwrap` in the handler for `tablejmp.doubled.fictive` to check if `$i` is `something * 2`.

sozud:
Ok I applied your suggestion

m2c/arch_sh.py
```diff
@@ -76,10 +77,31 @@
...
+        if (
+            not isinstance(entry, AsmSymbolicData)
+            or not isinstance(entry.data, BinOp)
+            or entry.data.op != "-"
+            or not isinstance(entry.data.lhs, AsmGlobalSymbol)
+            or entry.data.rhs != AsmGlobalSymbol(table_name)
+        ):
+            return None
+        targets.append(entry.data.lhs)
+    return targets or None
```
simonlindholm: Move this function's contents back into its caller again, it doesn't make much sense outside of it

PR #341, opened 2026-08-03
sh2: handle jump tables split by literal pools

sozud:
This adds handling for cases like:
```
  mova table, r0
  mov.w @(r0, index), index  // load table offset
  add index, r0             // calculate target
  bra dispatch              // skip literal pool
  nop
  ... literal pool ...
  dispatch:
  jmp @r0
  nop
  ```

### CODE REVIEW

File: m2c/arch_sh.py [hunk 1]
```diff
@@ -89,29 +110,40 @@ class JumpTablePattern(SimpleAsmPattern):
...
+                AsmInstruction("nop", []),
+            ],
+            len(m.body),
+        )
+
+
+class BranchedDoubledJumpTablePattern(SimpleAsmPattern):
+
+    pattern = make_pattern(
+        "add $i, $i",
```
simonlindholm: Remove this part of the pattern, and the `mova_index` arg to `jump_table_targets`

simonlindholm (commented):
Hm, given that the replacement code is identical, and `jump_table_targets` doesn't make sense outside of the functions, it would make sense to do the same as e.g. DivP2Pattern2 in arch_mips.py with multiple patterns in a single non-`Simple` `AsmPattern`.

sozud:
> Hm, given that the replacement code is identical, and jump_table_targets doesn't make sense outside of the functions, it would make sense to do the same as e.g. DivP2Pattern2 in arch_mips.py with multiple patterns in a single non-Simple AsmPattern.

Ok made it like DivP2Pattern2

PR #342, opened 2026-08-03
sh2: recover GCC far intra-function jumps

sozud:
This adds an AsmPattern for jumps that go beyond the normal branch range. This was found in the SOTN corpus.


m2c/arch_sh.py
```diff
@@ -79,6 +79,51 @@
...
+        "jmp @$t",
+        "mov.l @$s+,$t",
+    )
+
+    def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
+        m = matcher.try_match(self.pattern)
+        if m is None:
+            return None
+        if m.regs["s"] != Register("r15"):
+            return None
```
simonlindholm: You should be able to write r15 within the pattern

m2c/arch_sh.py
```diff
@@ -79,6 +79,51 @@
...
+        load = m.body[1]
+        assert isinstance(load, Instruction)
+        if not isinstance(load.args[0], AsmGlobalSymbol):
+            return None
+        entry = m.asm_data.values.get(load.args[0].symbol_name)
+        if entry is None:
+            return None
+        target = entry.data_at_offset(0, 4)
+        if isinstance(target, bytes):
+            target_name = f".L{int.from_bytes(target, 'big'):08X}"
```
simonlindholm: This does not belong here, but rather in your disassembler.

tests/end_to_end/sh-far-jump/orig.c
```diff
@@ -0,0 +1,571 @@
+extern int sink(int);
+
+int test(int value) {
```
simonlindholm: This is a very heavy test for a minor thing. Better reduce by to a smaller version manually (same also goes for the earlier jtbl test added, which I missed to comment on).

m2c/arch_sh.py
sozud: Ok added to the pattern

m2c/arch_sh.py
sozud: Removed

tests/end_to_end/sh-far-jump/orig.c
sozud: Ok replaced both tests with minimized asm versions

m2c/arch_sh.py
```diff
@@ -79,6 +79,47 @@
...
+            return None
+        target = entry.data_at_offset(0, 4)
+        if isinstance(target, AsmSymbolicData) and isinstance(
+            target.data, AsmGlobalSymbol
+        ):
+            target_name = target.data.symbol_name
+        elif isinstance(target, AsmSymbolicData) and isinstance(
+            target.data, AsmLiteral
+        ):
+            target_name = f".L{target.data.value:08X}"
```
simonlindholm: This too should not be here

m2c/arch_sh.py
```diff
@@ -79,6 +79,47 @@
...
+        if not isinstance(load.args[0], AsmGlobalSymbol):
+            return None
+        entry = m.asm_data.values.get(load.args[0].symbol_name)
+        if entry is None:
+            return None
+        target = entry.data_at_offset(0, 4)
+        if isinstance(target, AsmSymbolicData) and isinstance(
+            target.data, AsmGlobalSymbol
+        ):
+            target_name = target.data.symbol_name
```
simonlindholm: Feels like this could use a common helper (`Argument -> Optional[str]`); it's duplicated from DivisionHelperPattern and also occurs in `arm_jtbl_for_ldr` (and that copy handles an addend for the AsmGlobalSymbol too)

simonlindholm: I added `get_literal_pool_symbol` for this on latest master

sozud: Using the api now

m2c/arch_sh.py
sozud: Removed

PR #343, opened 2026-08-05
sh2: implement rotcl

sozud:
Similar to the other "T" instructions like addc.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -640,7 +640,7 @@ class Sh2Arch(Arch):
                 result = (
                     shifted
                     if carry_in == Literal(0)
-                    else BinaryOp.uint(shifted, "|", carry_in)
+                    else BinaryOp.int(shifted, "|", carry_in)
                 )
 
                 result = s.set_reg(a.reg_ref(0), result)
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -634,20 +634,17 @@ class Sh2Arch(Arch):
 
             def eval_fn(s: NodeState, a: InstrArgs) -> None:
                 original = a.reg(0)
-                unsigned = as_type(original, Type.u32(), silent=True)
-                shifted = fold_mul_chains(
-                    BinaryOp.int(unsigned, "<<", Literal(1)),
-                    allow_sll_chains=True,
-                )
+                shifted = BinaryOp.ushift(original, "<<", Literal(1))
+
                 carry_in = a.regs[t_reg]
                 result = (
-                    as_type(shifted, Type.u32(), silent=True)
+                    shifted
                     if carry_in == Literal(0)
                     else BinaryOp.uint(shifted, "|", carry_in)
                 )
-                s.set_reg(a.reg_ref(0), result)
-                carry = BinaryOp.uint(unsigned, ">>", Literal(31))
-                s.set_reg(t_reg, carry)
+
+                result = s.set_reg(a.reg_ref(0), result)
+                s.set_reg(t_reg, CarryBit(result))
 
         elif mnemonic in cls.instrs_shift:
             assert len(args) == 1 and isinstance(args[0], Register)
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -626,6 +626,29 @@ class Sh2Arch(Arch):
                 s.set_reg(Register("macl"), Literal(0))
                 s.set_reg(Register("mach"), Literal(0))
 
+        elif mnemonic == "rotcl":
+            assert len(args) == 1 and isinstance(args[0], Register)
+            t_reg = Register("condition_bit")
+            inputs = [args[0], t_reg]
+            outputs = [args[0], t_reg]
+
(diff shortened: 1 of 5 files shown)


m2c/arch_sh.py
```diff
@@ -623,6 +623,29 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+        elif mnemonic == "rotcl":
+            assert len(args) == 1 and isinstance(args[0], Register)
+            t_reg = Register("condition_bit")
+            inputs = [args[0], t_reg]
+            outputs = [args[0], t_reg]
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                original = a.reg(0)
+                unsigned = as_type(original, Type.u32(), silent=True)
+                shifted = fold_mul_chains(
```
simonlindholm: Surely rotcl isn't used to multiply by two in practice, is it?

m2c/arch_sh.py
```diff
@@ -623,6 +623,29 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
 
+        elif mnemonic == "rotcl":
+            assert len(args) == 1 and isinstance(args[0], Register)
+            t_reg = Register("condition_bit")
+            inputs = [args[0], t_reg]
+            outputs = [args[0], t_reg]
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                original = a.reg(0)
+                unsigned = as_type(original, Type.u32(), silent=True)
```
simonlindholm: Why this instead of just using `original` directly? If you add `unify=False` in order to prevent `original` from becoming a `u32` it would make sense to add a cast (and this is consistent with `shll`/`shlr`). I don't know if we want to do that here; I don't really care to think to deeply about it since `rotcl` is so rare, but the current code feels a bit off.

m2c/arch_sh.py
```diff
@@ -623,6 +623,29 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+                    allow_sll_chains=True,
+                )
+                carry_in = a.regs[t_reg]
+                result = (
+                    as_type(shifted, Type.u32(), silent=True)
+                    if carry_in == Literal(0)
+                    else BinaryOp.uint(shifted, "|", carry_in)
+                )
+                s.set_reg(a.reg_ref(0), result)
+                carry = BinaryOp.uint(unsigned, ">>", Literal(31))
```
simonlindholm: BinaryOp.ushift, but I'm actually more tempted to set the carry out to `CarryBit(return value from set_reg)` like how we handle `addc`. This gives C code output that's not compilable with the exact semantics of the asm, but that's not really the goal of m2c either; rather we aim to give output which is useful for matching decompilation and annotating the carry bit as such means that it's easier to puzzle together 64-bit arithmetic.

m2c/arch_sh.py
```diff
@@ -623,6 +623,29 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+            assert len(args) == 1 and isinstance(args[0], Register)
+            t_reg = Register("condition_bit")
+            inputs = [args[0], t_reg]
+            outputs = [args[0], t_reg]
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                original = a.reg(0)
+                unsigned = as_type(original, Type.u32(), silent=True)
+                shifted = fold_mul_chains(
+                    BinaryOp.int(unsigned, "<<", Literal(1)),
```
simonlindholm: If you want to make the output unsigned you can use `BinaryOp.ushift` here

m2c/arch_sh.py
sozud: Ok removed as_type

m2c/arch_sh.py
sozud: I think it's meant for a 64-bit rotation. I removed fold_mul_chains

m2c/arch_sh.py
sozud: Ok using ushift

m2c/arch_sh.py
sozud: Ok using CarryBit now

sozud:
This ought to be passing but the runner failed due to github problems it seems

m2c/arch_sh.py
```diff
@@ -626,6 +626,26 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                original = a.reg(0)
+                shifted = BinaryOp.ushift(original, "<<", Literal(1))
+
+                carry_in = a.regs[t_reg]
+                result = (
+                    shifted
+                    if carry_in == Literal(0)
+                    else BinaryOp.uint(shifted, "|", carry_in)
```
```suggestion
                    else BinaryOp.int(shifted, "|", carry_in)
```
to make the output a bit more readable

simonlindholm:
Yeah, GitHub Actions seems to be having major issues: https://www.githubstatus.com/. Let's wait until they settle

m2c/arch_sh.py

PR #344, opened 2026-08-08
sh2: implement rotcr

sozud:
Same as https://github.com/matt-kempster/m2c/pull/343
Disclosure: AI assistance was using during development.


simonlindholm (commented):
More skeptical towards using CarryBit() here, since it's no longer a 33'nd bit of a number. Here I'd probably just go with `BinaryOp.int(original, "&", Literal(1))`

sozud:
>More skeptical towards using CarryBit() here, since it's no longer a 33'nd bit of a number. Here I'd probably just go with BinaryOp.int(original, "&", Literal(1))

Ok adjusted this and added a couple of tests

PR #345, opened 2026-08-08
sh2: implement negc

sozud:
Another "T" instruction.
Disclosure: AI assistance was using during development.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -561,6 +561,36 @@ class Sh2Arch(Arch):
                 s.set_reg(a.reg_ref(1), result)
                 s.set_reg(t_reg, borrow)
 
+        elif mnemonic == "negc":
+            assert (
+                len(args) == 2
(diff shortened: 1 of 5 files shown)


m2c/arch_sh.py
```diff
@@ -561,6 +561,36 @@ def eval_fn(s: NodeState, a: InstrArgs) -> None:
...
+                and isinstance(args[1], Register)
+            )
+            t_reg = Register("condition_bit")
+            inputs = [args[0], t_reg]
+            outputs = [args[1], t_reg]
+
+            def eval_fn(s: NodeState, a: InstrArgs) -> None:
+                value = a.reg(0)
+                carry = a.regs[t_reg]
+                result = handle_sub(UnaryOp.sint("-", value), carry)
```
```suggestion
                result = handle_sub(UnaryOp.int("-", value), carry)
```

sozud: Fixed

PR #348, opened 2026-08-23
sh2: fix incorrect writeback increments

sozud:
Previously the writeback instructions were incrementing pointers by the wrong amount. This adds a `addbytes.fictive` instruction to keep this change separate from the stack handling. Now we get:
```
s16 *temp_r1;
temp_r1 = *ptr - 1;
```
instead of
```
temp_r1 = *ptr - 2;
```

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -234,7 +234,9 @@ class Sh2AddrModeWritebackPattern(AsmPattern):
         if addr.writeback == Writeback.PRE:
             return Replacement(
                 [
-                    AsmInstruction("add", [AsmLiteral(-stride), addr.base]),
+                    AsmInstruction(
+                        "addbytes.fictive", [AsmLiteral(-stride), addr.base]
+                    ),
                     AsmInstruction(instr.mnemonic, new_args),
                 ],
                 1,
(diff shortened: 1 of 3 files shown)


simonlindholm:
This is a pre-existing quirk of m2c, where its output for addition/subtraction of pointers deviates from normal C. It should probably be better documented, but it's somewhat intentional: my experience from IDA/Ghidra, which do pointer arithmetic according to normal C rules, is that what often happens is that a pointer to struct A (of size, say, 0xa0) gets misdetected as a pointer to u32, and then there's a loop that increments a pointer by 0x28 in each step, and you end up having to look at the declaration of puVar20 and do mental arithmetic to reconstruct 0x28 * 4 = 0xa0, and it's more effort than just having the 0xa0 written out in the first place.

It could for sure be argued that it's confusing. If we wanted to change it, the place to do it would be in the formatting logic for BinaryOp (with op = "+"/"-").

simonlindholm:
Possibly we could make --valid-syntax emit casts for pointer arithmetic, emitting e.g. `var_a0 = (struct A*)((u8*)var_a0 + 4);` instead of `var_a0 += 4;`.

sozud:
I see, thanks for explaining. I misunderstood and thought this was an sh2 bug. The current behavior isn't blocking anything in particular so I think it's fine for now.

PR #351, opened 2026-08-23
sh2: add cmp/str intrinsic

sozud:
`cmp/str` checks if any of the 4 bytes in a register are non-zero and sets T. It seems to be meant to enable faster scanning for the null byte in strings. I added an intrinsic for it since I think this only shows up in handwritten code.


simonlindholm:
Can you add M2C_CMP_STR to m2c_macros.h?

sozud:
> Can you add M2C_CMP_STR to m2c_macros.h?

done

PR #352, opened 2026-08-27
sh2: implement tas.b

sozud:
This adds an instrinsic for tas.b As far as I know it's only used in handwritten mutex-locking type code.


simonlindholm:
Sure. Although to be totally honestly I don't see much value in supporting handwritten asm, especially with such rare instructions.

PR #353, opened 2026-08-27
sh2: add shc braf jump table support

sozud:
SHC is the SH2 compiler by Hitachi. This adds support for a type of jump table generated by it that uses `braf`.  Disclosure: AI assistance was used in developing this PR.

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -164,6 +164,48 @@ class JumpTablePattern(AsmPattern):
         return None
 
 
+class BrafJumpTablePattern(SimpleAsmPattern):
+    pattern = make_pattern(
+        "mova _, $b",
+        "mov.w @($b,$i),$b",
+        "braf $b",
+        "nop",
+        ".base:",
+    )
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        mova = m.body[0]
+        assert isinstance(mova, Instruction)
+        if not isinstance(mova.args[0], AsmGlobalSymbol):
+            return None
+        table = m.asm_data.values.get(mova.args[0].symbol_name)
+        if table is None:
+            return None
+        base = m.labels[".base"]
+        targets: List[AsmGlobalSymbol] = []
+        for entry in table.data:
+            if (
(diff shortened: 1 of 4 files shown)


m2c/arch_sh.py
```diff
@@ -164,6 +164,48 @@ def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
...
+                or entry.data.rhs.symbol_name not in base.names
+            ):
+                return None
+            targets.append(entry.data.lhs)
+        if not targets:
+            return None
+        return Replacement(
+            [
+                AsmInstruction("tablejmp.doubled.fictive", [m.regs["i"], *targets]),
+                AsmInstruction("nop", []),
```
simonlindholm: It's a bit silly that we've chosen to make tablejmp.doubled.fictive a delay slot instruction, given that we always emit a nop after.

m2c/arch_sh.py
```diff
@@ -164,6 +164,48 @@ def match(self, matcher: AsmMatcher) -> Optional[Replacement]:
...
+                or not isinstance(entry.data.rhs, AsmGlobalSymbol)
+                or entry.data.rhs.symbol_name not in base.names
+            ):
+                return None
+            targets.append(entry.data.lhs)
+        if not targets:
+            return None
+        return Replacement(
+            [
+                AsmInstruction("tablejmp.doubled.fictive", [m.regs["i"], *targets]),
```
simonlindholm: Make a temp var for this to improve formatting

simonlindholm:
(Out of curiosity, is there some new target that you're looking into for that prompted this?)

m2c/arch_sh.py
sozud: Ok fixed for both

m2c/arch_sh.py

sozud:
> (Out of curiosity, is there some new target that you're looking into for that prompted this?)

I'm looking at Dreamcast. It's not that much new code so I figured it would be good to have. I'm fixing the missing parts of the SH2 implementation first

simonlindholm:
Exciting! Definitely sounds like a good target.

PR #354, opened 2026-08-28
sh2: support shc bsrf calls

sozud:
This is another SHC pattern where it uses bsrf + a pool value to jump to a function. Disclosure: AI assistance

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -26,7 +26,13 @@ from .asm_pattern import (
     SimpleAsmPattern,
     make_pattern,
 )
-from .flow_graph import ArchFlowGraph, FlowGraph, Node, get_literal_pool_symbol
+from .flow_graph import (
+    ArchFlowGraph,
+    FlowGraph,
+    Node,
+    get_literal_pool_data,
+    get_literal_pool_symbol,
+)
 from .instruction import (
     Instruction,
     InstructionMeta,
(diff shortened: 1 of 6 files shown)


m2c/arch_sh.py
```diff
@@ -195,6 +195,44 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        load = m.body[0]
+        delay = m.wildcard_items[0]
+        if not isinstance(load, Instruction) or not isinstance(delay, Instruction):
+            return None
+        if not isinstance(load.args[0], AsmGlobalSymbol):
+            return None
+        literal = m.asm_data.values.get(load.args[0].symbol_name)
+        if literal is None or len(literal.data) != 1:
```
```suggestion
        if literal is None or not literal.data:
```
to accept zero padding that comes after the literal data.

Would also be nice to get a version of `get_literal_pool_symbol` that returns `Optional[AsmSymbolicData]` instead of `Optional[str]`, so that we more consistently deal with `symbol + offset`, check is_text and entry.size == 4, and can remove the isinstance(load.args[0], AsmGlobalSymbol) boilerplate.

m2c/arch_sh.py
```diff
@@ -195,6 +195,44 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+        "mov.l _,$r",
+        "bsrf $r",
+        "*",
+        ".base:",
+    )
+
+    def replace(self, m: AsmMatch) -> Optional[Replacement]:
+        load = m.body[0]
+        delay = m.wildcard_items[0]
+        if not isinstance(load, Instruction) or not isinstance(delay, Instruction):
```
simonlindholm: Make isinstance(load, Instruction) an assert

m2c/arch_sh.py
sozud: Are you thinking like this? https://github.com/matt-kempster/m2c/commit/b93e0cf7e3d88438e703bf0b4187adfd87d560da

simonlindholm: Yeah, exactly. I would name it `get_literal_pool_symbolic_data` and not just `get_literal_pool_data` since data could refer to bytes-valued data as well

m2c/arch_sh.py

m2c/arch_sh.py
```diff
@@ -195,6 +201,41 @@ def replace(self, m: AsmMatch) -> Optional[Replacement]:
...
+        load = m.body[0]
+        assert isinstance(load, Instruction)
+        delay = m.wildcard_items[0]
+        if not isinstance(delay, Instruction):
+            return None
+        entry = get_literal_pool_symbolic_data(load.args[0], m.asm_data)
+        if entry is None:
+            return None
+        if (
+            not isinstance(entry.data, BinOp)
```
```suggestion
        if (
            entry is None
            or not isinstance(entry.data, BinOp)
```

simonlindholm (approved):

PR #355, opened 2026-08-28
sh2: support div0u sequences

sozud:
SHC inlines these sorts of division sequences:
```
div0u
rotcl r0
div1 r5,r1
etc ...
```
This adds a fictive instruction to handle unsigned division and modulo. Disclosure: AI assistance

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -235,6 +235,28 @@ class BsrfCallPattern(SimpleAsmPattern):
         )
 
 
+DIV1_STEP_COUNT = 32
+
+
+class Div0u32Pattern(SimpleAsmPattern):
+
+    pattern = make_pattern(
+        "xor $r, $r",
+        "div0u",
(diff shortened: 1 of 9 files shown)


simonlindholm:
It inlines 67-instruction sequences!? I guess that's... a priority you can have.

Just to double check: can there be any reorderings wrt adjacent instructions? Do you have any decomp.me link that examplifies this pattern?

sozud:
Ok so I investigated more with the compiler now and it seems like I got this wrong. The compiler does not emit this for any typical plain division I tested, it calls division helpers instead. There's various routines in these games that inline this sequence plus various other operations and I assumed these were inlined by the compiler since it wasn't just a plain division helper. I think these must either be handwritten or using compiler intrinsics that I haven't figured out how to use yet. I'll close this for now unless I come up with a better example.

simonlindholm:
Maybe it's triggered by -Ofast or similar? Though if it's only in certain helpers maybe that's less likely. Leaving this for now sounds reasonable.

PR #356, opened 2026-08-28
sh4: add target, test infra and shc to gas converter

sozud:
This adds some plumbing for sh4 and generating e2e tests. I'm using this https://github.com/decompme/compilers/blob/main/platforms/dreamcast/shc-v5.1r03/Dockerfile
A script is added to convert SHC syntax to GNU asm. AFAIK there's no precedent in m2c for alternative assembler syntaxes so I assumed this would be the preferred approach.
An e2e test is added to test the SHC output.
Sh4Arch is just empty at the moment and falls back to sh2. It's a superset of the sh2 instruction set so this is fine.

### DIFF
--- a/m2c/options.py
+++ b/m2c/options.py
@@ -43,6 +43,7 @@ class Target:
         ARM = "arm"
         GBA = "gba"
         SH2 = "sh2"
+        SH4 = "sh4"
 
         @property
         def arch(self) -> Target.ArchEnum:
@@ -58,8 +59,8 @@ class Target:
                 return Target.ArchEnum.ARM
             elif self == Target.PlatformEnum.GBA:
                 return Target.ArchEnum.ARM
-            elif self == Target.PlatformEnum.SH2:
-                return Target.ArchEnum.SH2
+            elif self in (Target.PlatformEnum.SH2, Target.PlatformEnum.SH4):
+                return Target.ArchEnum.SH
             else:
                 assert_never(self)
 
@@ -67,7 +68,7 @@ class Target:
         MIPS = "mips"
         PPC = "ppc"
         ARM = "arm"
-        SH2 = "sh2"
+        SH = "sh"
 
     class EndianEnum(ChoicesEnum):
         LITTLE = "little"
@@ -77,6 +78,7 @@ class Target:
         IDO = "ido"
         GCC = "gcc"
         MWCC = "mwcc"
+        SHC = "shc"
 
     class LanguageEnum(ChoicesEnum):
         C = "c"
@@ -108,12 +110,15 @@ class Target:
                 Target.PlatformEnum.MIPSEE,
                 Target.PlatformEnum.ARM,
                 Target.PlatformEnum.GBA,
+                Target.PlatformEnum.SH4,
             ):
                 endian = Target.EndianEnum.LITTLE
             arch = platform.arch
 
             if len(terms) >= 2:
                 compiler = Target.CompilerEnum(terms[1])
+            elif platform == Target.PlatformEnum.SH4:
+                compiler = Target.CompilerEnum.SHC
             elif arch == Target.ArchEnum.PPC:
                 compiler = Target.CompilerEnum.MWCC
             elif arch == Target.ArchEnum.MIPS and endian == Target.EndianEnum.BIG:
--- a/tests/add_test.py
+++ b/tests/add_test.py
@@ -11,6 +11,8 @@ from tempfile import NamedTemporaryFile
 from typing import Dict, List, Optional, Tuple
 from dataclasses import dataclass, field, replace
 
+import shc_to_gas
+
 logger = logging.getLogger(__name__)
 
 
@@ -27,6 +29,7 @@ class PathsToBinaries:
     MWCC_CC: Optional[Path]
     AGBCC: Optional[Path]
     SH_CC: Optional[Path]
+    SHC_CC: Optional[Path]
     WINE: Optional[Path]
 
 
@@ -54,6 +57,10 @@ def get_environment_variables() -> PathsToBinaries:
         "SH_CC",
         "env variable SH_CC should point to SHGCC. see tools/setup_sh_compiler.sh",
     )
+    SHC_CC = load(
+        "SHC_CC",
+        "env variable SHC_CC should point to SHC",
+    )
     WINE = None
     if MWCC_CC and sys.platform.startswith("linux"):
         WINE = load("WINE", "env variable WINE should point to wine or wibo binary")
@@ -62,6 +69,7 @@ def get_environment_variables() -> PathsToBinaries:
         MWCC_CC=MWCC_CC,
         AGBCC=AGBCC,
         SH_CC=SH_CC,
+        SHC_CC=SHC_CC,
         WINE=WINE,
     )
 
@@ -170,12 +178,41 @@ def get_sh_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
     return []
 
 
+def get_shc_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
+    if paths.SHC_CC is not None:
+        shc = Compiler(
+            name="shc",
+            cc_command=[
+                str(paths.SHC_CC),
+                "-comment=nonest",
+                "-cpu=sh4",
+                "-division=cpu",
+                "-endian=little",
+                "-fpu=single",
+                "-macsave=0",
+                "-sjis",
+                "-string=const",
+            ],
+        )
+        return [
+            (
+                "sh4-shc-o1",
+                shc.with_cc_flags(
+                    ["-optimize=1", "-speed", "-aggressive=2", "-code=asmcode"]
+                ),
+            )
+        ]
+    logger.warning("SHC compiler not found; skipping")
+    return []
+
+
 def get_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
     compilers: List[Tuple[str, Compiler]] = []
     compilers.extend(get_ido_compilers(paths))
     compilers.extend(get_mwcc_compilers(paths))
     compilers.extend(get_agbcc_compilers(paths))
     compilers.extend(get_sh_compilers(paths))
+    compilers.extend(get_shc_compilers(paths))
     return compilers
 
 
@@ -212,7 +249,12 @@ def do_compilation_step(
 def run_compile(in_file: Path, out_file: Path, compiler: Compiler) -> None:
     flags_str = " ".join(compiler.cc_command)
     logger.info(f"Compiling {in_file} to {out_file} using: {flags_str}")
-    if compiler.name in ("agbcc", "sh-gcc"):
+    if compiler.name == "shc":
+        with NamedTemporaryFile(suffix=".src") as temp_src_file:
+            do_compilation_step(temp_src_file.name, str(in_file), compiler)
+            hitachi_asm = Path(temp_src_file.name).read_text()
+        out_file.write_text(shc_to_gas.convert(hitachi_asm))
+    elif compiler.name in ("agbcc", "sh-gcc"):
         with NamedTemporaryFile(suffix=".i") as temp_i_file:
             subprocess.run(
                 ["cpp", "-P", "-o", temp_i_file.name, str(in_file)], check=True
@@ -261,7 +303,7 @@ def add_test_from_file(
             continue
         try:
             run_compile(orig_file, asm_file_path, compiler)
-            if compiler.name in ("mwcc", "agbcc", "sh-gcc"):
+            if compiler.name in ("mwcc", "agbcc", "sh-gcc", "shc"):
                 # If the flags file doesn't exist, initialize it with the correct --target
                 flags_path = test_dir / (asm_filename + "-flags.txt")
                 if not flags_path.exists():
@@ -269,6 +311,8 @@ def add_test_from_file(
                         target = "ppc-mwcc-c"
                     elif compiler.name == "agbcc":
                         target = "gba"
+                    elif compiler.name == "shc":
+                        target = "sh4-shc-c"
                     else:
                         target = "sh2-gcc-c"
                     flags_path.write_text(f"--target {target}\n")
--- a/tests/add_test.py
+++ b/tests/add_test.py
@@ -192,14 +192,13 @@ def get_shc_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
                 "-macsave=0",
                 "-sjis",
                 "-string=const",
+                "-code=asmcode",
             ],
         )
         return [
             (
                 "sh4-shc-o1",
-                shc.with_cc_flags(
-                    ["-optimize=1", "-speed", "-aggressive=2", "-code=asmcode"]
-                ),
+                shc.with_cc_flags(["-optimize=1", "-speed", "-aggressive=2"]),
             )
         ]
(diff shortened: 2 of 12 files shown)


tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
+#!/usr/bin/env python3
+
+import sys
+from typing import List, Optional, Tuple
+
+
+def ascii_upper(c: str) -> str:
```
simonlindholm: This is just `.upper()`, assuming you don't provide anything outside of ASCII

tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
...
+    if "a" <= c <= "z":
+        return chr(ord(c) - (ord("a") - ord("A")))
+    return c
+
+
+def ieq(a: str, b: str) -> bool:
```
simonlindholm: and this is `a.upper() == b.upper()`

tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
...
+    if len(a) != len(b):
+        return False
+
+    return all(ascii_upper(x) == ascii_upper(y) for x, y in zip(a, b))
+
+
+def istarts(s: str, prefix: str) -> bool:
```
simonlindholm: `s.upper().startswith(prefix.upper())`

tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
...
+                out.append(text[i])
+                i += 1
+        else:
+            out.append(text[i])
+            i += 1
+
+    return "".join(out)
+
+
+def parse_fr(s: str, pos: int) -> Optional[Tuple[int, str]]:
```
simonlindholm: What does this function do, can you add a comment?

tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
...
+    while pos < len(s) and s[pos].isdigit():
+        pos += 1
+
+    if pos == digits_start:
+        return None
+
+    return pos, s[start:pos]
+
+
+def convert_fmov_reg_reg(raw: str) -> Optional[str]:
```
simonlindholm: also warrants a comment

tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def istarts(s: str, prefix: str) -> bool:
+    if len(s) < len(prefix):
+        return False
+    return ieq(s[: len(prefix)], prefix)
+
+
+def trim_view(s: str) -> str:
+    return s.strip()
```
simonlindholm: just inline this wherever it's used

tests/add_test.py
```diff
@@ -164,12 +170,24 @@ def get_sh_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
     return []
 
 
+def get_shc_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
+    if paths.SHC_CC is not None:
+        shc = Compiler(
+            name="shc",
+            cc_command=[str(paths.SHC_CC)],
+        )
+        return [("sh4-shc-o1", shc.with_cc_flags(["-O1"]))]
```
simonlindholm: Is -O1 the flag commonly used in games? Feels so odd when you're used to GCC -O2

tests/shc_to_gas.py
```diff
@@ -0,0 +1,265 @@
+#!/usr/bin/env python3
```
simonlindholm, line 1: Can you provide a short doc comment at the top of the file exemplifying the kind of transform you do here? Like a 10-line asm before/after

tests/add_test.py
```diff
@@ -206,7 +224,14 @@ def do_compilation_step(
 def run_compile(in_file: Path, out_file: Path, compiler: Compiler) -> None:
     flags_str = " ".join(compiler.cc_command)
     logger.info(f"Compiling {in_file} to {out_file} using: {flags_str}")
-    if compiler.name in ("agbcc", "sh-gcc"):
+    if compiler.name == "shc":
+        from shc_to_gas import convert
```
simonlindholm: Put the import at the top of the file (with an `import as` or importing just `shc_to_gas` and using `sha_to_gas.convert` to make the name less generic).

The only reason the ppc disassembler breaks this rule is that it relies on capstone which is an optional dependency and has a tendency to break in annoying manners.

tests/add_test.py
```diff
@@ -54,6 +55,10 @@ def load(env_var_name: str, error_message: str) -> Optional[Path]:
         "SH_CC",
         "env variable SH_CC should point to SHGCC",
     )
+    SHC_CC = load(
+        "SHC_CC",
+        "env variable SHC_CC should point to SHC",
+    )
```
simonlindholm, line 63: Hm, we really need some documentation about how to point it there. Don't know where to put it, but definitely need it

m2c/arch_sh4.py
simonlindholm, line 1: I don't think we need a separate file for sh4.

m2c/options.py
```diff
@@ -58,7 +59,7 @@ def arch(self) -> Target.ArchEnum:
                 return Target.ArchEnum.ARM
             elif self == Target.PlatformEnum.GBA:
                 return Target.ArchEnum.ARM
-            elif self == Target.PlatformEnum.SH2:
+            elif self in (Target.PlatformEnum.SH2, Target.PlatformEnum.SH4):
                 return Target.ArchEnum.SH2
```
simonlindholm: Maybe this ArchEnum should be renamed SH?

simonlindholm (commented):
Do you have a reference manual for the sh4 ISA so I can follow along better? Plus ABI reference?

tests/add_test.py
sozud: My local setup looks like this. Is this something you want upstreamed? https://github.com/matt-kempster/m2c/commit/f9f25f65e5b2fe7675ac2509a64ae9f98aa5a1b1

sozud:
> Do you have a reference manual for the sh4 ISA so I can follow along better? Plus ABI reference?

ISA manual
[h14tp003d2.pdf](https://github.com/user-attachments/files/31564320/h14tp003d2.pdf)


ABI See 9.3.2 Function Calling Interface
[REN_rej10b0152_sh_MAT_20041215_1.pdf](https://github.com/user-attachments/files/31564246/REN_rej10b0152_sh_MAT_20041215_1.pdf)

tests/add_test.py
simonlindholm: Yes, that would be great.

I think instead of translating -O0/-O1 to -optimize=0/-optimize=1 it would be better to pass the latter flag directly.

The dependence on https://github.com/Xeeynamo/sotn-decomp is a bit uncomfortable (I could see that repo vanishing). Not a big deal though, just having the compiler name + how to use it is 99% of the value.

tests/shc_to_gas.py
sozud: ok removed

tests/shc_to_gas.py
sozud: removed

tests/shc_to_gas.py
sozud: removed

tests/shc_to_gas.py
sozud: It parses floating point register names, I added a comment

tests/shc_to_gas.py
sozud: Added comment

tests/shc_to_gas.py
sozud: inlined

tests/add_test.py
sozud: The compiler has optimizations on or off then you can select what to optimize.
image

tests/shc_to_gas.py
sozud: Added comment and example

tests/add_test.py
sozud: Moved import

m2c/arch_sh4.py
sozud: removed

m2c/options.py
sozud: renamed

tests/shc_to_gas.py
```diff
@@ -0,0 +1,280 @@
...
+            and text[i].upper() == "H"
+            and text[i + 1] == "'"
+            and is_hex_digit(text[i + 2])
+        ):
+            out.append("0x")
+            i += 2
+
+            while i < len(text) and is_hex_digit(text[i]):
+                out.append(text[i])
+                i += 1
```
simonlindholm:
```suggestion
```
Since this will be taken care of by the original loop

tests/add_test.py
sozud: This is rebased to use the new setup from the other PR now

tests/shc_to_gas.py
```diff
@@ -0,0 +1,280 @@
...
+import sys
+from typing import List, Optional, Tuple
+
+
+def is_hex_digit(c: str) -> bool:
+    return c in "0123456789abcdefABCDEF"
+
+
+def hex_converted(text: str) -> str:
+    """convert H'1234 into GNU as syntax 0x1234"""
```
simonlindholm, line 42: This kind of string manipulation is the perfect place to use regexes; you can replace the entire function by:
```py
    return re.sub(r"h'(?=[0-9a-f])", "0x", text, flags=re.IGNORECASE)
```

(the `?=` here is a positive lookahead; if you didn't want it you could instead do
```py
    return re.sub(r"h'([0-9a-f])", r"0x\1", text, flags=re.IGNORECASE)
```
)

tests/shc_to_gas.py
```diff
@@ -0,0 +1,280 @@
...
+        pos += 1
+
+    if pos == digits_start:
+        return None
+
+    return pos, s[start:pos]
+
+
+def convert_fmov_reg_reg(raw: str) -> Optional[str]:
+    """FMOV.S FRn,FRm -> FMOV FRn,FRm"""
```
simonlindholm, line 47: The regex version of this one is
```py
    return re.sub(r"\bFMOV\.S(\s+FR\d+\s*,\s*FR\d+)", r"FMOV\1", raw, flags=re.IGNORECASE)
```
(making the function return `str` instead of `Optional[str]`)

tests/shc_to_gas.py
```diff
@@ -0,0 +1,280 @@
...
+
+    if not all(c.isdigit() for c in s):
+        return None
+
+    return int(s, 10)
+
+
+def convert_line(line: str) -> str:
+    if line.endswith("\r"):
+        line = line[:-1]
```
simonlindholm:
```suggestion
    line = line.rstrip("\r")
```

tests/shc_to_gas.py
```diff
@@ -0,0 +1,280 @@
...
+
+    semi = line.find(";")
+
+    if semi != -1:
+        converted = line[:semi]
+
+        if semi + 1 < len(line):
+            converted += " !" + line[semi + 1 :]
+
+        line = converted
```
simonlindholm:
```suggestion
        line = line[:semi] + " !" + line[semi + 1 :]
```
(I assume it's fine to convert `;` at end of line to `!` at end of line)

tests/add_test.py
```diff
@@ -170,12 +178,41 @@ def get_sh_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
...
+    if paths.SHC_CC is not None:
+        shc = Compiler(
+            name="shc",
+            cc_command=[
+                str(paths.SHC_CC),
+                "-comment=nonest",
+                "-cpu=sh4",
+                "-division=cpu",
+                "-endian=little",
+                "-fpu=single",
```
simonlindholm, line 191: Curious about this fpu flag, since it impacts abi: is codegen the same without it, assuming you avoid using doubles? Do any games use doubles?

tests/add_test.py
```diff
@@ -170,12 +178,41 @@ def get_sh_compilers(paths: PathsToBinaries) -> List[Tuple[str, Compiler]]:
...
+                "-macsave=0",
+                "-sjis",
+                "-string=const",
+            ],
+        )
+        return [
+            (
+                "sh4-shc-o1",
+                shc.with_cc_flags(
+                    ["-optimize=1", "-speed", "-aggressive=2", "-code=asmcode"]
```
simonlindholm: is -code=asmcode an optimization flag, or should it be moved to the cc_command list?

tests/shc_to_gas.py
sozud: Done

tests/shc_to_gas.py
sozud: Done

tests/shc_to_gas.py
sozud: Done

tests/shc_to_gas.py
sozud: Done

tests/add_test.py
sozud: I copied this from a decomp.me preset, Hitachi says:

```
“There are two possible fpu options: fpu=single and fpu=double.
When fpu=single is selected, all floating-point data is handled as single-precision floating-point data.
When fpu=double is selected, all floating-point data is handled as double-precision floating-point data.
When no fpu option is used, floating-point data is used according to the declaration type of the source code.”
```
https://www.renesas.com/en/document/apn/sh-compiler-application-note-2-compiler-use-guide-coding-guide-using-fpu

So if a game was using the double type then yes this affects codegen. I'm not sure if there's any examples at the moment

tests/add_test.py
sozud: This makes it emit asm instead of an object, I moved it

tests/shc_to_gas.py
```diff
@@ -0,0 +1,196 @@
...
+    return int(s, 10)
+
+
+def convert_line(line: str) -> str:
+    line = line.rstrip("\r")
+
+    semi = line.find(";")
+
+    if semi != -1:
+        line = line[:semi] + " !" + line[semi + 1 :]
```
simonlindholm:
```suggestion
    line = line.replace(";", " !", 1)
```

simonlindholm (approved):

simonlindholm:
> AFAIK there's no precedent in m2c for alternative assembler syntaxes so I assumed this would be the preferred approach.

Yeah; in principle I'm not opposed to adding other syntaxes if the need comes up (especially if it's viable to support both syntaxes at the same time so it doesn't need a flag), but it doesn't feel like it's really needed in this case. I imagine sh4 decomps in the wild already use GNU as?

sozud:
> Yeah; in principle I'm not opposed to adding other syntaxes if the need comes up (especially if it's viable to support both syntaxes at the same time so it doesn't need a flag), but it doesn't feel like it's really needed in this case. I imagine sh4 decomps in the wild already use GNU as?

The three Dreamcast decomps I'm aware of all use asmsh

https://github.com/Exant64/sa2dc/blob/ba180d99e14936b7eaacabb62fcfb348dc038ca5/makefile#L47
https://github.com/fmil95/recv-dc-decomp/blob/7ee766b4f3a8eacb2f09577a88a565ae73f4adb9/makefile#L40
https://github.com/lhsazevedo/tbg-decomp/blob/65c4e5b4083d7fd308fafb99ff3400b7870bbfc3/Makefile#L126

I tried migrating one to GNU as and I don't think there's any issue, it matched https://github.com/fmil95/recv-dc-decomp/pull/1

PR #357, opened 2026-08-29
tools: add superh compilers

sozud:
These are the scripts to build the docker containers and use the superh compilers. The saturn one now uses my compilers repo rather than sotn-decomp. The dreamcast one now passes the optimization part directly.

### DIFF
--- a/tools/shc_compiler/compile.sh
+++ b/tools/shc_compiler/compile.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+set -eu
+
+optimize_flag=${1:--optimize=1}
+case "$optimize_flag" in
+    -optimize=0|-optimize=1) ;;
+    *)
+        echo "unsupported optimization flag: $optimize_flag (SHC has -optimize=0 and -optimize=1)" >&2
+        exit 2
+        ;;
+esac
+
+test -f /work/input.c
+
+rm -rf /tmp/shc-compile
+mkdir -p /tmp/shc-compile
+cp -a /opt/shc/bin/. /tmp/shc-compile/
+sed -e 's/\r$//' -e 's/$/\r/' /work/input.c > /tmp/shc-compile/input.c
+
+cd /tmp/shc-compile
+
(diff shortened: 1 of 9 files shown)


tools/shc_compiler/compile.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+set -eu
+
+optimize_flag=${1:--optimize=1}
+case "$optimize_flag" in
+    -optimize=0|-optimize=1) ;;
+    *)
+        echo "unsupported optimization flag: $optimize_flag (SHC has -optimize=0 and -optimize=1)" >&2
+        exit 2
+        ;;
```
simonlindholm: Instead of special-casing this like this, could you just pass all flags through?

tools/shc_compiler/shc
```diff
@@ -0,0 +1,47 @@
...
+            ;;
+        -o)
+            test "$#" -ge 2 || { echo "-o requires a filename" >&2; exit 2; }
+            output=$2
+            shift 2
+            ;;
+        -*)
+            echo "unsupported shc option: $1" >&2
+            exit 2
+            ;;
```
simonlindholm, line 21: and similarly here, just push every unrecognized option on to an array and send the whole array to shc (and for sh-gcc too)

simonlindholm (commented):
We should add a mention of these scripts somewhere, too, like in the descriptions that get printed in tests/add_test.py when a compiler is not found.

sozud:
> We should add a mention of these scripts somewhere, too, like in the descriptions that get printed in tests/add_test.py when a compiler is not found.

Ok added a mention of the script when it's not found

tools/shc_compiler/shc
sozud: Ok done

tools/shc_compiler/compile.sh
sozud: Ok done

PR #358, opened 2026-08-29
sh4: implement fldi0/1

sozud:
Implements FLDI0/1, which just load floating point 0/1 to a fp reg. Disclosure: AI assistance

### DIFF
--- a/m2c/arch_sh.py
+++ b/m2c/arch_sh.py
@@ -1294,4 +1294,55 @@ class Sh2Arch(Arch):
 
 
 class Sh4Arch(Sh2Arch):
-    pass
+    base_return_regs = [
+        (Register("r0"), False),
+        (Register("fr0"), True),
+    ]
+    all_return_regs = [Register("r0"), Register("r1"), Register("fr0")]
+
+    fp_argument_regs = [Register(f"fr{i}") for i in range(4, 12)]
+    fp_temp_regs = [Register(f"fr{i}") for i in range(12)]
+    fp_saved_regs = [Register(f"fr{i}") for i in range(12, 16)]
+
+    argument_regs = Sh2Arch.argument_regs + fp_argument_regs
+    temp_regs = Sh2Arch.temp_regs + fp_temp_regs + [Register("fpul")]
+    saved_regs = Sh2Arch.saved_regs + fp_saved_regs
+    all_regs = (
+        Sh2Arch.all_regs
+        + fp_temp_regs
+        + fp_saved_regs
+        + [Register("fpul"), Register("fpscr")]
+    )
+
+    @classmethod
+    def parse(
+        cls, mnemonic: str, args: List[Argument], meta: InstructionMeta
+    ) -> Instruction:
+        if mnemonic in ("fldi0", "fldi1"):
+            assert (
+                len(args) == 1
+                and isinstance(args[0], Register)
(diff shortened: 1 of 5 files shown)


m2c/arch_sh.py
```diff
@@ -1294,4 +1294,55 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+
+    @classmethod
+    def parse(
+        cls, mnemonic: str, args: List[Argument], meta: InstructionMeta
+    ) -> Instruction:
+        if mnemonic in ("fldi0", "fldi1"):
+            assert (
+                len(args) == 1
+                and isinstance(args[0], Register)
+                and args[0] in cls.fp_temp_regs + cls.fp_saved_regs
```
simonlindholm: I don't feel like this assert is needed

m2c/arch_sh.py
```diff
@@ -1294,4 +1294,55 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+                outputs=[args[0]],
+                eval_fn=eval_fn,
+            )
+
+        return super().parse(mnemonic, args, meta)
+
+    @staticmethod
+    def function_return(expr: Expression) -> Dict[Register, Expression]:
+        ret = Sh2Arch.function_return(expr)
+        ret[Register("fr0")] = as_type(expr, Type.f32(), silent=True, unify=False)
```
simonlindholm: I think you'll want a manually constructed Cast() here to mirror mips

m2c/arch_sh.py
```diff
@@ -1294,4 +1294,55 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+        + fp_temp_regs
+        + fp_saved_regs
+        + [Register("fpul"), Register("fpscr")]
+    )
+
+    @classmethod
+    def parse(
+        cls, mnemonic: str, args: List[Argument], meta: InstructionMeta
+    ) -> Instruction:
+        if mnemonic in ("fldi0", "fldi1"):
```
simonlindholm: Can you add a new InstrMap for this kind of const-loading instruction? I think it's nice if we can get all the actual instruction logic into those kinds of tables, leaving `parse` with only boilerplate.

m2c/arch_sh.py
```diff
@@ -1294,4 +1294,55 @@ def function_return(expr: Expression) -> Dict[Register, Expression]:
...
+
+            return Instruction(
+                mnemonic=mnemonic,
+                args=args,
+                meta=meta,
+                inputs=[],
+                clobbers=[],
+                outputs=[args[0]],
+                eval_fn=eval_fn,
+            )
```
simonlindholm, line 1342: Instead of returning an Instruction from each branch arm, can you do live Sh2Arch and have each arm set up inputs/outputs/eval_fn/etc.? Then at the end of the function we can do `if eval_fn is not None: return Instruction(...)`

sozud: Ok should be the same now

m2c/arch_sh.py
sozud: Added the map

m2c/arch_sh.py
sozud: Ok mirrored mips

m2c/arch_sh.py
sozud: removed

PR #360, opened 2026-08-30
flow_graph: resolve jump-table entries encoded as symbol+offset

StilesCrisis:
## Problem

m2c crashes with `AssertionError: jtbl list must not be empty` (in
`if_statements.py`'s `add_labels_for_switch`) on some switch statements,
instead of producing output. This showed up decompiling MWCC-compiled
PowerPC code whose asm was produced by decomp-toolkit.

## Root cause

Not every toolchain emits a dedicated label for every switch-case jump
target. In the failing case, some jump-table entries reference the
enclosing function symbol plus a byte offset instead of a bare label
(e.g. `.4byte some_func+0x38`), when the target address doesn't
otherwise need its own label - here, because several targets happen to
alias the switch's own post-dispatch continuation point.

The jtbl-resolution loop in `build_graph_from_block` only knew how to
resolve a bare-symbol entry (`entry.as_symbol_without_addend()`). When
that returned `None`, the loop assumed it had hit padding and `break`,
silently dropping every remaining case - including all real ones, since
in the trigger case a `symbol+offset` entry comes first. The resulting
empty case list was only caught much later, as an unrelated-looking
assertion in `if_statements.py`.

## Fix

Adds `get_symbol_plus_offset()` and `insert_self_relative_jtbl_labels()`
in `flow_graph.py` to synthesize labels for these targets before
block-building, and teaches the jtbl-resolution loop to fall back to
them instead of assuming padding. Bare-label jump tables are unaffected.

## Testing

- Added `tests/end_to_end/switch-jtbl-symbol-offset`: real MWCC/
  decomp-toolkit output that reproduces this exactly (I hit this
  decompiling a GameCube game's disassembly and pulled the minimal
  slice that still triggers it - the function name is genericized to
  `test`, but the instructions/jump-table entries are real compiler
  output, not hand-written). Confirmed it fails with the assertion
  above on unmodified `master`, and passes with this change.
- Full suite: 432/432 passing (`python run_tests.py`), including the
  new case.


simonlindholm:
Sorry, I've grown allergic to this kind of AI-generated PR descriptions and code. Can you explain what the issue is?

StilesCrisis:
No worries, I agree reading Claude text is exhausting.

I'm starting a decomp project and running m2c over CodeWarrior 2.6-era PowerPC code. I ran into a hard assert (`AssertionError: jtbl list must not be empty`) when decomping a function containing a switch. I asked Claude to dig into the failure and it found that m2c expects switch jump tables to point directly at labels, whereas decomp-toolkit had created a `label + offset` style jump table. The fix code is entirely Claude-written but it did solve the assertion in my case. I'm currently running m2c over my entire project and I don't think it's asserted once.

simonlindholm:
Right, I see. Do you think this could be fixed on the decomp-toolkit side? m2c has generally taken the approach of putting the responsibility for this sort of asm cleanup on the disassembler.

StilesCrisis:
I mean, I'm not opposed to doing both, but m2c is asserting on reasonable input that assembles and runs fine. It shouldn't assert, don't you think? And we've got a pretty self contained fix in hand.

simonlindholm:
`.include "rom.bin"` would also assemble and run fine, yet m2c does not ship with a disassembler, but rather treats that as another tool's job. It's a balance, and one which in this case I think comes down on the side of "this seems better to fix in the disassembler". And if it gets fixed in the disassembler, adding code for it in m2c feels like a poor idea, it just adds code bloat and maintenance overhead for no reason.

The error handling could indeed be made nicer, I haven't really prioritized that much.

StilesCrisis:
Well, in this case, honestly the assertion did us a favor, as it uncovered a class of functions which I couldn't decomp but now I can. If we just made it fail to decomp silently, that actually seems worse.

I'll take a look at making the fix in dtk and see if it's tractable.

StilesCrisis:
OK, so starting to look into this. It turns out that this happens when the jump table is bucketed in a different split than the function that uses it. 

That probably shouldn't happen in an ideal world, but I'm not sure yet if this is easy to fix or impossible. If the jump table is physically far away from the actual switch, it might have to be in two different buckets.

StilesCrisis:
I've sent a dtk PR. https://github.com/encounter/decomp-toolkit/pull/146

simonlindholm:
Does that PR fix all the jtbl issues? Or are there still ones that are problematic? (I don't quite follow why it might be impossible to fix.)

StilesCrisis:
If we make the (extremely safe IMO) assumption that jump tables aren't shared between multiple functions, then it does solve the issue. (if they are shared, then the code will come out like it does today and that assertion would fire)

That being said, no one has looked at the PR yet over there.

StilesCrisis:
I’ve gotten no response at all from @encounter on decomp-toolkit. I am not sure how to proceed here, as both PRs solve the issue but I’m not getting traction on either side. I just bumped into the issue again in a fresh checkout and was reminded about it the hard way :(

StilesCrisis:
Never mind! Encounter just wrote back. He wants cleaned up comments, no worries.

StilesCrisis:
The dtk PR is fairly complex; I ended up finding a bug with it and needed to abandon it. I’ll try again. That being said, this definitely seems to be a simpler path to a solution.

StilesCrisis:
I put up a new dtk PR at https://github.com/encounter/decomp-toolkit/pull/147 which fixes the issues from its pull/146, but is a lot more complex than this fix. If @encounter rejects it, I would appreciate a revisit of this PR, which feels much simpler overall.

simonlindholm:
Wow, that is incredibly complex. Can you explain roughly what's causing the complexity? I don't understand anything about what dtk does, so I don't have a sense of it, my feeling was that it was just a matter of resolving some labels.

It a dtk fix ends up not working out we can revisit this, yes (but it would also need to take a bit of iteration to get into a good place, probably involving me finding time to rewrite the code by hand because ugh, slop).

StilesCrisis:
I can de-slop all the comments here; I think it's totally reasonable to ask for that cleanup. (I've made the same request many times :)  ) I think the logic is sound but if you can point to a given place where it needs improvement, I am happy to rework it.

The challenge with dtk is because we need to concretely identify "this exact range is really a jump table for a function nearby" and if we are wrong, the linker is likely to blow up. Specifically, if you ever generate a range that forms a reference cycle (a -> b -> a), the linker will immediately trip an error. So you need to check your work very carefully.

With m2c, the problem is way more self contained. We just need to generate synthetic labels in the case where m2c used to bail out ("Cannot find jtbl target"). By comparison this work is trivial, and if your code already has correct labels, it's essentially inert/free since the new code will never even execute.

So that's my rationale for this being the better patch; I haven't heard from @encounter yet on the new https://github.com/encounter/decomp-toolkit/pull/147 PR but in my ideal world I think we're better off landing this and abandoning that one.

StilesCrisis:
The comments have now been de-Claudified. Hopefully that works.

StilesCrisis:
(PS I see the commit comment says "co-authored by Sonnet" but this co-authorship was limited to me saying "push this update to the PR"--I wrote the comments)

StilesCrisis:
@simonlindholm Friendly ping now that the comments have been deslopped?

simonlindholm:
It's not just the comments, it's also the code being kinda bloated and in the wrong place and applying even on architectures without 4-byte instructions. I'll get back to this

StilesCrisis:
I can try to thread it in differently if you could guide me a bit as to how you'd like the PR to work. I don't want to ask you to do the work, I'm happy to iterate on it with a bit of directional guidance.

PR #362, opened 2026-09-03
Fix __attribute__ position in visit_Decl for plain local declarations

StilesCrisis:
The permuter was generating this:
```
__attribute__((aligned(16))) f32 mtx[16];   // leading attribute
```

CodeWarrior 2.6 does not support attributes in this position; the traditional location would be:
```
f32 mtx[16] __attribute__((aligned(16)));   // trailing — correct, what the fix produces
```
This PR fixes the emitter to write attributes last, solving a lot of m2c failures.

Every other `_generate_attrs()` call site in the file emits trailing attributes; it's not clear why this one was different. This same fix was found via decomp-permuter's own frozen copy of this file (which was vendored from m2c) - see simonlindholm/decomp-permuter#210.


StilesCrisis:
This PR landed in the decomp-permuter project as https://github.com/simonlindholm/decomp-permuter/pull/211

PR #363, opened 2026-09-07
Parse computed gotos and label addresses

JohnPB-1:
GCC's `goto *expr` and `&&label` currently fail to parse. Ratchet & Clank 1 uses them to keep the original jump-table dispatch, for example in Aridia's func_0021F0B0:

```c
i = arg0 - 9;
if ((u32)i < 0xf) {
    goto *(void *)((u32 *)jtbl_001F8690)[i];
}
```

Add IndirectGoto with an expression child and LabelAddress with a label name. A label address has type void * in GCC. Keeping it separate from ID lets consumers distinguish labels from variables.

```sh
python - <<'PYCODE'
from m2c_pycparser.c_parser import CParser
from m2c_pycparser.c_generator import CGenerator
from m2c_pycparser.plyparser import ParseError

source = """
void dispatch(int index) {
    static void *const table[] = {&&one, &&two};
    goto *table[index];
one: return;
two: return;
}
"""
try:
    tree = CParser().parse(source)
except ParseError as error:
    print(error)
else:
    output = CGenerator().visit(tree)
    CParser().parse(output)
    print(output)
PYCODE
```

Before: a parse error at `&&`. 
After: it prints C that parses again.

Six tests cover expression children, label nnames, typedef names, target precedence, relative label tables, and ordinary goto/logical-and syntax. The full suite passes: 437 tests. Mypy passes. Generated c_ast.py and yacctab.py are included.

GCC documents these forms in [Labels as Values](https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html).


simonlindholm:
... that's a switch statement you have there, not a GCC computed goto. Closing due to slop

